PR #525, opened 2014-12-18
Added image file settings, admin panel

tcinelli:
Admin settings added to File settings panel:
Maximum preview image width (in pixels, optional) - if not set, will default to current 200px hardcoded setting
Maximum preview image height (in pixels, optional) - if not set, will default to current 200px hardcoded setting
Hide file info (name, size) for images on wall (defaults to current show info setting)

Added css, slight padding between preview images (images look better if larger, but stacked).


dimontajik:
must be like this

dimontajik:
it is necessary that it is automatically distributed across the width of the photo

dimontajik:
please finish everything is normal and not through the ass

acinelli:
Yes, that would nice, please implement dimontajik.  I see you haven't contributed any code, this would be a good place to start and not just talk "through the ass".

PR #540, opened 2014-12-22
improved IE rendering

skazi0:
added meta forces highest document mode available in installed IE browser. useful in environments where domain policies force compat mode.


andystrobel:
@skazi0 I'm not so happy to use respond.js together with modernizr.js. I read a lot about weird behaviors about this constellation (e.g. with icon fonts). This is maybe the reason why respond.js is no longer shipped with modernizr.js since 2012. But maybe I'm wrong and you have some other arguments?

skazi0:
I haven't noticed anything bad happening on IE8. All FontAwesome icons display correctly. I tried to put this whole thing into a module but there's no way to set order of loading js files and it's critical for respond.js to work.
What do you think about adding an option in the admin panel (disabled by default)?

PR #578, opened 2015-01-10
Update style.css to include basic truncate

WASasquatch:
In the event that a user title is to long, the truncate class should be applied to it.

For example, my Random User Module will break if a username is too long. The user account dropdown looks out of sync with a large user title, and also breaks if you minimize the browser into a smaller frame.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -47,6 +47,12 @@ hr {
 textarea {
   height: 1.5em;
 }
+.truncate {
+  width: 150px;
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
 .topbar {
   position: fixed;
   display: block;


andystrobel:
Thank you for your commit! Can you please add these class to the style.less file and compile then the style.css? Otherwise your changes will overwritten by next changes at the style.less file.

WASasquatch:
Thanks I still don't know how to 'theme' because of the lack of documentation, or even how to "compile" (which I have never had to do in almost 15 years of development). These features should be part of the ACP. 

So far, I do not update HumHub because of it's horrid implantation of file management and theme management, I do the edits from a master myself.

andystrobel:
As you already know, HumHub based on Yii and bootstrap. For these there are a lot of documentations out there. Regarding the theming we use the standard Yii theming mechanism and just complement and overriding the standard bootstrap classes in our theme to let it easily updatable.

You can learn more about here:
http://www.smashingmagazine.com/2013/03/12/customizing-bootstrap/
https://github.com/humhub/humhub/blob/master/protected/docs/guide/theming/index.md

WASasquatch:
Which doesn't work half the time. :( 

Half the views are not overriding. 80% of on-screen elements are stylable. Needs a lot of work besides relying on a unfinished templating model from Yii :(. Yii has abandoned Yii 1.x because of it's "happy coding" ideology which prevents true customization.

PR #592, opened 2015-01-12
Update File.php

hagalaz:
to prevent error


WASasquatch:
There will need to be a conditional to prevent the image handling after it fails.

luke-:
Can you please accept our CLA @ https://www.humhub.org/cla/github/sign

PR #606, opened 2015-01-16
Added arrayCompVal()

WASasquatch:
Added `Helpers::arrayCompVal()` for JSON/Array/Object value comparison. Useful in data handling.


luke-:
Is this used somewhere?

WASasquatch:
What do you mean? It's a method of comparing arrays? It is out of my old AJAX chat (never released)

PR #610, opened 2015-01-18
added social icons on wall entries

Lokke:


andystrobel:
Nice idea, but it should be comes with an option at the admin section to activate or deactivate this feature.

Lokke:
if i figured out how to create it as module i will do it

PR #615, opened 2015-01-18
Update theme.css

Lokke:
added overflow: scroll; to body - now the site won't look like skaking by loading more content


andystrobel:
Thanks @Lokke, but I changed it at https://github.com/humhub/humhub/commit/00c365594d2306c5cf30275314225fbfb0ae84c7 to:
```css
overflow-y: scroll;
```
Because it's not so pretty if you see always the horizontal scrollbar (which you never need). I hope it is okay for you so :-) Thanks for PR!!!

PR #623, opened 2015-01-19
Additional placeholder for top level foot code

skazi0:
We have to include some foot code on all pages. It seems that currently there's no way to do this but to copy whole main.php into our own theme. This PR introduces additional placeholder to handle such cases (similar to head.php which is inserted into head element).
The code is inserted just before the last script to enable things like replacement of standard input elements.

### DIFF
--- a/protected/views/layouts/main.php
+++ b/protected/views/layouts/main.php
@@ -227,6 +227,10 @@
 </div>
 <!-- end: Modal -->
 
+<!-- start: render additional foot -->
+<?php $this->renderPartial('//layouts/foot'); ?>
+<!-- end: render additional foot -->
+
 <script type="text/javascript">
 
     // Replace the standard checkbox and radio buttons


WASasquatch:
It is a little redundant I have to admit. Currently the system works off all placement through widgets. So it'd be best to have the footer as an actual widget, passing information to be used in the footer for necessary things (like theme credentials and other stuff down the road)

This is why I think we edit main.php as it includes all the resources one might need. 

Tell me if I'm wrong.

skazi0:
Actually our module is implementing a kind of chat widget that we want to have included on all pages. The code we wanted to place in the foot.php is just inserting this widget into the page. Copying the whole main.php to the theme just to insert this widget looks like a huge overkill as it needs to be manually synced after every change in upstream code. For the same reason I don't want to modify the original main.php. The idea was that with this foot.php the code would be a bit more modular. If you have a better idea for this kind of placeholder, I can try to implement it.

WASasquatch:
Yeah, that does make more sense to have a footer like that, I just feel this implementation is a little more of a hack, and you might as well just `include()` a file.

skazi0:
I didn't like to modify the original file so I copied it to the theme. The problem I feel with this approach is that I need to copy a big chunk of code that might change in future versions and I'd need to watch it closely. The idea with extracting this was supposed to be similar to head.php where you can have a limited scope for you local changes.
Anyway, I agree that it would be best if such top level 'injections' were somehow possible from modules. This way our chat would be completely self contained. I could try to implement this but at the moment I have too little knowledge on the internals of humhub so for now I'll have to stick to the copying of main.php approach.

PR #662, opened 2015-01-30
Add authentication mode for LDAP to use the local database

carlosgcampos:
Add a new LDAP setting auth-mode, that can be set to LDAP to use the
LDAP to authenticate users like always, or local to import the users
from LDAP but setting their authentication mode to local.
When a new user is imported from the LDAP and the auth-mode setting is
set to local, a Welcome email is sent to the user with a reset password
token so that the user can set up the password.


luke-:
Awesome, can you please send this PR to the develop branch?
We try to collect & tests new features there.

Thanks!

PR #669, opened 2015-01-31
Edit URL space

Kakiho:
Edit URL space


luke-:
GUIDs shouldn't be changeable. Maybe we should add an extra url attribute or use the name for space urls.

Kakiho:
this is a quick fix =)
"Maybe we should add an extra url attribute or use the name for space urls." - yes, and URL types of: http://mysite.com/my_awesome_space/

dakur:
Also there is a question, if the label shouldn't be passed through Yii::t() translator...

PR #679, opened 2015-02-04
[oembed] Added Flickr support to OEmbed

tanty:


tanty:
I have not tested it, though!

PR #856, opened 2015-03-28
Update WebUser.php

WASasquatch:
Add ability to check users for administration rights

### DIFF
--- a/protected/modules_core/user/components/WebUser.php
+++ b/protected/modules_core/user/components/WebUser.php
@@ -106,9 +106,9 @@ class WebUser extends CWebUser
      *
      * @return int
      */
-    public function isAdmin()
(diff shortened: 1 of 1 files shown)


luke-:
WebUser should only represent the current User.
Maybe we can add this to the User model?

So we can do: User::model()->findByPk($id)->isAdmin()

WASasquatch:
Sounds good. Just trying to get any hint of user permission validation in
the system without being that user. for a framework system we should really
have a basic permission system for modules/groups/spaces. lots of ideas I
want to starr but basic functionality doesn't exist.

luke-:
https://github.com/humhub/humhub/blob/master/protected/modules_core/user/models/User.php

PR #863, opened 2015-03-30
Update Language

JeXawoW:


luke-:
Do you know http://translate.humhub.org? Maybe it's easier to update the language there. You can also see the status (with progress) of each Module/File.

JeXawoW:
@luke- ok, thanks

PR #883, opened 2015-04-01
Update installation.md

WASasquatch:
Prevent 500 Non-Writable Errors warn user to pass CHMOD permissions to YIIC resources.

### DIFF
--- a/protected/docs/guide/administration/installation.md
+++ b/protected/docs/guide/administration/installation.md
@@ -18,7 +18,7 @@ Installation
     - /protected/runtime
     - /uploads/*
 
-4. Make the following files executable by the webserver
+4. Make the following files executable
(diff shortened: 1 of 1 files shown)


luke-:
+    - /protected/yiic
+    - /protected/yiic.bat
+    - /protected/yiic.php

Should be writable --> executable....

WASasquatch:
Ugh I purposely had Executable in my head, I swear. I have bad dyslexia. Coders worse nightmare haha. How do you edit pulls without it trying to create another?

luke-:
Thanks, but 4.) Make the following files executable ~~by the webserver~~

WASasquatch:
Edited, though not sure why we'd want it to be executable by a group or user and not only web server?

PR #927, opened 2015-04-09
translated a bunch of stuff to arabic

renoa:
if i have time i'll try to do others today.


renoa:
i translated all the files in this directory, but it's only showing widgets_views_stream 
do i need to make a pull request for each one individually? (sorry i'm pretty new to github)

PR #928, opened 2015-04-09
Update widgets_ContentFormWidget.php

renoa:
changed the word to a more suitable one


luke-:
You can also use our translation editor @ http://translate.humhub.org

Issue #929, opened 2015-04-10
LikeController error!


JeXawoW:
PHP Fatal error:  Call to a member function delete() on a non-object in /home/xxxxx/public_html/protected/modules_core/like/controllers/LikeController.php on line 74

ASPIRINpp:
"Double dislike" - if dislike was already sent. For example: two tabs in browser and dislike was sent from both of them... I`m added if statement :)

WASasquatch:
`empty()` should only be used on strings. Since we are dealing with objects, and variables, you should use `is_object()` or test `isset()` on a attribute.

PR #955, opened 2015-04-13
fix: LikeController error! #929

ASPIRINpp:


ASPIRINpp:
Sorry, my bad habit...

Issue #912, opened 2015-04-07
Set user group_id by ldap


freshp:
I want to map ldap groups to humhub user groups.

There is an issue in the UserIdentity->authenticate method. In this method one if condition call the HLdap::getInstance()->authenticate method. In the method handleLdapUser is called and the usermodel will be updated. But after all, the UserIdentity->onSuccessfulAuthenticate method will be called. This method get the usermodel, which was instanced before all modification. Now all updates will be overwritten.

sascha-seyfert:
+1

ychoucha:
+1

luke-:
@freshp Thanks for the fix, I've moved the code a little bit deeper - Hope this also fixes the problem with less overhead :-)

freshp:
thank you. i will wait for the next version.

PR #958, opened 2015-04-15
#912 fix user model overwrite after LDAP mapping

freshp:
get current user model for onSuccessfulAuthenticate to prevent overwrite after LDAP mapping

### DIFF
--- a/protected/modules_core/user/components/UserIdentity.php
+++ b/protected/modules_core/user/components/UserIdentity.php
@@ -49,7 +49,7 @@ class UserIdentity extends CUserIdentity
         } elseif ($user->status == User::STATUS_NEED_APPROVAL) {
             $this->errorCode = self::ERROR_NOT_APPROVED;
         } else {
-            $this->onSuccessfulAuthenticate($user);
+            $this->onSuccessfulAuthenticate($this->getUser());
         }
 
         return !$this->errorCode;


freshp:
The main Problem ist that the *else* function redaclares the usermodel and overwrite the changes from LDAP mapping. The LDAP mapping ist triggered from the *elseif* in *validatePassword*
```php
        $user = $this->getUser();
...
        } elseif (!$this->validatePassword($user)) {
            $this->errorCode = self::ERROR_PASSWORD_INVALID;
        } 
...
        } else {
            $this->onSuccessfulAuthenticate($this->getUser());
        }
```
*validatePassword* saves the usermodel and *onSuccessfulAuthenticate* uses the instanced usermodel from above. Thatswhy it is nessecary to get the current *user* model.

WASasquatch:
Noo... The elseif is 'testing' a already declared variable. Its not altering it or refeclaring it. And if we get to the else we are running the same functione`getUser()` concurrently which can actually cause errors, such as the user info updatonh between the two declarations via another tab/admin.

freshp:
I will try to explain it again with an example:

* the *authenticate* function is triggered
* we request the current usermodel with the group_id 1
```php 
$user = $this->getUser();
```
* in the if clause we check if it is null
```php
if ($user === null) {
```
* and the usermodel is not null, because we get a full usermodel
* the elseif will be checked
```php
} elseif (!$this->validatePassword($user)) {
```
* now the *validatePassword* function will be called
* in this function the *HLdap::getInstance()->authenticate(* will be called
* in this *authenticate* function the user model will be manipulated, if there are some other groupsettings in LDAP. for example the new group_id 99. Now the usermodel will be saved to the database with the group_id 99.
* after this manipulation the next *elseif* will be checked
* if there are no errors the *else* clause and the *onSuccessfulAuthenticate* function will be triggered
* this function get the 'old' usermodel with the 'old' group_id 1
* by calling the lines below the user with group_id 1 will be saved to the database
```php
$user->save();
```

This is the current situation. Now there are different way to fix this overwrite.

The simple way would be, to get the current usermodel from database after calling the *validatePassword* function. And this, the current pull request will do. 

If you prefer another solution for this bug, please tell me which?

WASasquatch:
I believe authentication needs to be fixed. That is effectively calling two DB calls of the same nature simultaneously. Asking for bugs and errors. including other imfo changed third party during the process.

PR #990, opened 2015-04-27
Update SpaceUrlRule.php

WASasquatch:
***Dual-Purpose GUID / Name system for Spaces***

Space names now used in URL for human-readable links.

**Non-compliant** characters cause URL controller to fallback to GUID.

I would like to add "article" like urls, but unfortunately I cannot until full text indexing is enabled on HumHub data tables. For example a url like 
`s/Nutrients+%26+Deficiencies/home` could be `s/Nutrients-Deficiencies/home`
and
`s/Seed+Banks/home` could be `s/Seed-Banks/home`

Using a similar query like this through Yii match functions
```sql
SELECT name FROM space WHERE MATCH (name) AGAINST ('%Nutrients-Deficiencies%' IN BOOLEAN MODE)
```

### DIFF
--- a/protected/modules_core/space/components/SpaceUrlRule.php
+++ b/protected/modules_core/space/components/SpaceUrlRule.php
@@ -33,9 +33,17 @@ class SpaceUrlRule extends CBaseUrlRule
     {
 
         if (isset($params['sguid'])) {
+            
+            /* Can we use a name instead? */
+            $space = Space::model()->findByAttributes(array('guid' => $params['sguid']));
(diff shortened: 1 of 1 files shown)


WASasquatch:
I could also add third handling option for a space token which could be set via the Space -> General Settings page. If a token is set it could be used instead. For example `nutes-and-defs` could be the token set for the space `Nutrients & Deficiencies`, if a token is used and refers to a valid token in the DB, it would be used instead of Space name or GUID. 

If that sounds like a good idea I can make the relevant changes.

JeXawoW:
+1

WASasquatch:
Made a optimization update to the code to lessen profiling

WASasquatch:
Is there more I can do for this pull, @luke- ?

Vatarus:
+1

PR #1076, opened 2015-06-02
Update Notification.php

badis:
Change to path to avoid error ( failed to open stream: No such file or directory )


luke-:
Thank you, can you please accept our CLA at https://humhub.org/cla/github/sign

PR #1080, opened 2015-06-08
Update StreamAction.php

badis:
Delete a useless semicolon symbol


luke-:
Please note the StreamAction.php is not longer used - use BaseStreamAction instead.
https://github.com/humhub/humhub/blob/master/protected/modules_core/wall/BaseStreamAction.php

https://github.com/humhub/humhub/blob/master/protected/modules_core/wall/StreamAction.php#L6

PR #1085, opened 2015-06-11
Update fileuploader.js

badis:
Check file size before start downloading


protected/modules_core/file/resources/fileuploader.js
```diff
@@ -26,6 +26,38 @@ function installUploader(uploaderId) {
                     $('#fileUploaderProgressbar_' + uploaderId).children().css('width', progress + "%");
                 }
             }
+        }).bind("fileuploadadd",function (e, data) {
+                        
+            var $max_file_size = 4 * 1024 * 1024;
```
bate: This will fail with default php.ini because default value for upload_max_filesize = 2M

PR #1246, opened 2015-10-16
Update wallCreateContentForm.php

davidjeddy:
Properly escape a single quote inside double quotes.


luke-:
Hmmm, is it really required to escape message strings?

luke-:
Sorry for the late reply :)

davidjeddy:
For IDE smart formatting, yes.

cebe:
it is not required, if your IDE complains I would consider that an IDE bug.

PR #1291, opened 2015-11-05
ADDED data-name property to DIV.media-body for testing

davidjeddy:
... making  enable/disabling modules much easier as now we can select the module no matter where in the list or order the module is.


luke-:
can you please name it 'data-module-name' and html encode the value?

davidjeddy:
Sure thing @luke- , will resubmit PR after words.

PR #1297, opened 2015-11-10
Fixes some french translations

pieuvre:


buddha87:
Thanks, seems to be already translated yet.

PR #1385, opened 2015-12-24
Update zh_tw translation

azdkj532:


luke-:
FYI you can also use http://translate.humhub.org - which provides an webinterface

azdkj532:
Thanks. But I didn't find space for zh_tw. Can I create one by myself?

PR #1425, opened 2016-01-13
Add missing French Translation

happyfredo:
Add missing term for search module

### DIFF
--- a/protected/humhub/modules/search/messages/fr/views_search_index.php
+++ b/protected/humhub/modules/search/messages/fr/views_search_index.php
@@ -17,8 +17,8 @@
  * NOTE: this file must be saved in UTF-8 encoding.
  */
 return [
-    'Advanced search settings' => '',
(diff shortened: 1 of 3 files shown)


protected/humhub/modules/search/messages/fr/views_search_index.php
```diff
@@ -17,8 +17,8 @@
  * NOTE: this file must be saved in UTF-8 encoding.
  */
 return [
-    'Advanced search settings' => '',
-    'Search for user, spaces and content' => '',
+    'Advanced search settings' => 'Recherche avancée',
```
Felli: This should be  ```'Advanced search settings' => 'Recherche avancée réglages',``` should it not?

happyfredo: you're right i will close it and open a new one

PR #1473, opened 2016-02-11
Multipart e-mail messages

dakur:
Issue 631


luke-:
Awesome!

Maybe it's better (in future) to but complexer commits like this to the next version branch (e.g. https://github.com/humhub/humhub/tree/v1.1-dev)

PR #1497, opened 2016-02-24
Fix  translation in views_auth_login.php

mazurva:
Now the Russian version of literally translated as follows: `user name or password`.
I just replaced the `пароль` (password) on `e-mail`.

### DIFF
--- a/protected/humhub/modules/user/messages/ru/views_auth_login.php
+++ b/protected/humhub/modules/user/messages/ru/views_auth_login.php
@@ -29,5 +29,5 @@ return [
     'Sign in' => 'Войти',
     'email' => 'email',
     'password' => 'Пароль',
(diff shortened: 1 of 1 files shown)


protected/humhub/modules/user/messages/ru/views_auth_login.php
```diff
@@ -29,5 +29,5 @@
     'Sign in' => 'Войти',
     'email' => 'email',
     'password' => 'Пароль',
-    'username or email' => 'Имя пользователя или пароль',
+    'username or email' => 'Имя пользователя или e-mail',
```
Felli: This can be replaced with ```почта``` meaning mail.

luke-:
Can you please change the translation in http://translate.humhub.org

Issue #1459, opened 2016-02-03
Notification of Comment Like indicates Post instead Comment.


Ardnived:
Whenever a user gets a notification that someone has liked one of their comments, the notification instead indicates that the other user has liked the parent post.

So as shown in the screenshot below:
Liesbeth and Sander have liked a comment by our user, but instead it is indicating that they have liked the Post which the comment is attached to.

![download](https://cloud.githubusercontent.com/assets/1362167/12799044/ac95a080-ca81-11e5-80e3-fb2a239a903d.png)

Issue #1442, opened 2016-01-25
'Firstname' should display as 'First name'...


davidjeddy:
...in visible content views.
{app root url}/u/admin/user/profile/about
as well as 
namespace humhub\modules\user\models\Invite->attributeLabels();

PR #1498, opened 2016-02-24
fix issue #1442 and #1459

helnokaly:
fix issue #1442 'Firstname' should display as 'First name' ....
fix issue #1459  Notification of Comment Like indicates Post instead Comment.


luke-:
thanks!!! - can you please accept our CLA at https://humhub.org/cla/github/sign

PR #1500, opened 2016-02-25
Fix $filePath generation with Nginx when X-Sendfile is used

phorque:
When I tried to use the "X-Sendfile" option with Nginx no images were displayed and I got a bunch of `substr() expects parameter 1 to be string, object given` in the logs.

Seems like we should use $filePath instead of $file here, I tried it and everything works great now :3


luke-:
Thanks!

PR #1517, opened 2016-02-29
Added foreign keys

VasileGabriel:


VasileGabriel:
Now that the tables have foreign keys, most of the "EVENT_BEFORE_DELETE" callbacks are not necessarry as the db will handle these deletes. If you think the same, then I will proceed and rework that.
I think this will also raise more issues in the tracker as some people will miss this detail and ask how the comments are magically deleted when the user is removed.

buddha87:
Hi, could you create another pull request for branch v1.2-dev please.

Issue #1515, opened 2016-02-29
Edit post or comment with empty string in [Firefox, Safari]


ali-abdalla:
In **Mozilla Firefox** and **Safari** browsers, there's an issue while editing the post or comment.
Here's the steps to produce this issue:
1. Step One, User creates new post.
![screen shot 2016-02-29 at 6 24 12 pm](https://cloud.githubusercontent.com/assets/14368377/13400799/2850ef1e-df12-11e5-912e-dc7b6e5ace4d.png)
2. Step Two, User edits the post and set an empty string.
![screen shot 2016-02-29 at 6 24 23 pm](https://cloud.githubusercontent.com/assets/14368377/13400815/35d76910-df12-11e5-9307-563ee0e4665c.png)
3. Step Three, The post changed to an input field.
![screen shot 2016-02-29 at 6 24 55 pm](https://cloud.githubusercontent.com/assets/14368377/13400821/3d21fb4a-df12-11e5-9c40-aee8ed12773c.png)

PR #1526, opened 2016-03-03
Fix issue #1515 Edit post with empty string

helnokaly:
Fix issue #1515 Edit post with empty string

Now editing a Post with empty string renders the Yii model rule error correctly
(Tested on Chromium, Firefox, Opera, Edge and IE11)
edit_post_rule


buddha87:
thank you.

PR #1549, opened 2016-03-09
allow manual setting of 'updated_by' property of ActiveRecord

helnokaly:
set 'updated_by' property to Yii::$app->user->id only when it is not set, otherwise use the current value.


luke-:
Hm, updated_by & updated_at should always refer to the latest modification.

With this PR, updated_at is the latest modificatoin and updated_by always the first?

helnokaly:
That's correct, I did not notice that, and that was not my intention from the change. So I'm closing this PR for now. Thanks luke

PR #1581, opened 2016-03-28
Role Concept

Tobi005:
Implemented a role concept which allows to display different profile fields to the different user groups.


chha25:
Awesome! ;-)

buddha87:
Hi, thanks for your efford, I did not have the time to review this in detail yet, but you should rather do a push request to v1.1-dev branch. Furthermore you should not edit the message files directly, we have a translation humhub instance for this purpose, it seems some message files you committed are older than the current master, there are missing translations (would be best if you do not commit the message files at all). And you'll have to fill out the cla agreement http://humhub.org/cla/github/sign. Note the v1.1 version supports multiple groups by user, perhaps this is affects your implementation.

Tobi005:
Ok, I will move the changes to the v1.1-dev branch.

PR #1585, opened 2016-03-30
profile field assignment

Tobi005:
Implemented a profile field assignment which allows to display different profile fields to the different user groups.

Thanks to @chha25, @leuthelt and @MrRioesTheOriginal


buddha87:
Thanks! We'll check this within the next days ;)

buddha87:
The installer fails when calling hasUserProfileField without a user, could you fix this. Furhtermore it would be great if you could add foreign keys to the migration as here: https://github.com/humhub/humhub/blob/v1.1-dev/protected/humhub/modules/user/migrations/m160229_162959_multiusergroups.php#L22

ERROR:

groupfielderror

buddha87:
If you want to execute the installer again on a already installed instance you can drop the tables of your test instance and remove the /protected/config/dynamic.php config.

PR #1586, opened 2016-03-30
Update index-test.php

davidjeddy:
Added ./config/dynamic to index-test to allow more parity between dev and testing application configurations.

### DIFF
--- a/index-test.php
+++ b/index-test.php
@@ -17,6 +17,10 @@ defined('YII_ENV') or define('YII_ENV', 'test');
 require(__DIR__ . '/protected/vendor/autoload.php');
 require(__DIR__ . '/protected/vendor/yiisoft/yii2/Yii.php');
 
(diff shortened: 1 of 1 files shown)


buddha87:
Hmm, I think the acceptence tests should not depend on the dynamic config at all, since all tests should execute with equal preconditions and should not interfere with the 'live' test environment.
Perhaps we should rather create and use an independent dynamic test config for acceptence tests, what do you think ?

davidjeddy:
Agreed, each test should execute in isolation. However some tests have dependencies, esp. acceptance tests. Exp: Login, change theme, check header color. Login, change theme, enable module, add content.

The theme changes no matter what, the deeper the dependency hole the longer each test takes to execute. Being able to set the theme during one 'changeTheme' test and have that change persist to 'add content module' is very valuable in acceptance test cases.

I am open to ideas for a solution as I am not a big fan of multi level dependency in my testing.

Issue #1616, opened 2016-04-07
Delete file from comment editing 


a-salah:
Comment creator can't  delete uploaded files in comment editing

PR #1617, opened 2016-04-07
Update File.php

a-salah:
#1616  Delete file from comment editing

the old code check only content creator not contentActiveRecord creator 

if Object is a comment (ContentAddonActiveRecord)
fire canWrite of it that means (comment creator can delete comment files)
else if object is a post (content) fire canWrite of it that means (post creator can delete post files)``


buddha87:
Hi, have merged it myself, not a fan of the no bracket if/else. Thanks for your effort !

PR #1643, opened 2016-04-13
profile field assignment

Tobi005:
Implemented a profile field assignment which allows to display different profile fields to the different user groups.

Thanks to @chha25, @leuthelt and @MrRioesTheOriginal

fixed bugs and added foreign keys


Tobi005:
are there any news for this pull request? :)

buddha87:
Hi, sorry did not have the time yet, but I will definetly test this within the next days ! :)

buddha87:
So I've tested it a bit and it looks good (on first sight), the problem is, you can't just unassign all fields by default :) So my suggestion would be to automaticlly assign all fields  (+new fields) by default to all groups, or perhaps it would be better to be able to exclude fields from groups instead of including fields. We just can't force everyone to assign all fields to all groups and I think an optional exclude of fields for specific groups would be more comfortable, since not everyone will needs this feature. What do you think ?

buddha87:
Furthermore it would be nice to be able to define groups when creating a new field. The User edit has an group chooser dropdown, which we could use for this.

PR #1675, opened 2016-04-27
Added "disabled" attribute to un-editable <select> fields

TadWohlrapp:
Now <select> fields which are un-editable (for example fields with ldap values) have the "disabled" attribute, so the user can't edit them. Before, they seemed to be editable, the user could change a value and click save, but the value would remain unchanged. Now the user sees that the field is un-editable.
Also fixed a typo and a small LdapException error.


luke-:
Thanks!

PR #1678, opened 2016-04-29
fixed redirect in login to avoid executing login logic again

cebe:
redirect should always return.


cebe:
seems this is broken in a lot of other places too...

cebe:
> seems this is broken in a lot of other places too...

-> #1682

PR #1682, opened 2016-05-01
cleanup overall usage of redirect()

cebe:
- make sure response instance is returned and nothing else is executed
  after redirect
- use consistent syntax, avoid extra calls to Url helper, which are not
  needed
- use controller redirect methods instead of
  Yii::$app->response->redirect()
- use goHome() and goBack() shortcut methods where applicable
- fixed some excpetion imports and catch blocks I found while fixing the
  above.


buddha87:
thanks!

PR #1736, opened 2016-05-19
make unimplemented functions in search class abstract

cebe:
does not make much sense otherwise. This ensures proper implementation of child classes.


luke-:
Maybe it's interesting for you: the Enterprise module contains a Solr search driver.

cebe:
it's not right now but thanks for the info ;)

PR #1741, opened 2016-05-19
Fixed typo in German translation: "Benachrichtungen"

cebe:
korrekt: Benachricht**ig**ungen


luke-:
Better use http://translate.humhub.org (master) and join the 'DE' space.

Otherwise there will be conflicts :-)

luke-:
Actually we removed extra UK language files for some reason:
https://github.com/humhub/humhub/blob/master/protected/humhub/components/i18n/I18N.php#L25-L26

I'm going to delete the en_* message folders and creation 
https://github.com/humhub/humhub/blob/master/protected/humhub/config/i18n.php#L8
in 1.1.

Maybe we should fix the wrong locale ids and historic handling in 1.2+

Or do you need extra uk messages in your project?

luke-:
Oups, wrong thread. :-)

cebe:
> Better use http://translate.humhub.org (master) and join the 'DE' space.

done.

cebe:
I do not need en_GB I am just confused why en_uk messages exist but the language selector offers en_gb.

luke-:
Yes, really confusing - seems to be a hack (maybe during Yii1 - Yii2 rewrite).

We should fix this in 1.2.x

We're going to release a 1.1 beta in the next days. (Hopefully)

Issue #1749, opened 2016-05-24
How to create a custom module from scratch in HumHub


nadilhassan:
I was start to work with hum hub and i would like to create a new module , So how can i do this ?
Is there any other documentations or examples other than www.humhub.org/docs/guide-dev-module.html

benmag:
I would love to see some improvement made to this. Creating my first few modules required a fair bit of trial and error just to get it showing up in the modules page. 

All the information required to make a module is there it's just it can take awhile on your first try. 

What if we built a module generator? So we could  start up a new module with a yiic command?

buddha87:
A Gii module generator would be great, can't predict when we'll have the resources to implement this but hopefully soon.

petersmithca:
I've started a tutorial here. tho haven't gotten very far, nor can I claim to understand everything.  http://showcase.thepetersmith.com/custom_pages/view?id=2  - note this is on my own little VPS, so not sure how much traffic it will support.  I have more written, just not added yet, will try to do that soon.

benmag:
Just opened a pull request which implements a Humhub module generator via the Gii module generator. 

When you run it, it will spit out an automatically generated humhub module with your ID and namespaces correctly set so you don't have to spend time doing that yourself. 

The module itself is very basic. It just adds an admin only page + menu item and a public page + top menu  item. 

In the future it could be expanded it to have a template for a basic content container module to help demonstrate how that can be implemented.

nadilaccura:
@benmag  Do you mean that attache the Gii Module to humhub application and then start generating modules from that Gii generator ? 
If that so can you please provide me the steps..

benmag:
@nadilaccura I updated the [dev-module docs](https://github.com/benmag/humhub/blob/16f7ab625c5b678d9c0e880c82b44fce7f6056fc/protected/humhub/docs/guide/dev-module.md) as part of my PR, not sure if you saw that or not.

nadilaccura:
@benmag When i was try to integrate yiigii module in humhub application it gives an Error

PHP Fatal Error – yii\base\ErrorException

Class 'HWebModule' not found

benmag:
@nadilaccura I don't think the `HWebModule` class is in the current version of HumHub? What version are you on? I've tested this on v1.0.0 and v1.0.1

nadilaccura:
@benmag  i am sorry the Error i get is 
ReflectionException

Class humhub\generators\module\Generator does not exist

I am using version: 1.0.1

[...]

PR #1751, opened 2016-05-25
HumHub Module Generator

benmag:
Implements a HumHub Module Generator #1749


luke-:
Cool!

Maybe we should create an extra module & repository for that?
Something like https://github.com/humhub/humhub-modules-devtools

Should I create this repo?

benmag:
A new repository & module would also work nicely

buddha87:
Hi, here is the repository: https://github.com/humhub/humhub-modules-devtools

PR #1759, opened 2016-06-01
Update statistic space after new message

kozhemin:
image


luke-:
Cool, can you please send this PR to the 1.2 branch?

Maybe we should also do the same at user profiles...

kozhemin:
ready (https://github.com/humhub/humhub/pull/1804)

PR #1785, opened 2016-06-08
Removed share-panel styles

cebe:
share widget has been removed in 0471bdc229dfe80e8b1ef03f3535037561513257 but style is still present in .less file in the theme.

Am I correct, that version 1.1 is now developed on master?


buddha87:
Thanks, yes v1.1 is on master.

PR #1789, opened 2016-06-09
ensure InnoDB Engine on mysql database

fredyns:
i found some shared-hosting set default DB-Engine to MyISAM. Which cause error when executing migration.

### DIFF
--- a/protected/humhub/components/Migration.php
+++ b/protected/humhub/components/Migration.php
@@ -1,5 +1,4 @@
 <?php
-
 /**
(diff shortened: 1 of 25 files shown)


luke-:
Thanks!

Maybe a better solution would be to overwrite the createTable method in https://github.com/humhub/humhub/blob/master/protected/humhub/components/Migration.php
and add the default db engine code logic once there.

What do you think about this approach?

fredyns:
i'll see that. thanks luke.

Issue #962, opened 2015-04-16
Move ico into Theme


davetheus:
Like to see the ico directory and all the assets which control favicon moved into the Theme and then the following should be moved from the main.php into the head.php to allow for customization of the favicon.

Remove from main.php:

```
<!-- start: Favicon and Touch Icons -->
<link rel="apple-touch-icon" sizes="57x57" href="<?php echo Yii::app()->baseUrl; ?>/ico/apple-icon-57x57.png">
<link rel="apple-touch-icon" sizes="60x60" href="<?php echo Yii::app()->baseUrl; ?>/ico/apple-icon-60x60.png">
<link rel="apple-touch-icon" sizes="72x72" href="<?php echo Yii::app()->baseUrl; ?>//ico/apple-icon-72x72.png">
<link rel="apple-touch-icon" sizes="76x76" href="<?php echo Yii::app()->baseUrl; ?>/ico/apple-icon-76x76.png">
<link rel="apple-touch-icon" sizes="114x114" href="<?php echo Yii::app()->baseUrl; ?>/ico/apple-icon-114x114.png">
<link rel="apple-touch-icon" sizes="120x120" href="<?php echo Yii::app()->baseUrl; ?>/ico/apple-icon-120x120.png">
<link rel="apple-touch-icon" sizes="144x144" href="<?php echo Yii::app()->baseUrl; ?>/ico/apple-icon-144x144.png">
<link rel="apple-touch-icon" sizes="152x152" href="<?php echo Yii::app()->baseUrl; ?>/ico/apple-icon-152x152.png">
<link rel="apple-touch-icon" sizes="180x180" href="<?php echo Yii::app()->baseUrl; ?>/ico/apple-icon-180x180.png">
<link rel="icon" type="image/png" sizes="192x192"  href="<?php echo Yii::app()->baseUrl; ?>/ico/android-icon-192x192.png">
<link rel="icon" type="image/png" sizes="32x32" href="<?php echo Yii::app()->baseUrl; ?>/ico/favicon-32x32.png">
<link rel="icon" type="image/png" sizes="96x96" href="<?php echo Yii::app()->baseUrl; ?>/ico/favicon-96x96.png">
<link rel="icon" type="image/png" sizes="16x16" href="<?php echo Yii::app()->baseUrl; ?>/ico/favicon-16x16.png">
<link rel="manifest" href="/manifest.json">
<meta name="msapplication-TileColor" content="#ffffff">
[...]

PR #1790, opened 2016-06-09
move ico to theme head

Daha62:
issue #962


cebe:
:+1: however this may break themes that override only one of these files.

Daha62:
Don't understand, how this may break themes

PR #1815, opened 2016-06-24
Show space acronym instaed of blank image in search results

cebe:
use the image widget instead of displaying the image directly as it is done in other places.


cebe:
@buddha87 @luke- could you please check this one? let me know if anything needs to be adjusted here. thanks!

PR #1820, opened 2016-06-29
Fix ru translation.

Daha62:


buddha87:
Hi, if you like you can contribute to our translation community under: http://translate.humhub.org/
Just join the space of your language code and strat translating. Its better that way, else there can occure some commit conflicts.

PR #1845, opened 2016-07-11
added access control by allowed groups

onmotion:
now in acl can write

    'acl' => [
                'class' => \humhub\components\behaviors\AccessControl::className(),
                'allowedGroups' => ['Administrators', 'Foo', 'Bar']
            ],


onmotion:
added use ArrayHelper namespace

PR #1846, opened 2016-07-11
added access control rules for acl

onmotion:
now in acl can write

    'acl' => [
                'class' => \humhub\components\behaviors\AccessControl::className(),
                'guestAllowedActions' => ['tmp'],
                'rules' => [
                    [
                        'actions' => ['index'],
                        'groups' => ['Administrator', 'Foo', 'bar'],
                    ],
                    [
                        'actions' => ['tt', 'save'],
                        'groups' => ['Сотрудник офиса'],
                    ],
                ]
            ],

So, for this example, if you set the rules, the guest will be able to perform the action of 'tmp', but it will not be able to perform those specified in the rules


luke-:
Awesome, can you please accept our CLA at https://humhub.org/cla/github/sign

onmotion:
Accepted

PR #1906, opened 2016-08-05
Added database port to installer

cesardraw2:
To facilitate database configuration, I added the "port " field.
Some database servers , do not use the default port of Mysql .


cesardraw2:
Docker test running success in Linux Alpine 3.3
https://hub.docker.com/r/cesardraw/humhub2-docker/
https://hub.docker.com/r/cesardraw/humhub2-docker/builds/

Issue #431, opened 2014-11-27
Feature: hashtags


acki:
Hi there

Is it planned to integrate a hashtag function?

Cheers

acki:
Any news? It sucks that you can't just hashtag some keywords

sandipp:
+1

samet:
+1

AlDemion:
+1

enginhu:
+1

CKGrafico:
+1

kkayacan:
+1

CKGrafico:
Suggestion:
With js: search '#' (more or less) in the posts and highlight the word.
Backend: when user click on a hastahg only filter on current space with this word.

not the best form but it's something functional

JulienMcJay:
I would really appreciate the feature in the next release.

aerth:
+1 sounds like a module to me

WASasquatch:
+1

Yeah I though hashtags and post searching was going to be part of 0.11.0-1.
Oh well, next time.

JeXawoW:
+1

ToonSoft:
+1 I hope the hashtags can be part of next release.

binitprads:
+1

jakeyjdavis:
+1

paoloimpelluso:
+1

IceSweets:
+1

brainscript:
+1

Censtudios:
This feature would be really awesome. You could then click the tags to find other related posts.

feldrim:
+1

lucasrcsantana:
+1

aerth:
+9000, who is doing it?

vladanPro:
+1 Is it going to be soon or i better start on my own? tnx

Censtudios:
Probably should start on your own lol =/ Luke said over a year ago he wanted to do this but I guess he focused on other things first. I think hashtags are like a key feature in social networks though.

feldrim:
https://github.com/humhub/humhub/issues/1347

tatianaJacques:
+1 !

primantok:
Hello @luke-

It seems that HumHub use at.js library (https://github.com/ichord/At.js) to handle @mention and :smileys

Is it posible to use Linky library (https://github.com/AnSavvides/jquery.linky) to handle linkify and #hashtag function for v1.2?

[...]

PR #1939, opened 2016-08-18
Hashtags

zhelyabuzhsky:
Hashtag function for feature request #431


Lokyaziis:
Post block on creation (infinite 3 points loader)!! But the post is created cause when I refresh the page the post is here with the hashtag... clicking on the hashtag link to a result page with some direct code on the left ({n, plural, =1{Found {n, number} post} =other{Found {n, number} posts}}) and no results... (doesn't found nothing) !! And as a french user, it created hashtag with all word with ' , Such as " l%#38amour" instead of "l'amour" !!

gigabytedevelopers:
Please can someone send me the table hashtag_post from their database, I have even tried the yii migration tool but the table has not been created in my database.. I need it, please.

b17:
@gigabytedevelopers Try this command to execute migrations:
```
./yii help migrate --includeModuleMigrations
```

luke-:
It's planned to add some topic feature to the 1.3 - since this solution is only for Posts we can't merge it to the core.

We need something more flexible for Content and also maybe Users & Spaces.

PR #1957, opened 2016-08-26
allow customization of Stream action via events

cebe:
added before and after action events that can be used to modify the
stream action filter.

### DIFF
--- a/protected/humhub/modules/content/components/actions/Stream.php
+++ b/protected/humhub/modules/content/components/actions/Stream.php
@@ -26,10 +26,12 @@ class Stream extends \yii\base\Action
     /**
      * @event ActionEvent Event triggered before this action is run.
      * This can be used for example to customize [[activeQuery]] before it gets executed.
+     * @since 1.1.1
      */
     const EVENT_BEFORE_RUN = 'beforeRun';
     /**
      * @event ActionEvent Event triggered after this action is run.
+     * @since 1.1.1
      */
     const EVENT_AFTER_RUN = 'afterRun';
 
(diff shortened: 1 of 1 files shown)


001101:
Can you please explain how i now can customize a stream after this update?

cebe:
For customization form a module, you can register an event handler in module config:

```php
    // to customize all streams
    ['class' => \humhub\modules\content\components\actions\Stream::className(), 'event' => \humhub\modules\content\components\actions\Stream::EVENT_BEFORE_RUN, 'callback' => ['humhub\modules\mymodule\Events', 'onStreamActionBeforeRun']],
    // OR to customize a specific stream (in this case the dashboard)
    ['class' => \humhub\modules\dashboard\components\actions\DashboardStream::className(), 'event' => \humhub\modules\dashboard\components\actions\DashboardStream::EVENT_BEFORE_RUN, 'callback' => ['humhub\modules\mymodule\Events', 'onStreamActionBeforeRun']],
```

The event handler would look like this:

```php
    /**
     * @param ActionEvent $event
     */
    public static function onStreamActionBeforeRun($event)
    {
        /** @var $action Stream */
        $action = $event->action;

        $action->activeQuery->andWhere(/* your custom condition */);
    }

```

001101:
Ah I understand now, thanks! Could you tell me how I can change the notifications when a user register and joins the default spaces, because that blows the activity streams (i have 28 default join spaces).

cebe:
you could check for `$this->mode == self::MODE_ACTIVITY` and the filter out the activity you are interested in. Not exactly sure how to do that.

PR #1999, opened 2016-09-22
Put "(count) post" (number of post in a space) of the space header in the user header

andrerudi95:
Hi!!!
 someone could help me to put "(count) post" (number of post in a space) of the space header in the user header? 
I did something but it doesn't work...
thanks for the attention
Andrea


skleveno:
I noticed the auto updater says no updates. If I installed humhub via
cpanel what is the best way to update and pull the latest files. If I use
SSH and putty to connect to my cpanel and do I github push will that
overwrite all my settings what about a github clone.. ty

Scott

PR #2000, opened 2016-09-23
Horizontal scrolling images on mobile devices

onmotion:
Added seeting for turn on/off horizontal scrolling images on mobile devices in wall entry.

[](http://postimg.org/image/89vixjrwj/)

Turning on

[](http://postimg.org/image/xemjayrcz/)

Now images is dragged left / right on any mobile devices.
 
[](http://postimg.org/image/ash7y8dmr/)


luke-:
Great! 
Maybe we should add this without an extra option in future releases (e.g. 1.2)

PR #2043, opened 2016-10-23
#1538 richtext translate domain name into url

### LINKED ISSUE #1538: Auto-link URLs without http
It would be nice if URLs would be auto linked to even if http:// was not added

b17:
Fixes #1538


buddh4:
Thanks and sorry for the late reply, I did some tests and there is a problem for links like www.humhub.com which is not matched right, think those cases should also be included. Futhermore there was a problem in my test environment somehow the gethostbyname always returns the ip address of my hoster for invalid hostnames, but think this more of a problem of the environment.

PR #2058, opened 2016-10-31
Changing Russian translation of "Spaces"

devochka:
I advise Russian-speaking contributors to change the translation of the word "Spaces" in Russian to more human.
From "Пространство" to "Сообщество".
I realize, that all of us are technical specialists and we are used to different terms, but for the average people, normal users of social network, the website has to be more clear.
Please let me know, if you think if it's reasonable or not, so I can change the rest of files.
Thank you!


devochka:
I advise Russian-speaking contributors to change the translation of the word "Spaces" in Russian to more human.
From "Пространство" to "Сообщество".
I realize, that all of us are technical specialists and we are used to different terms, but for the average people, normal users of social network, the website has to be more clear.
Please let me know, if you think if it's reasonable or not, so I can change the rest of files.
Thank you!

luke-:
@devochka Maybe this is a topic which should be discussed at http://translate.humhub.org in the Russian language space (RU). At the moment we prefer also in German the word "Space" as it is. Maybe we find a solution in future to make the wording configurable.

PR #2075, opened 2016-11-03
Change icon to a more explicit one (trash)

acs-ferreira:
Moves from a circle with cross (similar to "Close") to a Trash icon (similar to "Delete/Remove").


buddha87:
Hmm, since the x is used as delete symbol all over the applicatio I won't change it here. And personally I think the circle symbol looks a bit better in this case.

acs-ferreira:
No problem!

PR #2077, opened 2016-11-04
Load JS on HEAD instead of BODY

acs-ferreira:
Since the `<head>` loads before the `<body>`, placing JS in the <head> ensures that it is available when needed.


luke-:
We're heavily improved the Javascript part (combine, compress assets) in the 1.2 release.
There may to many side effects to change this in the stable branch.

PR #2078, opened 2016-11-04
Load .min. versions of jQuery and Bootstrap Assets

acs-ferreira:
This way in `Development` environment you'll have the unminified version, but in `Production` Yii2 will use the minified file.


buddha87:
Thanks, but in 1.2 the asset management changes anyways, all core scripts will be compressed and served as single file in production.

acs-ferreira:
Good news :)

PR #2086, opened 2016-11-10
Update grunt-contrib-cssmin

coliff:
Update grunt-contrib-cssmin to 1.0.2


buddha87:
This is not required since '^1.0.1' should include the latest version in the same major release https://semver.npmjs.com/

PR #2087, opened 2016-11-10
Fix capitalisation of YouTube

coliff:
Fix capitalisation of YouTube on Profile > Social Bookmarks page


Cara-W:
This does not match with humhub/protected/humhub/modules/installer/libs/InitialData.php line 375.

`        $field->title = 'Youtube URL';`

The translation does not work for humhub/protected/humhub/modules/user/messages/_language_code_/models_Profile.php due to the change.

PR #2089, opened 2016-11-10
Fixes #2085

### LINKED ISSUE #2085: register-email input type should be email
The input textbox for '`register-email`' is currently '`text`', it would be good to change that to '`email`' (to take advantage of browser-based email validation as well as bringing up the email keyboard on mobile devices).

acs-ferreira:


cebe:
seems to be fine especially because of:

> Browsers without support for these types will fall back to using the "text" type.
> http://caniuse.com/#feat=input-email-tel-url

PR #2090, opened 2016-11-10
Update jquery.nicescroll.min.js

coliff:
Update to latest version from
https://github.com/inuyaksa/jquery.nicescroll/blob/master/dist/jquery.nicescroll.min.js


buddha87:
Please note the following: https://github.com/humhub/humhub/blob/master/protected/humhub/assets/AppAsset.php#L54

The issue seems not to be fixed yet: https://github.com/inuyaksa/jquery.nicescroll/issues/574

PR #2093, opened 2016-11-11
Fixes #2091

### LINKED ISSUE #2091: Posts images are being served as dynamic content, not using browser cache
I'm running Humhub on a wireless network (city size), our (self) hosting and bandwith is reaching its limits, we need to squeeze every little bit of performance.

Deployed last friday, already have more than 700 users, 80-160 concurrent, people love it.

Current setup is nginx, php-fpm, mariadb and redis for cache and sessions.

I just realized images are being served as dynamic content, not directly from disk, that is a huge performance blow, if nginx could serve those images directly the server would fly.

Don't know how rooted is this behavior to your architecture but seriously, this could be that low hanging fruit which could fire your scalability to the clouds.

See here, every time I hit refresh the content images are being re-downloaded.

pic34

acs-ferreira:
Add `HttpCache` filter to `FileController` behavior, returns `304` since last post update.

### DIFF
--- a/protected/humhub/modules/file/controllers/FileController.php
+++ b/protected/humhub/modules/file/controllers/FileController.php
@@ -35,9 +35,8 @@ class FileController extends \humhub\components\Controller
             ],
             'httpCache' => [
                 'class' => 'yii\filters\HttpCache',
-                'lastModified' => function ($action, $params) {
-                    $q = new \yii\db\Query();
-                    return $q->from('post')->max('updated_at');
+                'etagSeed' => function ($action, $params) {
+                    return serialize([\yii\helpers\Url::current()]);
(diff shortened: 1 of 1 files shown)


acs-ferreira:
I've submitted this to `v1.2-dev`, feel free to fork it to `master`

buddha87:
Thanks, I think the lastmodified query is not neccessary, since the files do not change anyways. The the Cache-Control should be sufficient (www.yiiframework.com/doc-2.0/yii-filters-httpcache.html#$cacheControlHeader-detail).

acs-ferreira:
Using `ETag` instead of `Last-Modified` seems more efficient.

PR #2094, opened 2016-11-12
Attached files browser caching, fixes #2091

### LINKED ISSUE #2091: Posts images are being served as dynamic content, not using browser cache

acs-ferreira:
Using ETag and max-age

### DIFF
--- a/protected/humhub/modules/file/controllers/FileController.php
+++ b/protected/humhub/modules/file/controllers/FileController.php
@@ -34,7 +34,14 @@ class FileController extends \humhub\components\Controller
             'acl' => [
                 'class' => \humhub\components\behaviors\AccessControl::className(),
                 'guestAllowedActions' => ['download']
-            ]
+            ],
+            'httpCache' => [
+                'class' => 'yii\filters\HttpCache',
+                'cacheControlHeader' => 'public, max-age=31536000',
+                'etagSeed' => function ($action, $params) {
+                    return serialize([\yii\helpers\Url::current()]);
+                },
+            ],
         ];
     }
 


buddha87:
The etag won't be neccessary either, since the files do not change and in your implementation you'll return the same etag for a file anyways. I think we just should set the 'cacheControlHeader' to a reasonable value, a year seems to be an overkill.

I did some tests with cacheControlHeade only and this had no effect, somehow the beforeAction of yii\filters\HttpCache returns before applying the header in this case.

I added the following line to my humhub\modules\file\actions\DownloadAction (without the use of a filter) and it worked:

`Yii::$app->response->headers->add('Cache-Control', 'public, max-age=604800');`

Either we use this approach or add something like this to the HttpCache filter:

`'lastModified' => function ($action, $params) {
        return time();
    },`

acs-ferreira:
IMO, my approach is the best one, because:

- it set a never expire header with the `max-age = 1 year`, see https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html

> To mark a response as "never expires," an origin server sends an Expires date approximately one year from the time the response is sent.

- it generates a unique ETag for each file, it remains the same ETag even after cache clearing

But if in your opinion your approach is better, i'm just contributing after all, not owning ;)

buddha87:
Yeah I understand the purpose of the etag and the max-age. I just tought the one year caching is a bit much, but I think its ok. Thanks for the pr ;)

buddha87:
We'll merge this after the 1.1.2 release.

Issue #2008, opened 2016-10-01
Problems Upload Media


carbans:
I have a problem when I upload media. Disappear the button to send message.

PD: Attach a capture of screen
![captura-humhub](https://cloud.githubusercontent.com/assets/5936829/19014444/0eaa7e8e-87ed-11e6-8bf4-8c3f53877ac8.png)

buddha87:
please check for javascript errors in the developer console of your browser (Ctrl + Shift +i)

acs-ferreira:
Same issue here, either the action (Submit buttons) disappear or the Upload just **don't start** after selecting the file to upload (PDF); no upload ba, no network activity, nothing..

Console is completely empty, tried on PHP 5.6 and 7.0.

acs-ferreira:
Also, sometimes i edit and upload a file to the post, nothing happens.
Save the post, no file appears on the post. 

I then do `CTRL+F5` and the file appears after page refresh.

acs-ferreira:
Is this line really necessary?
I've removed it to make some tests and my problems are all gone. I think there is no need to hide the container.

https://github.com/humhub/humhub/blob/3bcc3e67addbe61852c625891f593dc83b3655e3/protected/humhub/modules/content/widgets/views/wallCreateContentForm.php#L87

PR #2096, opened 2016-11-12
Don't hide the buttons container and fix remove focus

acs-ferreira:
See #2008, i think there is no need to hide it in any circumstances.
Also fix remove focus.

My opinion, discussion opened :)


acs-ferreira:
Ok, the container should be hided when upload bar show up...or sometimes the bar doesn't show up.

2 solutions:

- add a simple margin to the bar and show it under the buttons
- or, **better solution**, find why it doesn't show every time. Investigating...

buddha87:
The wallcreateform is rewritten in v1.2 especially the js part:

https://github.com/humhub/humhub/blob/v1.2-dev/protected/humhub/modules/content/assets/js/humhub.content.form.js

Don't know if this is still an issue, I have never seen this behaviour.

acs-ferreira:
Ok, i'll test it.

PR #2103, opened 2016-11-15
Losslessy optimised images with ImageOptim

coliff:
Saved 103.8KB out of 345.2KB. 46.2% per file on average (up to 94.9%)


coliff:
I removed the glyphicons-halflings PNG images as those weren't being used for anything (they were included with Bootstrap 2 which isn't being used in this project). I had meant to remove those in a separate pull request to the image optimisation one though, hope that hasn't caused any issues.

buddha87:
Thanks, we'll probably merge this with v1.2.

PR #2113, opened 2016-11-23
Prepare for PHP 7.1 Release

acs-ferreira:
Actually on its final RC, PHP 7.1 final will be released in the next few weeks.
https://github.com/php/php-src/blob/php-7.1.0RC6/UPGRADING


acs-ferreira:
Maybe switch this to [v1.2-dev branch](https://github.com/humhub/humhub/tree/v1.2-dev)

PR #2127, opened 2016-12-01
Fix(?) for Like event notification

wookba:
Like module don't send event to users once their post/reply has been liked. Culprit is "if ($this->getSource()->hasAttribute('content') && $this->getSource()->content->createdBy !== null)"
I tried to workaround here but couldn't find any except removing condition entirely.

### DIFF
--- a/protected/humhub/modules/like/models/Like.php
+++ b/protected/humhub/modules/like/models/Like.php
@@ -85,12 +85,10 @@ class Like extends ContentAddonActiveRecord
         $activity->create();
         
         // source itsself does not need to have creadedBy attribute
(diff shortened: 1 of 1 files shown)


buddha87:
Hi, I think this was removed in a previous version since it led to a flood of mails in some cases. We'll have to find a solution for this, perhaps make it configurable.

wookba:
@buddha87 I agree, but people do like to have notifications about likes (in some cases).
Maybe to send an e-mail only 30 minutes after last new like with number of likes or something...

luke-:
@wookba At the moment we're improving the notification part for the 1.2 version - we will add a possiblity to disable particular notifications and in later step also to delay and combine similar.

PR #2184, opened 2017-01-02
Update ZendLdapClient.php

### LINKED ISSUE #1961: Urgent problem signing in with LDAP user after upgrading to 1.1
Hi Humhub team. We've recently updated to 1.1. Great work on fixing the LDAP bug where users couldn't log in with their email addresses! But it looks like the new ZendLDAPClient auth library may have caused some more problems, at least with older users. Whenever a user logs in, they get the error "Could not automatically determine unique user ID from LDAP node" from line 195 in ./protected/humhub/modules/user/authclient/ZendLdapClient.php.

It looks to me like this is caused by code which checks the LDAP $attributes array within the getUserAttributes function which returns an aray list of user attributes. It seems to be checking for the existence of a key named objectguid, but none exists on the users I've tried with. There _is_ a key named entryuuid which seems to hold the Humhub's user UUID. All of the other attributes seem to be populated without problem. I did try changing the code to also accept the 'entryuuid' key as a substitute for 'objectguid', but without any luck. What am I doing wrong?

Thanks so much in advance,

Daniel

`
            if (isset($attributes['objectguid'])) {
                $this->idAttribute = 'objectguid';
                $this->userTableIdAttribute = 'guid';
            } /*elseif (isset($attributes['entryuuid'])) {
                $this->idAttribute = 'entryuuid';
                $this->userTableIdAttribute = 'guid';
            } */elseif (isset($attributes['mail'])) {
                $this->idAttribute = 'mail';
                $this->userTableIdAttribute = 'email';
            } else {
                throw new \yii\base\Exception("Could not automatically determine unique user id from ldap node!");
            }
`

Felli:
Should fix #1961

### DIFF
--- a/protected/humhub/modules/user/authclient/ZendLdapClient.php
+++ b/protected/humhub/modules/user/authclient/ZendLdapClient.php
@@ -182,7 +182,7 @@ class ZendLdapClient extends BaseFormAuth implements interfaces\AutoSyncUsers, i
             } elseif (isset($attributes['mail'])) {
                 $this->idAttribute = 'mail';
                 $this->userTableIdAttribute = 'email';
-            } else {
+            } elseif (isset($attributes['uid'])) { $this->idAttribute = 'uid'; $this->userTableIdAttribute = 'username';
                 throw new \yii\base\Exception("Could not automatically determine unique user id from ldap node!");
             }
         }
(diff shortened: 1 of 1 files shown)


Felli:
Closing because was fixed.

PR #2190, opened 2017-01-07
Adding more french translations

Becold:
I tried translate.humhub.org but the FR espace seems inactive so I created this PR. Feel free to merge or close it.

And I can't find these translations in the files : 
image


gadgondhali:
I want to add Marathi language translation as same as you add french translation so please guide me how can I do that. and also where to do this work

PR #2198, opened 2017-01-11
FancyIndexing

Felli:
Hides the index.php from url view unless directed to /index.php


acs-ferreira:
There is no need for this if you use https://github.com/humhub/humhub/pull/2095

001101:
There is use, if you using modern web servers like nginx.

acs-ferreira:
In that case you are not using .htaccess

Le 25 janv. 2017 00:38, à 00:38, 001101 <notifications@github.com> a écrit:

PR #2225, opened 2017-01-26
Admin User Updates E-Mail Notifications

denntl:
Created easy E-Mail notifications to all admins about users changes: Comments, New Content, Likes, Following.

### DIFF
--- a/protected/humhub/modules/notification/widgets/views/notificationSettingsForm.php
+++ b/protected/humhub/modules/notification/widgets/views/notificationSettingsForm.php
@@ -2,7 +2,7 @@
 /* @var $model \humhub\modules\notification\models\forms\NotificationSettings */
 /* @var $form yii\widgets\ActiveForm */
 
-use yii\bootstrap\Html
+use yii\bootstrap\Html;
 ?>
 
 <br />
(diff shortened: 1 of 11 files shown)


protected/humhub/modules/notification/widgets/views/notificationSettingsForm.php
```diff
@@ -39,8 +39,10 @@
                     </td>
                     <?php foreach ($model->targets() as $target): ?>
                         <td class="text-center">
-                            <?php $disabled = !$target->isEditable($model->user) || $category->isFixedSetting($target) ?>
-                            <?= Html::checkbox($model->getSettingFormname($category, $target), $target->isCategoryEnabled($category, $model->user), ['style' => 'margin:0px;', 'disabled' => $disabled]) ?>
+                            <?php if(! ($target->id == 'web' && $category->id == 'admin_user_updates')): ?>
```
buddh4: You should be able to use the isVisible(User $user) function instead of this if statement. Just filter out all non admins in the isVisible function.

buddh4:
Thanks, we'll discuss this feature.

Issue #1786, opened 2016-06-08
Question: invalid datetime 


MatthiasHertel:
i got this issue by installing ... 

what is the problem ?

![image](https://cloud.githubusercontent.com/assets/8426566/15911576/b88cf7f0-2dd0-11e6-8516-745fda200848.png)

MatthiasHertel:
ok the issue appears only by installing the task module ...
maybe a issue of the package ?

buddha87:
seems to be an issue with the sql_mode of your db. Which mysql version are you using ?

MatthiasHertel:
```bash
vagrant@vagrant:~$ mysql --version
mysql  Ver 14.14 Distrib 5.7.12, for Linux (x86_64) using  EditLine wrapper
```

MatthiasHertel:
i have a little bit specify this issue - i hope it helps

### in short way: 
ubuntu/trusty64 14.04 **works**
ubuntu/xenial64 16.04 **failed**

### appendix:
the vagrantfile and provisioning bootstrap files of both install versions

### notes
i know that i have to do some settings manually after vagrant up but that is not the problem
the problem is that i dont know which sql mode i should use

**1.** _create table and set db user_
```mysql
CREATE DATABASE `humhub` CHARACTER SET utf8 COLLATE utf8_general_ci;
GRANT ALL ON `humhub`.* TO `humhub_dbuser`@localhost IDENTIFIED BY 'password_changeme';
FLUSH PRIVILEGES;
```
**2.** _fetch dependencies via composer_
```bash
php composer.phar global require "fxp/composer-asset-plugin:~1.1.1"
composer update
```

### 14.04

**Vagrantfile**
```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

# Vagrantfile API/syntax version. Don't touch unless you know what you're doing!
VAGRANTFILE_API_VERSION = "2"

Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|

  # Every Vagrant virtual environment requires a box to build off of.
  # config.vm.box = "geerlingguy/ubuntu1604"
  # config.vm.box = "ubuntu/xenial64"
  config.vm.box = "ubuntu/trusty64"

  # Create a private network, which allows host-only access to the machine using a specific IP.
  config.vm.network "private_network", ip: "192.168.33.22"
  config.vm.network "forwarded_port", guest: 80, host: 8080

[...]

Issue #1777, opened 2016-06-06
Support SQL_Mode: ONLY_FULL_GROUP_BY


nejcneo:
https://i.snag.gy/YHwF3c.jpg
That is what i got after installing In Requirements I got only 2 alerts both are yellow type.
After that I go all trough and after login I hit this error.
![image](https://cloud.githubusercontent.com/assets/1667825/15820229/6b9a7286-2be8-11e6-90a7-bf31332dddcf.png)

luke-:
You need to turn off the sql_mode ONLY_FULL_GROUP_BY. At the moment HumHub doesn't support this mode.

michaelgeisler:
how to turn off this mode?

buddha87:
Set the follwoing value in your my.ini:

`sql_mode=TRADITIONAL`

PR #2236, opened 2017-02-04
workaround for #1777 and #1786

danielkesselberg:
I setup a local humhub copy today on my vagrant box (ubuntu 16.04, php7, mysql5.7). The initial setup was a bit frustrating ;-) First setup crashed when task-module enabled (#1786) and later with (#1777).  

Both problems may solved if mysql is running with sql_mode traditional. I can change this on my vagrant box but not on a shared hosting. 

I ended up with adding a small helper function which changes the sql_mode for the current session. Then changed the db configuration to run this callback after a successfull mysql connection.

This is not a real solution for the issues but a neat workaround until all sql strict mode related issues are solved.

### DIFF
--- a/protected/humhub/libs/Helpers.php
+++ b/protected/humhub/libs/Helpers.php
@@ -213,7 +213,7 @@ class Helpers
     {
         /* set sql_mode only for mysql */
         if ($event->sender->driverName == 'mysql') {
-            $event->sender->createCommand('SET sql_mode="TRADITIONAL"')->execute();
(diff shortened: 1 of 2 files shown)


danielkesselberg:
#1777 #1786

001101:
would it be possible that you write which version/branch you did installed?

danielkesselberg:
Hey, i used the latest master branch at this time.

luke-:
Danke! 

#1786 We should fix the migration failing
#1777 should be fixed by: https://github.com/humhub/humhub/pull/2338

PR #2238, opened 2017-02-06
Added option to use APCu instead of APC

SpaceCafe:
screenshot apcu

### DIFF
--- a/protected/humhub/libs/DynamicConfig.php
+++ b/protected/humhub/libs/DynamicConfig.php
@@ -109,9 +109,11 @@ class DynamicConfig extends \yii\base\Object
 
         // Add Caching
         $cacheClass = Yii::$app->settings->get('cache.class');
+        $cacheUseApcu = Yii::$app->settings->get('cache.useApcu');
         if (in_array($cacheClass, ['yii\caching\DummyCache', 'yii\caching\ApcCache', 'yii\caching\FileCache'])) {
             $config['components']['cache'] = [
                 'class' => $cacheClass,
+                'useApcu' => $cacheUseApcu,
                 'keyPrefix' => Yii::$app->id
             ];
         }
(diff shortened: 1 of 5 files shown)


protected/humhub/tests/codeception/fixtures/data/setting.php
```diff
@@ -17,6 +17,7 @@
     array('name' => 'maxFileSize', 'value' => '1048576', 'module_id' => 'file'),
     array('name' => 'forbiddenExtensions', 'value' => 'exe', 'module_id' => 'file'),
     array('name' => 'cache.class', 'value' => 'CFileCache', 'module_id' => 'base'),
+    array('name' => 'cache.useApcu', 'value' => '0', 'module_id' => 'base'),
```
Felli: Shouldn't this also work? Just trying to understand why `0` would be the better choice.
`array('name' => 'cache.useApcu', 'value' => 'CFileCache', 'module_id' => 'base'),`
Or this should also work being Yii2.
`array('name' => 'cache.useApcu', 'value' => 'FileCache', 'module_id' => 'base'),`

protected/humhub/libs/DynamicConfig.php
```diff
@@ -109,9 +109,11 @@ public static function rewrite()
 
         // Add Caching
         $cacheClass = Yii::$app->settings->get('cache.class');
+        $cacheUseApcu = Yii::$app->settings->get('cache.useApcu');
         if (in_array($cacheClass, ['yii\caching\DummyCache', 'yii\caching\ApcCache', 'yii\caching\FileCache'])) {
             $config['components']['cache'] = [
                 'class' => $cacheClass,
+                'useApcu' => $cacheUseApcu,
```
cebe: this will break FileCache and DummyCache as they do not have this property.

Felli: Hmm... Would `'class' => $cacheUseApcu,` work for this?

cebe: no, that does not make sense, the if needs to have different branches, one for Apc and one for DummyCache and FileCache.

```php
         if (in_array($cacheClass, ['yii\caching\DummyCache', 'yii\caching\FileCache'])) {
             $config['components']['cache'] = [
                 'class' => $cacheClass,
                 'keyPrefix' => Yii::$app->id
             ];
         } elseif ($cacheClass === 'yii\caching\ApcCache') {
             $config['components']['cache'] = [
                 'class' => $cacheClass,
                 'useApcu' => $cacheUseApcu,
                 'keyPrefix' => Yii::$app->id
             ];
         }
```

luke-:
Thanks! We decided to automatically switch to APCu when available.

PR #2241, opened 2017-02-08
Correct PHP Version hint

benmag:


cebe:
composer.json also needs adjustment: https://github.com/humhub/humhub/blob/04918306707a2b5704ab29deefe1d668106eaf2f/composer.json#L17

PR #2244, opened 2017-02-08
Fix: Broken Link #1

Felli:
Should fix issues with broken links within the docs.


luke-:
Thank you! Please use relative links.

e.g.    ../theme/migrations.md

luke-:
Thank you and sorry for the late reply - we completely refactored the theme guide in the meantime.

http://docs.humhub.org/theme-README.html

Issue #2169, opened 2016-12-21
1.2 dev installation error


soju7:
http://prntscr.com/dm5py4
Trying to install humhub1.2 dev .unfortunately i got **SQLSTATE[HY000] [1044] Access denied for user ''@'localhost' to database 'humhub'.**  why this happen? . i can able to install humhub 1.1.2  with no errors. Thanks in advance.

001101:
cause your password is wrong??!

ajilmv138:
I am also stuck in the same issue. When I am trying to install humhub1.2 dev, first page shows this error message.  "SQLSTATE[HY000] [1044] Access denied for user ''@'localhost' to database 'humhub'". How can i solve this. please help me. Thanks in advance

soju7:
no actually the error is from protected/humhub/config/common.php .
http://prntscr.com/doe69e
this happen during first step.
 http://localhost/humhubv1.2/index.php?r=installer%2Findex
i never had this issue on other versions installation.
pls help

buddh4:
Seems to be an issue with the new View logic.

soju7:
still got the same error after pulling new changes

buddh4:
Sorry there was a file missing in the commit.

spsinghats:
I have both these commits and I am having the same issue. Tried, v.1.2.0-beta.1 and the master as on 23 Feb 2017.

PR #2292, opened 2017-02-23
Fix the installer issue #2169

spsinghats:
Disable flash checking for installer


luke-:
Thank you, this should be already fixed - see if statement above

spsinghats:
Doesn't work with the above if statement. I had to add this code to make it work

PR #2318, opened 2017-03-02
Fix small gap on TopNav mouse hover

acs-ferreira:
Before:

After:

### DIFF
--- a/static/less/topbar.less
+++ b/static/less/topbar.less
@@ -274,6 +274,7 @@
     text-shadow: none;
     font-weight: 600;
     font-size: 10px;
+    min-height: 50px;
     text-transform: uppercase;
     text-align: center;
     &:hover,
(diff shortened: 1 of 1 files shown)


buddh4:
The removal of min-height destroys the space chooser dropdown in my test. The gap problem only occures after zooming the page am I right?.

acs-ferreira:
No, you don't need to zoom the page.


What you mean by "space chooser dropdown"?

buddh4:
The following happens on firefox (on chrome it seems to be ok):

topbarpr

acs-ferreira:
Ah ok, i see, you're right.
Can you test latest commit?

acs-ferreira:
Ok, let me do some tests on my side and i'll be back.

acs-ferreira:
Things should be OK now, can you test on your side please?
Tested here against Firefox, Chrome and IE Edge.

buddh4:
thanks!

PR #2325, opened 2017-03-03
- 16Mb trash

githubjeka:
From /static/assets


buddh4:
Unfortunately those assets are currently required for running in production mode.

githubjeka:
I guess Ok, thanks. I don't try prod mode yet :)

001101:
i only  can see trash in this post

PR #2332, opened 2017-03-04
Human readable file sizes

acs-ferreira:
Moves from `237.5167356 kilobytes` to `237.5 KB`


protected/humhub/modules/file/libs/FileHelper.php
```diff
@@ -119,7 +119,7 @@ public static function getFileInfos(File $file)
             'size' => $file->size,
             'mimeType' => $file->mime_type,
             'mimeIcon' => MimeHelper::getMimeIconClassByExtension(self::getExtension($file->file_name)),
-            'size_format' => Yii::$app->formatter->asSize($file->size),
+            'size_format' => Yii::$app->formatter->asShortSize($file->size, $decimals = 1),
```
cebe: just pass 1 as the second parameter instead of defining the $decimals variable.

acs-ferreira: True, done :)

PR #2353, opened 2017-03-14
Keep local configuration

acs-ferreira:


luke-:
We can't simply ignore the 'protected/config/' files.

If we add this to ignore, we also need to check the existence of protected/config/common.php,console.php,web.php

As workaround you can use "git assume unchanged" for these files.

PR #2386, opened 2017-03-21
Spring Code Cleaning

acs-ferreira:
Clean, sanitazing, optimize lisibility.


githubjeka:
Such PR like this is very hard to merge, because it contains a lot of changes. In the future, it is better to send 100 minor edits than one such large one.

luke-:
I really like this change and thank you for your work - but we can't merge it, because it's impossible to check and validate.

Can you please separate this at least by module (e.g. user module cleanup)?

We're also near completion of 1.2...

Sorry!

PR #2407, opened 2017-03-26
Activity Module Clean

acs-ferreira:


luke-:
Great, we need to be careful with formatting of plaintext e-mail templates.
I'll fix this issue.

protected/humhub/modules/activity/views/activityEntry.php
```diff
@@ -1,5 +1,10 @@
+<?php
+/**
+ * @link https://www.humhub.org/
+ * @copyright Copyright (c) 2017 HumHub GmbH & Co. KG
+ * @license https://www.humhub.com/licences
+ */
+
```
luke-: Oooups

PR #2409, opened 2017-03-27
Admin & Comment Modules Clean

acs-ferreira:


luke-:
First of all, thank you very much for the cleanups!

But please try to reduce the changed files per PR and better check changed files.
It's very hard for us to manually check and fix 143 files at once.

Required fixes:
https://github.com/humhub/humhub/commit/ac5de195a31f7227a762aeec0200a93c860fac3d

Also we're aiming to publish the 1.2.0 stable version in the next week - so we may should delay further "mass" cleanups until 1.2.1.

PR #2417, opened 2017-03-28
remove reinitialize stream

githubjeka:
INIT called in the constructor.
```php
//from yii code of controller
$action = $this->createAction($id); // this INIT by default
$result = $action->runWithParams($params); // This you again call INIT. Why?
```


luke-:
Oups

PR #2434, opened 2017-04-03
Fix: Misplacement of both Likes & Comments

Felli:


buddh4:
What exactly is misplaced, do you have a screenshot?

Felli:
Within custom themes the comment button will always come before the like but, even with the Wiki module, with this it will force the like button to come first.

acs-ferreira:
I've tried this _(trough Firebug)_ here is what i get:

image

Also, `Comment` is always the first link, even in HumHub default theme, am i wrong?

Felli:
As @acs-ferreira said; Comments are seen as first within an installation of the software, using this does provide a fix, I've tested myself on my personal servers with success. The only issue that I face now is fixing the like a comment...
untitled

acs-ferreira:
Ok, i see @Felli , but i prefer to have **Comments first** ;)
_Personal opinion, nothing else. Is this is merged, i'll revert it locally._

buddh4:
So, I close this for now, I don't see this problem in the community theme?

Felli:
I suspect this is an issue that happens for self hosted sites.

PR #2497, opened 2017-04-19
The notification ContentCreated can be disabled

githubjeka:
Now when created any new content the followers get notify about this.
In my case I want create draft content model without this notify.

```php
$issueModel = new Issue(['status' => 'DRAFT']);
$issueModel->content->setContainer($container);
$issueModel->save();
```
I add new mechanism to regulation of this behavior and I described about new feature in documentation for the attribute `\humhub\modules\content\models\Content::$notifyUsersOfNewContent` .

To can write code:

```php
$issueModel = new Issue(['status' => 'DRAFT']);
$issueModel->content->notifyUsersOfNewContent = false;
$issueModel->content->setContainer($container);
$issueModel->save();
```

P.S. In a right way, "Notify of ContentActiveRecord" had to be attached through Yii2 Events and then could be off method of yii/base/Event without this PR, but it would have drawn a big change. Therefore, while there will be so. :)


acs-ferreira:
Any news about this? @buddh4 @luke-

buddh4:
Currently ther is the [muteDefaultSocialActivities ](https://github.com/githubjeka/humhub/blob/4f25cff0da78f9d208a0f5e62fc1f9735d05fe5f/protected/humhub/modules/content/models/Content.php#L80) flag in the `Content` class, but I like the idea of @githubjeka a bit more, since we can manage the behaviour within the `ContentActiveRecord` without setting the content flag manually. The `muteDefaultSocialActivities` is currently not present in a stable release, so we can replace it with the given solution before the v1.2.3 release.

githubjeka:
Your new `$muteDefaultSocialActivities` in master branch fixed my issue.

githubjeka:
Thanks

PR #2500, opened 2017-04-21
Disable DEBUG MODE as default

acs-ferreira:
Actually, DEBUG MODE is on as default, and user should manually disable it before go to production.

Here, it sets the DEBUG MODE as ON if application is accessed from local IPS _(or a manually added IP to the array)_, and sets it OFF as default.

IMO, this should be the default:
- OFF as default,
- ON on localhost, 
- ON if the user manually requests it

> Documentation should be updated **ONLY IF** this PR is accepted: https://www.humhub.org/docs/guide-admin-installation.html#disable-errors-debugging

### DIFF
--- a/index.php
+++ b/index.php
@@ -6,13 +6,22 @@
  * @license https://www.humhub.com/licences
  */
 
-// comment out the following two lines when deployed to production
-defined('YII_DEBUG') or define('YII_DEBUG', true);
-defined('YII_ENV') or define('YII_ENV', 'dev');
+// IPs to access HumHub on DEBUG/DEV mode
+$debugmode = [
+    '127.0.0.1',
+    '::1',
+    //'YOUR_IP_HERE'
+];
+
+// DEBUG and DEV are ON as defaut when accessing HumHub from local IPs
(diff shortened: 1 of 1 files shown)


githubjeka:
To better fix it need add new console command to fast switch dev/prod environment like in Advanced template Yii2. And this check in_array for every request is bad IMHO

michelem09:
There is no documentation how to disable debug mode, please add it. I waste time to look at files to find it.

acs-ferreira:
@michelem09 http://docs.humhub.org/admin-security.html#disable-errors-debugging

michelem09:
Well, if you try to search it (form on top right corner) that page doesn't come up

acs-ferreira:
It comes, but not very specific...
image

michelem09:
Not if you put only "debug"

screen shot 2017-09-15 at 11 55 09

githubjeka:
http://www.yiiframework.com/doc-2.0/guide-tutorial-performance-tuning.html#disable-debug

PR #2546, opened 2017-05-05
Enh: Remove success flash messages

githubjeka:
From https://community.humhub.com/content/perma?id=105301

If you agree with this then I will continue to send new commits about it in this PR.


buddh4:
In the stream etc, I think it's ok to remove the messages, but we should keep them in forms and everywhere where you don't have another visible feedback. Another option in case this message annoys a bit, would be reducing the duration of the success message.

PR #2630, opened 2017-06-17
Add Fullscreen for apple and android

Joe2824:
add apple and android fullscreen for web app

### DIFF
--- a/themes/HumHub/views/layouts/head.php
+++ b/themes/HumHub/views/layouts/head.php
@@ -20,3 +20,13 @@
 <meta name="msapplication-TileImage" content="<?= $this->theme->getBaseUrl(); ?>/ico/ms-icon-144x144.png">
 <meta name="theme-color" content="#ffffff">
 <!-- end: Favicon and Touch Icons -->
+
+<!-- start: Apple Fullscreen and Webapp Title -->
+<meta name="apple-mobile-web-app-title" content="HumHub" />
+<meta name="apple-mobile-web-app-capable" content="yes" />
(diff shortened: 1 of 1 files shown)


Felli:
You you explain why this is needed?

Joe2824:
With these lines HumHub use the standalone mode to look more like a native app. For iOS Safari there is no browser URL Textfield at the top of the screen or button bar. Only the status bar appears in black on the screen.
The mobile-web-app-capable is for Android and have the same use.

img_0109
img_0108

Felli:
Thanks for giving your opinion, as I'm not in control of any of HumHub's repositories, I'll look into this and talk with the Development team about adding this. 👍

themes/HumHub/views/layouts/head.php
```diff
@@ -20,3 +20,13 @@
 <meta name="msapplication-TileImage" content="<?= $this->theme->getBaseUrl(); ?>/ico/ms-icon-144x144.png">
 <meta name="theme-color" content="#ffffff">
 <!-- end: Favicon and Touch Icons -->
+
+<!-- start: Apple Fullscreen and Webapp Title -->
+<meta name="apple-mobile-web-app-title" content="HumHub" />
```
buddh4: Should rather contain `Html::encode(Yii::$app->name)` instead of `humhub`

buddh4:
Thanks https://github.com/humhub/humhub/commit/d40f1b12efca8188d256c75061dcd06acab08338

PR #2642, opened 2017-06-26
undefined var fix in JSONResponse.php

artturik:


Felli:
Isn't a 100% fix, more of a masker and could cause more issues.

buddh4:
Think this class is not in use anymore yii now supports `$this->asJson();` But thanks for the fix.

PR #2645, opened 2017-06-28
Update theme.css

hagalaz:


buddh4:
Thanks for the PR, but the css file should not be edited directly (of course in your own theme you can handle it as you wish) I've added your change to https://github.com/humhub/humhub/commit/1353c263cad0ee6737ae7f22189605c535e4ae31#diff-db128232595ebf54cd56c722da5aeadb

PR #2649, opened 2017-07-03
Fix: Fix for minor Firefox issues

Felli:
Should be looked over first before merging just in case I've missed or misspelled something.


buddh4:
Don't edit the css file directly. Use the less files under https://github.com/humhub/humhub/tree/master/static/less

PR #2676, opened 2017-07-26
Fixes yii\base\ErrorException: Trying to get property of non-object.

githubjeka:


buddh4:
Can you provide a scenario, where this error occures the WallEntry base class should already provide a `jsWidget` js class.

githubjeka:
See your code https://github.com/githubjeka/humhub/blob/167692a7125278bae4815d4c8981e6878de61074/protected/humhub/modules/content/components/ContentActiveRecord.php#L281-L295

protected/humhub/modules/stream/actions/Stream.php
```diff
@@ -295,17 +295,26 @@ protected function isInitialRequest()
      */
     public static function renderEntry(ContentActiveRecord $record, $options =  [], $partial = true)
     {
+        // TODO should be removed in next major version
         // Compatibility with pre 1.2.2
         if(is_bool($options)) {
             $partial = $options;
             $options = [];
         }
```
buddh4: How about just adding

```
if(!$record->wallEntryClass) {
 return "";
}
```

And in v1.3 we should get rid of the static `@humhub/modules/content/views/layouts/wallEntry` and integrate the div into `@humhub/modules/content/widgets/views/wallEntry.php`, so the WallEntry widget is more flexible and not dependend on the `$jsWidget` etc.

githubjeka:
Done

PR #2681, opened 2017-07-27
Changed 'Add an user' to 'Add a user'

JavaDevVictoria:
'Add a user' is the correct form in English. Use of 'a' or 'an' are not determined by spelling, but pronunciation. Any word which sounds like it begins with a vowel (it may or may not actually begin with a vowel) will take 'an' and any word which begins with sound of a consonant will take 'a' as the preceding article.


buddh4:
This userpicker logic is deprecated, but thanks for the fix ;)

PR #2691, opened 2017-08-01
remove strtolower($username)

### LINKED ISSUE #2613: username encoding with pretty url
If you enable pretty url in configuration (web.php or common.php, inside protected/config) :

> 'components' =>[
>   'urlManager' => [
>             'enablePrettyUrl' => true,
>         ],
> ],

Then the user module (and space), creates the URLs like this: 
    http://localhost/index.php/u/username/user/profile/home
The problem here is the urlencode. The user module UrlRule does this:
`$url = "u/" . urlencode(strtolower($username)) . "/" . $route;`

So when the $username have special characters, strtolower() returns some ilegal characters, and so the urlencode() returns a bad encoding, which throws "page not found" errors anytime you try to go to this malformed URLs. You cannot go to "My profile" for example.
Just for note, when you log in through linkedin, many countries have special characters in their names, and these name becomes the username.

Anyway I've fixed this, just replace the line of code with the following:
`$url = "u/" . urlencode(mb_strtolower($username)) . "/" . $route;`

githubjeka:
To fix #2613

### DIFF
--- a/protected/humhub/modules/user/components/UrlRule.php
+++ b/protected/humhub/modules/user/components/UrlRule.php
@@ -44,7 +44,7 @@ class UrlRule extends Object implements UrlRuleInterface
                     $route = "";
                 }
 
-                $url = "u/" . urlencode(strtolower($username)) . "/" . $route;
+                $url = "u/" . urlencode($username) . "/" . $route;
                 if (!empty($params) && ($query = http_build_query($params)) !== '') {
                     $url .= '?' . $query;
                 }


Felli:
This does work, but doesn't help that it is also within the account dropdown menu as well. :+1:

githubjeka:
Where? Screenshot please

Buliwyfa:
I found some servers does not allow the + on username.
Any work around?

githubjeka:
`+` will be converted to `%2B`

Felli:
I even cleared cache from site, server, and browser, but didn't work 100%, did work however on wall posts with user thumbnail.
screenshot_1

githubjeka:
Something strange with your local enviroment

Felli:
Using a public CloudLinux (CentOS) 6 and a private one, both this doesn't work sadly, I would love to have this work as it makes likes too big for profiles, so if you can get this working thanks! :+1:

Buliwyfa:
Some servers does not allow + or : etc. on their base URL (they do on querystring)
For example, my IIS-10/windows 2016 or the CentOS mentioned by @Felli 

So is there a work around?

PR #2693, opened 2017-08-02
fix error

githubjeka:
```
PHP Warning 'yii\base\ErrorException' with message 'call_user_func() expects parameter 1 to be a valid callback, array must have exactly two members'

in \humhub\protected\vendor\yiisoft\yii2\base\Component.php:547
```

https://github.com/humhub/humhub/commit/83f69c8b0972bb2ed098a7ed1c4fccdff5f0faf2#commitcomment-23435580

PR #2694, opened 2017-08-02
Fix doc for Helpers::SqlMode

githubjeka:
Rel #2693

### DIFF
--- a/protected/humhub/libs/Helpers.php
+++ b/protected/humhub/libs/Helpers.php
@@ -234,12 +234,12 @@ class Helpers
 
     /**
      * Set sql_mode=TRADITIONAL for mysql server.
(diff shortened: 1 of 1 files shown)


luke-:
Please apologize the so delayed reply.
Is this PR and #2693 still a problem? I could not reproduce it, maybe it was fixed somehow else in the meantime.

githubjeka:
Hmm... On local, in one place it works, but in another place, this code doesn't work.
It's local problem probably

githubjeka:
Again... 
I catch it when clear cache and module uninstall & install...

githubjeka:
Yii2 issue https://github.com/yiisoft/yii2/issues/1988

PR #2695, opened 2017-08-02
Fix: URL Unnecessarily being too long

Felli:
Makes the URL unnecessarily too long and helps make #2691 work with CentOS slightly better.


buddh4:
In my tests the resulting url redirects to the dashboard and not to the user profile. Can you explain what you mean with url unnecessarily long, do you mean if you've activated prettyurl and the username is to long?

Felli:
On my server when you click on the "My profile" option on the account menu it redirects you to `/u/Felli/user/profile/home` where as removing this will redirect you straight to your profile `/u/Felli` either works but one is shorter in the long run.

buddh4:
perhaps this only works with  active urlrewrite

PR #2696, opened 2017-08-03
Applying Yii2 Coding Standards

acs-ferreira:
This is a huge PR, so it should be reviewed before merge.

All Yii2 Coding Standards applied and code cleaning of whole project.


githubjeka:
If you don't take it quickly, then it will be even more difficult. I wouldn't want the work to be lost in vain.

luke-:
Sorry we cannot check & merge this heavy commit.

PR #2708, opened 2017-08-08
Use secure Cookies

### LINKED ISSUE #2706: Session cookie should have secure flag if set over https
This cookie, which has been sent over an encrypted channel, doesn't have the secure flag set. As a result, an active network attacker can easily recover it.

acs-ferreira:
Should fix #2706

### DIFF
--- a/protected/humhub/config/web.php
+++ b/protected/humhub/config/web.php
@@ -15,7 +15,7 @@ $config = [
     'components' => [
         'request' => [
             'class' => 'humhub\components\Request',
-+           'csrfCookie' => ['httpOnly' => true, 'secure' => YII_ENV_PROD],
+            'csrfCookie' => ['httpOnly' => true, 'secure' => YII_ENV_PROD],
         ],
         'response' => [
             'class' => 'humhub\components\Response',
--- a/protected/humhub/config/web.php
+++ b/protected/humhub/config/web.php
@@ -15,7 +15,7 @@ $config = [
     'components' => [
         'request' => [
             'class' => 'humhub\components\Request',
-            'csrfCookie' => ['httpOnly' => true, 'secure' => true],
+            'csrfCookie' => ['httpOnly' => true, 'secure' => (isset(Yii::$app->request->isSecureConnection) ? false : true)],
         ],
         'response' => [
             'class' => 'humhub\components\Response',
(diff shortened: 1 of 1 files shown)


protected/humhub/config/web.php
```diff
@@ -15,23 +15,26 @@
     'components' => [
         'request' => [
             'class' => 'humhub\components\Request',
+            'csrfCookie' => ['httpOnly' => true, 'secure' => (isset(Yii::$app->request->isSecureConnection) ? false : true)],
```
githubjeka: Don't like because.
You use `Yii::$app->request` in config for `Yii::$app->request`...
Probable better add this check to EVENT BEFORE_REQUEST and configure it in there `$event->sender->getRequest()...`

protected/humhub/config/web.php
```diff
@@ -15,23 +15,26 @@
...
-            'loginUrl' => ['/user/auth/login']
+            'loginUrl' => ['/user/auth/login'],
         ],
         'errorHandler' => [
             'errorAction' => 'error/index',
         ],
         'session' => [
             'class' => 'humhub\modules\user\components\Session',
             'sessionTable' => 'user_http_session',
+            'cookieParams' => ['httpOnly' => true, 'secure' => (isset(Yii::$app->request->isSecureConnection) ? false : true)],
```
githubjeka: instead `(isset(Yii::$app->request->isSecureConnection) ? false : true)` you can use `Yii::$app->request->isSecureConnection`

protected/humhub/config/web.php
```diff
@@ -15,23 +15,26 @@
...
             'class' => 'humhub\components\Request',
+            'csrfCookie' => ['httpOnly' => true, 'secure' => (isset(Yii::$app->request->isSecureConnection) ? false : true)],
         ],
         'response' => [
             'class' => 'humhub\components\Response',
         ],
         'user' => [
             'class' => 'humhub\modules\user\components\User',
             'identityClass' => 'humhub\modules\user\models\User',
+            'identityCookie' => ['name' => '_identity', 'httpOnly' => true, 'secure' => (isset(Yii::$app->request->isSecureConnection) ? false : true)],
```
githubjeka: why isSecureConnection = false ?

acs-ferreira:
@githubjeka this check isn't even needed:

> `secure` - if `true`, cookie will be set **only** if HTTPS is used.

See https://github.com/samdark/yii2-cookbook/blob/master/book/cookies.md

githubjeka:
@samdark why isn't by default in core?

githubjeka:
@acs-ferreira if I used http,then I don't get the cookie ?

acs-ferreira:
Hum, maybe you're right @githubjeka, i've read it as the `flag` would be set only if served over https, not the `cookie` himself... 👎 

Even if https://scotthelme.co.uk/still-think-you-dont-need-https/

acs-ferreira:
Otherwise it can be set outside Yii on `.htaccess`, like:

```
<IfModule mod_headers.c>
    Header always edit Set-Cookie ^(.*)$ $1;HttpOnly;Secure env=HTTPS
</IfModule>
```

One-size-fits-all?

samdark:
`httpOnly` could be set but you need to make sure no HumHub JavaScript code is using cookies. `secure` doesn't make sense to have by default for install-able system such as HumHub.

PR #2735, opened 2017-08-23
fixed russian translate

hellomoto177:


Felli:
It's recommended to do any changes to languages over on http://translate.humhub.org rather than here, as once the humhub bot pulls from that site it will overwrite any here, and sometimes conflicts with coding happens that way..

Issue #2745, opened 2017-09-06
Yii2 1.2.3 will break the admin interface


devantoine:
See https://github.com/yiisoft/yii2/issues/14767

acs-ferreira:
You mean **2.0.13**.

For now, a non-merged PR to upgrade to 2.0.**12** _(fully-compatible, non-BC breaking)_ is [available](https://github.com/humhub/humhub/pull/2629).

devantoine:
Oh yes, I meant 2.0.13, I guess my brain was on idle when I created this issue.

devantoine:
@acs-ferreira But setting `yiisoft/yii2": ~2.0.12` in the composer file will tell composer to checkout Yii2 starting from 2.0.12. So when Yii2 2.0.13 will be released and someone will install Humhub from the sources it will break.

acs-ferreira:
I know, but Yii2 2.0.x should no BC...we're talking about minor versions after all.
But agree that we should stick to 2.0.12 fixed while this is not clear.

devantoine:
Nop, they say that at the moment that it's totally fine for them to break BC on minor versions.

acs-ferreira:
They are not aware of [SEMVER](http://semver.org/) means?

buddh4:
At least it does not seem to affect [SocialActivity::instance](https://github.com/humhub/humhub/blob/master/protected/humhub/components/SocialActivity.php#L93)

samdark:
@acs-ferreira it is minor. Yii uses extra number for patches (these are rare): https://github.com/yiisoft/yii2/releases/tag/2.0.11.2

acs-ferreira:
@samdark the problem is that we can't rely on SemVer to keep our apps up-to-date with minor changes (keeping BC).

Is there any doc that clearly specifies what versioning schema Yii is using?

Yii [recommends](http://www.yiiframework.com/doc-2.0/guide-rest-versioning.html) SemVer but doesn't rely on it?

samdark:
https://github.com/yiisoft/yii2/blob/master/docs/internals/versions.md

acs-ferreira:
I've found my answer [here](https://github.com/yiisoft/yii2/issues/7408#issuecomment-77088965), it's not SemVer, not Ferver, but ___YiiVer___ !

Regarding this, HumHub should stick to concrete version, so require (at this time/date):

`"yiisoft/yii2": 2.0.12`

samdark:
[...]

PR #2747, opened 2017-09-07
Stick to Yii2 concrete version

acs-ferreira:
See https://github.com/humhub/humhub/issues/2745


samdark:
The version could be changed with the next release of HumHub so it won't break unexpectedly in case of Yii team makes a mistake and pass a hidden breaking change in minor release. I support this change.

PR #2775, opened 2017-10-09
[RFC] Adding a null check to comment/notifications/NewComment::getCommentedRecord()

### LINKED ISSUE #2752: Call to a member function getCommentedRecord() on null

I did not reproduced, but I suppose the problem happens when there is a notification of a comment and this comment is removed.

### Additional info

1. in /app/humhub/protected/humhub/modules/comment/notifications/NewComment.php at line 178
```
171    /**
172     * The commented record e.g. a Post
173     *
174     * @return \humhub\modules\content\components\ContentActiveRecord
175     */
176    public function getCommentedRecord()
177    {
178        return $this->source->getCommentedRecord();
179    }
```
2. yii\base\ErrorHandler::handleFatalError()

| Q                | A
| ---------------- | ---
| HumHub version   | 1.2.0 (docker)

leuprechtroman:
This aims to fix #2752 .

By far this is not a perfect solution, but it does the job.
Any improvements are welcome!
With this fix, the notification looks like this:

humhub_bug_fix

IMHO there are a few ways we can tackle this:
 1) Find where the comment is deleted that is referenced here (could not be reproduced by now, see issue)
 2) Maybe add a [DELETED] Placeholder instead of null ?

\RFC


leuprechtroman:
Added the [Deleted] placeholder and fixed style issues.
The notification now looks like this
humhub_fix_2

luke-:
Thank you, in the rare case of orphaned notifications this is a good error handling.

Could you please accept our CLA at: https://www.humhub.org/cla/github-sign

leuprechtroman:
Accepted the CLA, merged upstream, now all tests are passing

PR #2793, opened 2017-10-20
preload="metadata" + add .webm

hagalaz:
To reduce traffic from the vidos -> preload = "metadata"
And render WEBM videos


acs-ferreira:
`<video>` tag is a non-void HTML element. It should be closed with `</video>` instead of `/>` @buddh4 @hagalaz

buddh4:
Thanks for the note, I've added the closing tags.

PR #2802, opened 2017-10-24
Clean and format core Activity module

acs-ferreira:


luke-:
Thank you!

It's really hard for us to review such large merges which over 40 files. Can you please try to commit such cleanups in smaller doses? :-)

PR #2838, opened 2017-11-24
Limit spaces in space chooser to make it scale

### LINKED ISSUE #2835: Space chooser performance issue with large amount of spaces

We have the following scenario:
The humhub installation contains a lot of spaces (e.g. 20000 in the test setup, should be similar in production). The admin user is usually member of all spaces. When rendering the space chooser, all 20000 spaces are rendered in the spacechooser, which takes a lot of request time and also slows down the browser significantly.

To solve this, the space chooser should limit the number of spaces to display and implement loading script on scrolling.

### Additional info

| Q                | A
| ---------------- | ---
| HumHub version   | 1.2 / master at a5559da7a958360298f9a42d31ef6020f670868d
| PHP version      | 7.1
| Operating system | Debian GNU/Linux

cebe:
- limit the number of items initially loaded for the space chooser to 100.
- implemenated loading on scroll to load items when needed (inspired by
  the implementation of the latest activity widget)
- tested this with a setup with 20000 spaces.
- fixes #2835

By setting the batch size to 100 there should be no difference in
behavior for normal use cases but it would still work in large setups.

### DIFF
--- a/protected/humhub/modules/space/widgets/Chooser.php
+++ b/protected/humhub/modules/space/widgets/Chooser.php
@@ -16,6 +16,11 @@ use yii\helpers\Html;
  */
 class Chooser extends Widget
 {
+    /**
+     * Maximum number of spaces to load per request.
+     */
+    const SPACE_BATCH_SIZE = 100;
+
 
     public static function getSpaceResult($space, $withChooserItem = true, $options = [])
     {
@@ -61,7 +66,7 @@ class Chooser extends Widget
     protected function getMemberships()
     {
         if (!Yii::$app->user->isGuest) {
-            return Membership::findByUser(Yii::$app->user->getIdentity())->all();
+            return Membership::findByUser(Yii::$app->user->getIdentity())->limit(static::SPACE_BATCH_SIZE)->all();
         }
     }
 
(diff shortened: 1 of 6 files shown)


Buliwyfa:
Thanks @cebe

protected/humhub/modules/space/widgets/Chooser.php
```diff
@@ -16,6 +16,11 @@
  */
 class Chooser extends Widget
 {
+    /**
+     * Maximum number of spaces to load per request.
+     */
+    const SPACE_BATCH_SIZE = 100;
```
cebe: This should probably be set even lower, e.g. 25, as there is still one query per space to determine new content count.

luke-:
@cebe We plan to merge this PR into the v1.3-dev branch as soon as we found a solution regarding the Enterprise Edition space chooser which also uses this base code.

The EE space chooser additionally needs to deal with space types and has no support for loading more items on scroll.

Felli:
@cebe does this also support mobile use?

BDSM-Master:
Can someone tell me why Administrators can not delete another members Space page?

cebe:
@Felli it does not change anything about the layout so I suppose this would also work on mobile as it worked before. But I have not tested it.

PR #2872, opened 2017-12-30
Remove dead code

danielkesselberg:
Found some dead code

### DIFF
--- a/protected/humhub/modules/installer/Module.php
+++ b/protected/humhub/modules/installer/Module.php
@@ -8,12 +8,12 @@
 
 namespace humhub\modules\installer;
 
-use Yii;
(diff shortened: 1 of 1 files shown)


githubjeka:
\humhub\modules\installer\Module::checkDBConnection no usages found in any places HumHub. Could be removed too

luke-:
@githubjeka It's used by the installer controllers e.g. https://github.com/humhub/humhub/blob/master/protected/humhub/modules/installer/controllers/ConfigController.php#L57

luke-:
@danielkesselberg With this PR PDOException is not longer handled. We should change also the import from:

```
use yii\base\Exception;
```

to 

```
use Exception;
```

PR #2873, opened 2017-12-30
Use temporary connection object

danielkesselberg:
I would prefer a temporary connection object for checking connectivity instead of the global db object


Felli:
I believe this should be an optional thing rather than a default option.

Issue #2848, opened 2017-12-06
Birthday date picker does not allow to set birthday more than 80 years ago


enguerrand:
### What steps will reproduce the problem?
In the profile edit view, try to set a birthday more than 80 years ago through the date picker and hit "save".

### What is the expected result?
That the birthday is set to the chosen date.

### What do you get instead?
An error message next to the birthday field that states that the birthday must be before today and the current value set to the current date.

### Additional info
The problem can be worked around by manually entering the full year into the input field (e.g. 1936)

| Q                | A
| ---------------- | ---
| HumHub version   | 1.2.2
| PHP version      | 7.0
| Operating system | Server: Debian Stretch: Client: Arch Linux with firefox 57

> Note: Please use the [community](https://community.humhub.com) site for any questions on customizing HumHub.
> Note: Also provide the full error stacks from `Administration --> Information --> Logging` and errors from your browser console if possible.

enguerrand:
Additional info: This issue can also be worked around by using the custom date display format in the admin settings. (Which leads to the date being entered as in the workaround mentioned before)

Felli:
A known workaround for this would be just using full year format `05/12/1918` or `05/12/2018` to differentiate between the years used from the past to the years in the future.

danielkesselberg:
@Felli @enguerrand is this issue fixed?

Felli:
For me, yes. :smile:

enguerrand:
Sorry, did not get the chance to look into this earlier. I just tested in version 1.2.4 using the default date format and the problem still occurs. For me personally the workaround to use a different date format ist fine though.

danielkesselberg:
@buddh4 @enguerrand patch for this issue has been merged to 1.3-dev branch (https://github.com/humhub/humhub/pull/2881). Issue should be close.

PR #2881, opened 2018-01-02
Use medium as date format

danielkesselberg:
In general all DateTime-Fields should follow the global and configurable dateInputFormat. For the Birthday-Field we should choose a default which protects the user from breaking functionality when changing global HumHub configuration.

Like described in #2848 the date format 'short' does not work for birthdates (18 = 2018 or 1918?).

### DIFF
--- a/protected/humhub/modules/user/models/fieldtype/Birthday.php
+++ b/protected/humhub/modules/user/models/fieldtype/Birthday.php
@@ -92,7 +92,7 @@ class Birthday extends Date
         $rules[] = [$this->profileField->internal_name . "_hide_year", 'in', 'range' => [0, 1]];
         $rules[] = [$this->profileField->internal_name,
             \humhub\libs\DbDateValidator::className(),
-            'format' => Yii::$app->formatter->dateInputFormat,
+            'format' => 'medium',
             'convertToFormat' => null,
             'max' => time(),
             'tooBig' => Yii::t('base', 'The date has to be in the past.')
@@ -107,7 +107,7 @@ class Birthday extends Date
     {
         return [$this->profileField->internal_name => [
                 'type' => 'datetime',
-                'format' => Yii::$app->formatter->dateInputFormat,
+                'format' => 'medium',
                 'class' => 'form-control',
                 'readonly' => (!$this->profileField->editable),
                 'yearRange' => (date('Y') - 100) . ":" . date('Y')
--- a/protected/humhub/modules/user/models/fieldtype/Birthday.php
+++ b/protected/humhub/modules/user/models/fieldtype/Birthday.php
@@ -123,7 +123,7 @@ class Birthday extends BaseType
         $rules[] = [
             $this->profileField->internal_name . '_hide_year',
             'in',
-            'range' => [0, 1]
+            'range' => [self::HIDE_AGE_NO, self::HIDE_AGE_YES]
         ];
 
         return parent::getFieldRules($rules);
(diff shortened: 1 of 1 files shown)


Felli:
Doesn't work for me sadly, makes input invalid.

danielkesselberg:
Hey, which language and date format? Could you add me a screenshot?Am 02.01.2018 23:04 schrieb Sarah Tsumayoi <notifications@github.com>:Doesn't work for me sadly, makes input invalid.

Felli:
I'm using the auto format based on user language.

danielkesselberg:
Ok. I tested this with english and german. Maybe there is another problem. Can you share me a screenshot?

Felli:
Sure once I'm done working on what I'm doing I'll go ahead and do this for you. :+1:

Felli:
Forces the following which then doesn't allow other methods to work as they should.

Felli:
Would seem that you missed `Yii::$app->formatter->dateInputFormat` within Date.php & DateTime.php as well.

danielkesselberg:
I changed it only for the Birthday field. Could you look into the generated source. At the bottom should be a line like this:
 `jQuery('#profile-birthday').datepicker({"changeYear":true,"yearRange":"1918:2018","changeMonth":true,"disabled":false,"dateFormat":"m\/d\/y"});`

Could you share your dateFormat with me?

danielkesselberg:
I guess its important the all date/datetime fields relay on dateInputFormat. I would only change the format for the birthdate field.

danielkesselberg:
Does it work for you when you change date & datetime too?

Felli:
Sadly @danielkesselberg the only thing that actually made it work for me was changing date/datetime as well which then makes everything fine.

Felli:
### Examples

danielkesselberg:
Which language do you configured as administrator?

Felli:
I always use the default English language as most of my users are English speakers.

Felli:
I've also tested this within my own language with it not working till I've changed both `Date.php` & `DateTime.php`. :smile:

danielkesselberg:
@Felli Thanks for your feedback. The current code does not fix the problem. Birthdate-Field extends from Date-Field an the validation rules are merged. See the attached screenshot.
image

danielkesselberg:
@Felli could you please test the latest update? Maybe there is more testing required for installations with older data.

protected/humhub/modules/user/models/fieldtype/Birthday.php
```diff
@@ -41,90 +52,121 @@ public function rules()
...
-        return [$this->profileField->internal_name => [
+        return [
+            $this->profileField->internal_name => [
                 'type' => 'datetime',
                 'format' => 'medium',
                 'class' => 'form-control',
                 'readonly' => (!$this->profileField->editable),
-                'yearRange' => (date('Y') - 100) . ":" . date('Y')
+                'yearRange' => (date('Y') - 100) . ":" . date('Y'),
+                'dateTimePickerOptions' => array(
```
Felli: It's not a big deal but using `[]` rather than `array()` would make for better code consistency.

Felli:
I'll test once I'm able t, at the moment it's under maintenance.

Felli:
@danielkesselberg now it works up to 1918, and the age now goes up to 100 in some cases (Depending on which month you are born in).

luke-:
Thank you @danielkesselberg  & @Felli - I'm merged this to v1.3-dev

PR #2915, opened 2018-01-12
Clean DirectoryController

acs-ferreira:


danielkesselberg:
Do you prefer single quotes over double quotes? We may add an additional rule to our ruleset.xml. @luke- @buddh4 what do you think?

acs-ferreira:
I do not "prefer" single, simply applying best practices IMHO :)
https://secure.php.net/manual/en/language.types.string.php#language.types.string.syntax.single

githubjeka:
codeclimate — 3 issues to fix

PR #2923, opened 2018-01-13
Exclude codeception generated

danielkesselberg:
Files inside codeception/_support/_generate could be ignored by codeclimate


githubjeka:
Should be in .gitignore

PR #2930, opened 2018-01-14
Fix ControllerAccess component

acs-ferreira:


githubjeka:
codeclimate — 3 issues to fix

PR #2948, opened 2018-01-18
Export users csv

danielkesselberg:


protected/humhub/modules/admin/controllers/PendingRegistrationsController.php
```diff
@@ -56,103 +40,73 @@ public function getAccessRules()
...
-            ['language'],
-            ['source'],
-            ['created_at', 'type' => 'datetime'],
+        return [
+            'id',
+            'user_originator_id',
+            'space_invite_id',
+            'email',
+            'source',
+            'token',
```
githubjeka: For what token?

protected/humhub/modules/admin/controllers/PendingRegistrationsController.php
```diff
@@ -56,103 +40,73 @@ public function getAccessRules()
...
+            'resultConfig' => [
+                'fileBaseName' => 'humhub_user',
+                'writerType' => $format,
+            ],
+        ]);
+
+        return $exporter->export()->send();
     }
 
-    public function createCVS($dataProvider, ActiveRecord $model, $format = null)
```
githubjeka: BC break

protected/humhub/modules/admin/controllers/PendingRegistrationsController.php
```diff
@@ -56,103 +40,73 @@ public function getAccessRules()
         ];
     }
 
-    public function actionIndex($export = false, $format = null)
+    public function actionIndex()
```
githubjeka: BC break

protected/humhub/modules/admin/controllers/PendingRegistrationsController.php
danielkesselberg: Thank you! We should add a note to changelog that createCVS is removed.

protected/humhub/modules/admin/controllers/PendingRegistrationsController.php
danielkesselberg: Thank you. Good point.

PR #2953, opened 2018-01-18
Fix UserModelModules behavior

acs-ferreira:


protected/humhub/modules/user/behaviors/UserModelModules.php
```diff
@@ -24,8 +25,8 @@
 class UserModelModules extends Behavior
 {
 
-    public $_enabledModules = null;
-    public $_availableModules = null;
+    public $enabledModules = null;
+    public $availableModules = null;
```
githubjeka: BC break

acs-ferreira:
@githubjeka why this?

githubjeka:
If some modules use `$userModelModules->_availableModules;` will be error

acs-ferreira:
I see, but ANY of HumHub modules use them at this time.
Just for the record.

And `_` should not be used to indicate visibility.

githubjeka:
> ANY of HumHub modules use them at this time.
This does not cancel BC

This code should be merged for 2.0.0 version HumHub, not for master (1.3dev now)

luke-:
(Closed since this behavior is not longer available in v1.3)

PR #2956, opened 2018-01-20
Enh: Oauth2

Felli:
Switch from Oauth1 to Oauth2 use.

> Note: This also supports Font Awesome 5 icons for those that use them.


protected/humhub/modules/user/authclient/Twitter.php
```diff
@@ -20,49 +21,32 @@ protected function defaultViewOptions()
         return [
             'popupWidth' => 860,
             'popupHeight' => 480,
-            'cssIcon' => 'fa fa-twitter',
+            'cssIcon' => 'fab fa-twitter',
```
githubjeka: fab ?

Felli: `fab` is used in Font Awesome 5 `fa` is deprecated.

githubjeka: Ok

githubjeka: But https://github.com/humhub/humhub/blob/958d940da60080d1a2cf2f02b429826c865b5a79/composer.json#L40

version for HumHub is `>=4.7.0 <5.0.0`

Felli: Sadly using `fa fa-twitter` or even `fa-twitter` wouldn't work anyways (Which are 4+), would also seem that they still have this issue within 5...

luke-: @Felli Maybe we can change to FA 5.0 in the v1.3?

luke-:
@Felli can you please recreate this PR directly for v1.3?

PR #2963, opened 2018-01-23
Better ruleset

acs-ferreira:


ruleset.xml
```diff
@@ -1,16 +1,49 @@
...
+    <description>HumHub coding standards</description>
 
     <!-- Include PSR-1 -->
     <rule ref="PSR1"/>
+    
+    <rule ref="PSR2">
+        <!-- Opening parenthesis of a multi-line function call must be the last content on the line. -->
+        <!-- Closing parenthesis of a multi-line function call must be on a line by itself. -->
+        <!-- Only one argument is allowed per line in a multi-line function call. -->
+        <exclude name="PEAR.Functions.FunctionCallSignature"/>
```
danielkesselberg: I like this rule. Why do you want to remove this rule?

acs-ferreira: Agree :)

PR #2965, opened 2018-01-23
Update phpdoc (add missing throws)

danielkesselberg:
- Added throws to phpdoc
- Remove is_dir (https://github.com/yiisoft/yii2/blob/master/framework/helpers/BaseFileHelper.php#L545 the first check inside createDirectory is is_dir)

@acs-ferreira fyi


acs-ferreira:
Nice catch about `is_dir` 👍

Felli:
Stupid question, but wouldn't you also want to check to make sure that the file name isn't already taken?

acs-ferreira:
@Felli filename doesn't matter as for each export, a temporary directory is created with a unique ID based on `time()`

PR #2969, opened 2018-01-26
Update phpdoc, Remove Yii methods, Codestyle

danielkesselberg:
@githubjeka @acs-ferreira


danielkesselberg:
@luke- @buddh4 I removed the event code from Yii1

PR #2987, opened 2018-02-05
Format and clean

acs-ferreira:


luke-:
(This is not longer available in v1.3)

PR #2992, opened 2018-02-05
Update Bundle of CA Root Certificates

acs-ferreira:


danielkesselberg:
Just wondering. Where is this from?

acs-ferreira:
https://curl.haxx.se/docs/caextract.html
https://github.com/humhub/humhub/blob/325eea05dbfac30abcb8166b437dfb41bb07204a/protected/humhub/libs/CURLHelper.php#L33

danielkesselberg:
Good to know. Thank you.

PR #2957, opened 2018-01-20
PHP 7.2 and Yii 2.0.13 compatibility

acs-ferreira:
https://github.com/yiisoft/yii2/blob/2.0.13.1/framework/UPGRADE.md#upgrade-from-yii-2012

Should be merged **ONLY** when HumHub is ready to switch to Yii 2.0.13
https://github.com/humhub/humhub/blob/d755cc1b1258ec66d7260ba2b8c3f6dda2bd6bc1/composer.json#L18

as `yii\base\BaseObject` has been introduced in Yii 2.0.13

Should fix #2885

PR #3016, opened 2018-02-14
PHP 7.2 and Yii 2.0.13 compatibility

githubjeka:
REF #2957

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
     "minimum-stability": "stable",
     "require": {
         "php": ">=5.6.0",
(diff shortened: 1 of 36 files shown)


acs-ferreira:
We should switch directly to 2.0.14 as this is the latest 2.0.x that shoul break BC. See http://www.yiiframework.com/news/165/yii-2-0-14-is-released/

githubjeka:
2.0.14.1

githubjeka:
Better wait some weeks after release....

PR #3023, opened 2018-02-15
Fix for Delete stream entry in permalink view

### LINKED ISSUE #2912: Delete stream entry in permalink view

1. Enter a permalink of a post
2. Delete this post by means of the stream entry dropdown

-  The stream should be reloaded with all remaining entries with an success message.

### What do you get instead?

- An empty stream message

| Q                | A
| ---------------- | ---
| HumHub version   | 1.2

githubjeka:
fix #2912

### DIFF
--- a/protected/humhub/modules/stream/resources/js/humhub.stream.js
+++ b/protected/humhub/modules/stream/resources/js/humhub.stream.js
@@ -876,8 +876,14 @@ humhub.module('stream', function (module, require, $) {
 
     WallStream.prototype.onChange = function () {
         var hasEntries = this.hasEntries();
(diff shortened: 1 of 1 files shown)


Felli:
Doesn't fully reload for me :thinking:

screenshot_1

githubjeka:
Try delete all assets and check DEV env

Felli:
Works on dev, not production but doesn't truly reload should think about using `.empty()` or `.remove()` for content deletion. :thinking:

buddh4:
https://github.com/humhub/humhub/commit/f385e0cd5d829743ddc297c784434786ee25e5b5

PR #3024, opened 2018-02-16
fix Invalid 200 OK when comment is empty

githubjeka:
When POST request with empty comment sends then get 200OK. It increases counters comment in wallEnrty, but comment not flashed

PR fixed it

### DIFF
--- a/protected/humhub/modules/comment/controllers/CommentController.php
+++ b/protected/humhub/modules/comment/controllers/CommentController.php
@@ -16,6 +16,8 @@ use humhub\components\behaviors\AccessControl;
 use humhub\modules\comment\models\Comment;
 use humhub\modules\comment\widgets\Comment as CommentWidget;
 use humhub\modules\comment\widgets\ShowMore;
+use yii\web\BadRequestHttpException;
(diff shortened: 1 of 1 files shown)


protected/humhub/modules/comment/controllers/CommentController.php
```diff
@@ -78,22 +86,20 @@ public function actionShow()
...
+            );
         }
 
         $message = Yii::$app->request->post('message');
         $files = Yii::$app->request->post('fileList');
 
         if (empty(trim($message)) && empty($files)) {
-            // do not create empty comments
-            return '';
+            throw new BadRequestHttpException('The comment is wrong');
```
Felli: Instead of using `The comment is wrong` you should think of using `Comment could not be created`

githubjeka: ... and users will be created topics "why I can't create comment ?" ;)

P.S. this action used by javascript HubHub code. Before PR javascript get 200OK if comment empty. Now 400 error with flash message.

 TODO: Javascript code should not be passed empty comments.

PR #3031, opened 2018-02-18
Rename MIME_* into ICON_* to make purpose more clear

danielkesselberg:


acs-ferreira:
Should be extended to CSS IMO.
https://github.com/humhub/humhub/blob/132af2fcda577c55b95874a05b5363bd218f63c2/static/less/file.less#L115

PR #3037, opened 2018-02-19
Update userexport

danielkesselberg:


protected/humhub/modules/admin/views/user/list.php
```diff
@@ -59,11 +66,10 @@
...
                 ['class' => UserActionColumn::class],
             ],
-        ]);
-        ?>
+        ]); ?>
     </div>
-    <?php if ($showPendingRegistrations): ?>    
-        <br />
+    <?php if ($showPendingRegistrations): ?>
+        <br/>
```
acs-ferreira: Simple `<br>` is sufficient :)

PR #3040, opened 2018-02-20
Clean

acs-ferreira:


acs-ferreira:
You're right, didn't realize GIF can be animations too...

PR #3052, opened 2018-02-26
WIP [Active record]

githubjeka:
Problems `\humhub\components\ActiveRecord`:

- [ ] ActiveRecord's required `Yii::$app->user` component
- [ ] All ActiveRecord's models use FileManager
- [ ] Some core models used functional of ActiveRecord(`created_at`, `updated_by` ... )but haven't tables in database.
- [ ] One's of core models used only `updated_` fields, others - only `created_`
- [ ]   7 out of 13 cases `beforeSave()` is not involved for one reason or another. See test 
image

- [x] `getCreatedBy` and `getUpdatedBy` do not check the existence of a property `created_by` and `updated_by`

### DIFF
--- a/protected/humhub/modules/user/models/Invite.php
+++ b/protected/humhub/modules/user/models/Invite.php
@@ -8,7 +8,6 @@
 
 namespace humhub\modules\user\models;
 
-use humhub\components\ActiveRecord;
 use Yii;
 use yii\helpers\Url;
 
@@ -21,16 +20,13 @@ use yii\helpers\Url;
  * @property string $email
  * @property string $source
  * @property string $token
- * @property string $created_at
- * @property integer $created_by
- * @property string $updated_at
- * @property integer $updated_by
  * @property string $language
  * @property string $firstname
  * @property string $lastname
  */
-class Invite extends ActiveRecord
+class Invite extends \humhub\db\HActiveRecord
 {
+    use \humhub\db\traits\MetaActiveRecordTrait;
 
     const SOURCE_SELF = "self";
     const SOURCE_INVITE = "invite";
@@ -90,7 +86,7 @@ class Invite extends ActiveRecord
         if ($insert && $this->token == '') {
             $this->token = Yii::$app->security->generateRandomString(self::TOKEN_LENGTH);
         }
-
+        $this->setMeta($insert, $this->user_originator_id);
         return parent::beforeSave($insert);
     }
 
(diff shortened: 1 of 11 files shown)


githubjeka:
I propose to add `@deprecated` to current AR and move functional AR to specific places, for example, traits

luke-:
@githubjeka Good point! Should be no problem to mark the HumHub AR as deprecated and implement the functions as Traits. From our side you could start - thank you!

luke-:
There are also some Events using the HumHub AR.
https://github.com/humhub/humhub/blob/master/protected/humhub/modules/notification/config.php#L21

We also need to change this to the default Yii2 AR.

danielkesselberg:
We may consider http://www.yiiframework.com/doc-2.0/yii-behaviors-timestampbehavior.html and http://www.yiiframework.com/doc-2.0/yii-behaviors-blameablebehavior.html.

githubjeka:
So... Need to think more about behaviors and theirs expand

githubjeka:
Added traits. Need your review and thoughts

protected/humhub/modules/user/models/Invite.php
```diff
@@ -21,16 +20,13 @@
...
- * @property string $created_at
- * @property integer $created_by
- * @property string $updated_at
- * @property integer $updated_by
  * @property string $language
  * @property string $firstname
  * @property string $lastname
  */
-class Invite extends ActiveRecord
+class Invite extends \humhub\db\HActiveRecord
```
githubjeka: Removed `FileManager` functional for Invite model.
Meta information about created_..., updated_... available by `\humhub\db\traits\MetaActiveRecordTrait`

protected/humhub/modules/user/models/Invite.php
```diff
@@ -90,7 +86,7 @@ public function beforeSave($insert)
         if ($insert && $this->token == '') {
             $this->token = Yii::$app->security->generateRandomString(self::TOKEN_LENGTH);
         }
-
+        $this->setMeta($insert, $this->user_originator_id);
```
githubjeka: Need to throw false in beforeSave() If created_by or updated_by will set to NULL(user_id is wrong) then?

githubjeka: For example:

```php
public function beforeSave($insert)
    {
        if (!parent::beforeSave($insert)) {
            return false;
        }
        if ($insert && $this->token == '') {
            $this->token = Yii::$app->security->generateRandomString(self::TOKEN_LENGTH);
        }
        return $this->setMeta($insert, $this->user_originator_id);
    }
```

buddh4:
As a small note I'am not a fan of those prefixed class names like`HActiveRecord`. I Think `humhub\db\ActiveRecord` should be enought to distinguish the classes, we do the same for other components as controllers etc.

Issue #2679, opened 2017-07-27
Translations (translate.humhub.org)


githubjeka:
When you update translation from translate.humhub.org then you reset translation which that has been added in GitHub

See for example https://github.com/humhub/humhub/commits/master/protected/humhub/modules/space/messages/ru/widgets_SpacePickerField.php

Please check your method to update from translate.humhub.org

and apply my commit please again https://github.com/humhub/humhub/commit/68539807b3f1fad2db2e5c1ab4cbf2c8a4b7e840

UPDATE: https://github.com/humhub/humhub/pull/2680 PR to revert

buddh4:
Don't know how the synchronization is done in translat.humhub.org, but translations over translate.humhub.org are prefered over direct git commits, since there can be conflicts with bot commits.

PR #3060, opened 2018-02-28
Update widgets_UploadInput.php

Felli:
Fix for `es` language breaking.


protected/humhub/modules/file/messages/es/widgets_UploadInput.php
```diff
@@ -1,22 +1,6 @@
 <?php
-/**
```
githubjeka: Why is deleted?

Felli: This is an auto-generated message which isn't really needed and can be removed.

githubjeka: auto-generated message if the file is new and was created. After removed the message you never see it again. 
That why to need restore it.

Felli: Shouldn't matter once a pull from the translation site is made, originally the message wasn't included in the translation file. :thinking:

githubjeka: According to my observations, the bot sends diff changes translation from lasted saved commit. Like `git pull`... Sometimes he pushes incorrect commit and rewrite GitHub translation, if it not was fixed. See https://github.com/humhub/humhub/issues/2679

githubjeka: If it fixed, Bot will send only new translation row, not this message. If it not fixed,  Bot misses your changes and rewrite like my commit.
In any case, I recommend returning this message.

Felli: Good point, once I'm on my work laptop I'll fix this. :smile:

Felli: Should be fixed now. :+1:

Issue #3058, opened 2018-02-27
Copyright year


githubjeka:
I saw you merge PR with update Copyright year value in head files.

https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated

danielkesselberg:
Its a good point obviously. After reading stackoverflow and reddit i would recommend to do it like yii (they always write 2008 as year) or drop the year (like shopware does).

luke-:
Maybe we should generally update the copyright header?

Any suggestions?

PR #3067, opened 2018-03-03
Added Database name on admin > information > database panel

githubjeka:


protected/humhub/modules/admin/views/information/database.php
```diff
@@ -1,8 +1,27 @@
+<?php
+/**
+ * @link https://www.humhub.org/
+ * @copyright Copyright (c) 2017 HumHub GmbH & Co. KG
```
Felli: Should change this to `2018` to keep with the times. :wink:

githubjeka: Ref https://github.com/humhub/humhub/issues/3058

githubjeka: This file was created much earlier :/

Felli: Hmm... Then wouldn't it be better to use `2017-2018`? I remember seeing something along the lines of being used, either way, is fine, just my thoughts on the whole copyright signing. :thinking:

protected/humhub/modules/admin/controllers/InformationController.php
```diff
@@ -42,7 +43,7 @@ public function init()
     public function getAccessRules()
     {
         return [
-            ['permissions' => \humhub\modules\admin\permissions\SeeAdminInformation::className()]
+            ['permissions' => \humhub\modules\admin\permissions\SeeAdminInformation::className()],
```
danielkesselberg: Please use ::class instead of className().

protected/humhub/modules/admin/tests/codeception/unit/DatabaseInfoTest.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace humhub\modules\admin\tests\codeception\unit;
+
+use humhub\modules\admin\controllers\services\DatabaseInfo;
+use tests\codeception\_support\HumHubDbTestCase;
+
+class DatabaseInfoTest extends HumHubDbTestCase
```
danielkesselberg: I guess it is not required to extend from HumHubDbTestCase. `Codeception\Test\Unit`

protected/humhub/modules/admin/controllers/services/DatabaseInfo.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+/**
+ * @link https://www.humhub.org/
+ * @copyright Copyright (c) 2018 HumHub GmbH & Co. KG
+ * @license https://www.humhub.com/licences
+ */
+
+namespace humhub\modules\admin\controllers\services;
```
danielkesselberg: I would prefer something like humhub\modules\admin\components

PR #3068, opened 2018-03-03
IDE autocompletion

githubjeka:
- Added `Yii.php` to IDE code autocompletion for HumHub components
- Fixed phpDoc for Modules entities

PR #3069, opened 2018-03-03
WIP  Added Activity on Space level when Module enabled/disabled

githubjeka:
ref #3068 

image

See http://docs.humhub.org/dev-activities.html
*Problem:* How to return Module name? There is no functional in the Activity models to render messages with some parameters other than `$originator`

### DIFF
--- a/protected/humhub/modules/content/activities/ModuleDisabledActivity.php
+++ b/protected/humhub/modules/content/activities/ModuleDisabledActivity.php
@@ -1,0 +1,52 @@
+<?php
+/**
+ * @link https://www.humhub.org/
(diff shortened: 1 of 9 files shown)


githubjeka:
To fix the problem I could add new primary key `id` to `contentcontainer_module` table like https://github.com/humhub/humhub/pull/3076/files#diff-a4642a29063b113f7c8e8363bbd11915

luke-:
Maybe it makes sense to enhance the Activity/Notification models a little bit to add custom data.

I had a similar problem few days ago with a Notification about deleted content. In that case, there wouldn't even be an ID.

PR #3071, opened 2018-03-03
Added param "the since date" for members on their about pages

githubjeka:
image

### DIFF
--- a/protected/humhub/modules/user/views/profile/about.php
+++ b/protected/humhub/modules/user/views/profile/about.php
@@ -20,7 +20,7 @@ $categories = $user->profile->getProfileFieldCategories();
     <div class="panel-body">
 
         <p>
-            The member of <?= \Yii::$app->name ?> since date <?= Yii::$app->formatter->asDate($user->created_at) ?>
+            A proud member of <?= \Yii::$app->name ?> since date <?= Yii::$app->formatter->asDate($user->created_at) ?>
         </p>
 
         <?php $firstClass = "active"; ?>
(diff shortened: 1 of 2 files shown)


protected/humhub/modules/user/views/profile/about.php
```diff
@@ -12,20 +12,32 @@
...
-        class="panel-heading"><?= Yii::t('UserModule.views_profile_about', '<strong>About</strong> this user'); ?></div>
+
+    <div class="panel-heading">
+        <?= \Yii::t('UserModule.views_profile_about', '<strong>About</strong> this user'); ?>
+    </div>
+
     <div class="panel-body">
+
+        <p>
+            The member of <?= \Yii::$app->name ?> since date <?= Yii::$app->formatter->asDate($user->created_at) ?>
```
Felli: Instead of using `The` use `A` better grammar, or even use `A proud member of <?= \Yii::$app->name ?> since date <?= Yii::$app->formatter->asDate($user->created_at) ?>`. :smile:

danielkesselberg:
I would prefer this as another row within the general tab. Something like Member since   XXX

githubjeka:
The general tab generates via user-profile property by admin panel settings. Roughly in that, it is filled manually. Member since is the direct query from `created_at` of the database. It's too expensive to change something

luke-:
Maybe we should queue up this feature until we refactored the profile details view.
It looks a bit out of place at the moment.

I hope we can start with the refactoring here after the 1.3 release

luke-:
@githubjeka I'll close this PR in the meantime. Anyway thank you for the help!

PR #3074, opened 2018-03-05
Enh/Fix: Cache Handling + File Preview Fix

Felli:
For better caching of commonly uploaded file types also added `Options +FollowSymLinks` to help fix some issues with file previews in some hosting environments.


Felli:
Also thought of adding `Block bad bots` but there were a lot of entries and just didn't add it in. :thinking:

Felli:
@danielkesselberg can you review this?

Felli:
Have to find a working way of using `Content-Security-Policy` & `X-Frame-Options` (Most likely going to just drop the idea of `X-Frame-Options` altogether because it blocks all iframe even if ALLOWALL is given...)

acs-ferreira:
IMHO, these options should not be sent by the framework/application itself, it could lead to a lot of problems.

These should be set **by the user** itself, if one has the knowledge to do so.

danielkesselberg:
@Felli please split this pr. The Content-Security & X-Frame-Options is something else than the cache stuff.

Felli:
@danielkesselberg I just didn't want to create too many p/r felt slightly lazy, but once I'm on my laptop I'll do that.
@acs-ferreira might be for the best to comment them out then. :thinking:

Felli:
@danielkesselberg done.

Felli:
I believe this applies for some dedicated Linux servers.

PR #3079, opened 2018-03-06
Added notification for members of groups

githubjeka:
image


protected/humhub/modules/admin/notifications/IncludeGroupNotification.php
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function getMailSubject()
+    {
+        return \Yii::t(
+            'AdminModule.notification',
+            'Notify from {appName}. You were removed from the group.',
```
luke-: You were added to the group.

protected/humhub/modules/admin/controllers/GroupController.php
```diff
@@ -181,8 +191,10 @@ public function actionEditManagerRole()
 
         if ($group == null) {
```
danielkesselberg: Should be safe to use === here

protected/humhub/modules/admin/controllers/GroupController.php
```diff
@@ -181,8 +191,10 @@ public function actionEditManagerRole()
...
             throw new HttpException(404, Yii::t('AdminModule.controllers_GroupController', 'Group not found!'));
-        } else if ($value == null) {
-            throw new HttpException(400, Yii::t('AdminModule.controllers_GroupController', 'No value found!'));
+        } else {
```
danielkesselberg: What do you think about `$group !== null && $value === null` as else if? I think its more clear that $value should be not null when $group is not null.

PR #3081, opened 2018-03-06
improve performance for search (avoid memory issues)

cebe:
when limiting a search query to spaces, only the space IDs are necessary.
When a user is member in many spaces, loading all the spaces objects
consumes a lot of memory and in extreme cases can cause Humhub to crash
with out of memory error.

This problem is fixed by adding a new method
`Membership::GetUserSpaceIds()` a new private method is added to ensure
the query is the same for `GetUserSpaceIds()` and `GetUserSpaces()`.

### DIFF
--- a/protected/humhub/modules/space/models/Membership.php
+++ b/protected/humhub/modules/space/models/Membership.php
@@ -170,22 +170,51 @@ class Membership extends \yii\db\ActiveRecord
         $spaces = Yii::$app->cache->get($cacheId);
         if ($spaces === false || !$cached) {
 
-            $orderSetting = Yii::$app->getModule('space')->settings->get('spaceOrder');
-            $orderBy = 'name ASC';
-            if ($orderSetting != 0) {
-                $orderBy = 'last_visit DESC';
-            }
-
-            $query = self::find()->joinWith('space')->orderBy($orderBy);
-            $query->where(['user_id' => $userId, 'space_membership.status' => self::STATUS_MEMBER]);
-
             $spaces = [];
-            foreach ($query->all() as $membership) {
+            foreach (static::getMembershipQuery($userId)->all() as $membership) {
                 $spaces[] = $membership->space;
             }
             Yii::$app->cache->set($cacheId, $spaces);
         }
         return $spaces;
+    }
+
+    /**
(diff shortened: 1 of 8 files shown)


cebe:
> codeclimate — 4 issues to fix 

most of what codeclimate complains about is in the original code, should I fix it?

githubjeka:
this is fixed in my PR's for 1.3-dev

githubjeka:
Try to push in 1.3-dev. This next planned version.

protected/humhub/modules/space/models/Membership.php
```diff
@@ -170,24 +170,53 @@ public static function GetUserSpaces($userId = "", $cached = true)
...
         return $spaces;
     }
 
+    /**
+     * Returns a list of all spaces' ids of the given userId
+     *
+     * @param integer $userId
+     * @since 1.2.5
+     */
+    public static function GetUserSpaceIds($userId = "")
```
githubjeka: By http://docs.humhub.org/dev-coding-standards.html you could rename to `getUserSpaceIds($userId = '')`

cebe: done.

protected/humhub/modules/space/models/Membership.php
```diff
@@ -159,35 +165,61 @@ public function countNewItems()
      * @param boolean $cached use cached result if available
      * @return Space[] an array of spaces
      */
-    public static function GetUserSpaces($userId = "", $cached = true)
+    public static function getUserSpaces($userId = '', $cached = true)
```
githubjeka: For I know, on Windows `ClassName === className` is true, on Linux it will syntax error, about the name of methods I don't know. Are you sure about getUserSpaces for this case isn't BC ?

cebe: PHP does not care about method case, the classname issue on windows/linux is about file names and autoloading. No autoloading is involved in this case, so its fine.

>     Note: Function names are case-insensitive, though it is usually good form to call functions as they appear in their declaration. 
> https://secure.php.net/manual/en/functions.user-defined.php

cebe:
> codeclimate — 6 fixed issues

PR #3082, opened 2018-03-07
Better handle of CURLOPT_FOLLOWLOCATION

acs-ferreira:


protected/humhub/models/UrlOembed.php
```diff
@@ -179,9 +180,11 @@ public static function fetchUrl($url)
         curl_setopt($curl, CURLOPT_RETURNTRANSFER, 1);
         curl_setopt($curl, CURLOPT_TIMEOUT, 15);
 
-        // Not available when open_basedir is set.
-        @curl_setopt($curl, CURLOPT_FOLLOWLOCATION, true);
-
+        // Not available when open_basedir or safe_mode is set.
+        if (!function_exists('ini_get') || !ini_get('open_basedir')) || !ini_get('safe_mode')) {
```
danielkesselberg: Just wondering. Why do you call function_exists('ini_get')?

acs-ferreira: To avoid warnings like `PHP Warning: ini_get_all() has been disabled for security reasons' :)

danielkesselberg: I was afraid of something like that ;-)

PR #3107, opened 2018-03-20
Apply Yii security patch

acs-ferreira:
See http://www.yiiframework.com/news/168/releasing-yii-2-0-15-and-database-extensions-with-security-fixes/

> Upgrading Yii addresses the SQL injection but doesn't make findOne() and findAll() safe in general. Check all usages of findOne() and findAll() in your application. Also note, that where() and filterWhere() never escape column names, so if you need to pass a variable as a column name, make sure it is safe.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
     "minimum-stability": "stable",
     "require": {
         "php": ">=5.6.0",
-        "yiisoft/yii2": "2.0.14.*",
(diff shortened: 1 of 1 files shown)


luke-:
Thanks: https://github.com/humhub/humhub/commit/ceb6f4c7fb97222a8189e03f8e64dd8fc9fae278

PR #3112, opened 2018-03-24
Fix: Install in Sub-directory

Felli:
Issue can be seen [here](https://community.humhub.com/content/perma?id=144928).

### DIFF
--- a/.htaccess.dist
+++ b/.htaccess.dist
@@ -2,6 +2,9 @@
 <IfModule mod_rewrite.c>
 RewriteEngine on
 
+# uncomment if you're using a subdomain example: http://yourdomain.com/humhub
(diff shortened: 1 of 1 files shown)


danielkesselberg:
I think that subdomain is confusing. In fact we are not talking about a subdomain here. "When you installed humhub into a folder relative to your webroot like mydomain.com/humhub please adjust RewriteBase to /humhub or whatever you named the folder)".

Felli:
Should shorten as much as possible and be clear as well when dealing with it, I can agree with that. So maybe something like

`Uncomment if you've installed HumHub into a subdirectory relative to your webroot & adjust RewriteBase /humhub to match the install point`?

PR #3154, opened 2018-05-29
Show 'never' instead of fake date

acs-ferreira:
image

### DIFF
--- a/protected/humhub/modules/space/modules/manage/views/member/pending-invitations.php
+++ b/protected/humhub/modules/space/modules/manage/views/member/pending-invitations.php
@@ -1,6 +1,5 @@
 <?php
 
-
 use yii\helpers\Html;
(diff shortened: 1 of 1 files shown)


protected/humhub/modules/space/modules/manage/views/member/pending-invitations.php
```diff
@@ -27,29 +26,31 @@
                     [
                         'attribute' => 'last_visit',
                         'format' => 'raw',
-                        'value' =>
-                        function($data) use(&$groups) {
-                            return humhub\widgets\TimeAgo::widget(['timestamp' => $data->last_visit]);
+                        'value' => function ($data) use (&$groups) {
+                            if ($data->last_visit == '') {
```
danielkesselberg: I would prefer empty($data->last_visit) because it works for '' or 0 or null

PR #3177, opened 2018-06-11
Enh: Make Latest activities collapsible

Felli:
# Example
[https://gyazo.com/05a0e6ec03ad8f6b697cfbb04fd1a124](https://gyazo.com/05a0e6ec03ad8f6b697cfbb04fd1a124)

### DIFF
--- a/protected/humhub/modules/activity/widgets/views/activityStream.php
+++ b/protected/humhub/modules/activity/widgets/views/activityStream.php
@@ -3,28 +3,24 @@
 /**
  * @link https://www.humhub.org/
  * @copyright Copyright (c) 2017 HumHub GmbH & Co. KG
- * @license https://www.humhub.com/licences
+ * @license https://www.humhub.org/licences
  */
- 
+
 /* @var $this humhub\components\View */
 /* @var $streamUrl string */
 
 \humhub\modules\activity\assets\ActivityAsset::register($this);
 
-$this->registerJsConfig([
-    'activity' => [
-        'text' => [
-            'activityEmpty' => Yii::t('ActivityModule.widgets_views_activityStream', 'There are no activities yet.')
-        ]
-    ]
-]);
-
+$this->registerJsConfig(['activity' => ['text' => ['activityEmpty' => Yii::t('ActivityModule.widgets_views_activityStream', 'There are no activities yet.') ]]]);
 ?>
 
-<div class="panel panel-default panel-activities">
-    <div class="panel-heading"><?= Yii::t('ActivityModule.widgets_views_activityStream', '<strong>Latest</strong> activities'); ?></div>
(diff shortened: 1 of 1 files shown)


buddh4:
Sorry did miss this one, if you like repen it.

Felli:
Would seem that in 1.3beta it's not working right...

Felli:
Should be ready now! :+1:

protected/humhub/modules/activity/widgets/views/activityStream.php
```diff
@@ -16,11 +16,13 @@
...
 
-
-<div class="panel panel-default panel-activities">
+<div class="panel panel-default panel-activities" id="panel-activities">
+   <?= \humhub\widgets\PanelMenu::widget(['id' => 'panel-activities']); ?>
     <div class="panel-heading"><?= Yii::t('ActivityModule.widgets_views_activityStream', '<strong>Latest</strong> activities'); ?></div>
+   <div class="panel-body">
     <?= Html::beginTag('div', $options) ?>
         <ul id="activityContents" class="media-list activities" data-stream-content>
         </ul>
```
danielkesselberg: Could you place </ul> on line 24?

protected/humhub/modules/activity/widgets/views/activityStream.php
```diff
@@ -16,11 +16,13 @@
...
+      </div>
```
danielkesselberg: I think this is a bit confusing. We open the div around `<ul id="activityContents">` with `<?= Html::beginTag('div', $options) ?>` and close it with a regular `</div>` while we close the div for `<div class="panel-body">` with `<?= Html::endTag('div')?>`. I would prefer to close a tag opened by `<?= Html::beginTag() ?>` with `<?= Html::endTag() ?>` and vice versa.

Felli: Sadly this is the only way that I've gotten it to actually work on collapsing and advise offered would be much appreciated, thanks. :smiley:

protected/humhub/modules/activity/widgets/views/activityStream.php
Felli: I can do this, the reason I didn't was that Atom auto line breaks to clean my code up like I like it. lol

protected/humhub/modules/activity/widgets/views/activityStream.php
danielkesselberg: Sounds strange. What happens when u switch line 26/27?

danielkesselberg:
@Felli could you please review https://github.com/humhub/humhub/pull/3248? I used another approach to implement this feature. Some of the widgets are hard to change caused by inheritance structure.

PR #3179, opened 2018-06-13
Fix for humhub\components\Theme

Tahrz:
Fix for Theme class force `pathMap` change. So now, you can change module default views in your config files with `pathMap`.

### DIFF
--- a/protected/humhub/components/Theme.php
+++ b/protected/humhub/components/Theme.php
@@ -52,9 +52,9 @@ class Theme extends \yii\base\Theme
             $this->setBasePath('@webroot/themes/' . $this->name);
         }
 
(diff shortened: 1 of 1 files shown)


protected/humhub/components/Theme.php
```diff
@@ -52,9 +52,9 @@ public function init()
             $this->setBasePath('@webroot/themes/' . $this->name);
         }
 
-        $this->pathMap = [
+        $this->pathMap = array_merge($this->pathMap, [
```
danielkesselberg: Build fails because $this->pathMap could be (and is by default) null. Please wrap a if (is_array($this->pathMap)) around array_merge.

PR #3190, opened 2018-06-27
fix #3156 Modal of list members is not closed after click on a member

### LINKED ISSUE #3156: Modal of list members is not closed after click on a member
Go to a Space, show List Members

image

Click on a member

Redirected to member profile, modal closed

### What do you get instead?

Redirected to member profile, modal NOT closed

### Additional info

| Q                | A
| ---------------- | ---
| HumHub version   | master

reinq:
I think it's easier to control pjax from view widgets

### DIFF
--- a/protected/humhub/modules/space/widgets/views/listBox.php
+++ b/protected/humhub/modules/space/widgets/views/listBox.php
@@ -66,6 +66,6 @@ use yii\helpers\Html;
 
     // scroll to top of list
     $(".modal-body").animate({scrollTop: 0}, 200);
-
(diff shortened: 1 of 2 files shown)


buddh4:
If we decide to close modals after each pjax call we should add it to the `humhub.modules.ui.modal` js module init logic. But I don't know if this behaviour is always wished. The problem with your logic is that the `pjax:end` handler is only registered in case you open this modal, and it stays active afterwards, this can have strange side effects. You could register the handler by means of ` $(document).one(...)` but even this is inconsistent in case the modal is closed.

reinq:
Yes you are right. I'll try to suggest another option.

PR #3193, opened 2018-07-02
#3156 Modal of list members is not closed after click on a member

reinq:


buddh4:
Thanks for the effort, I've solved this with https://github.com/humhub/humhub/commit/7643b04878a5de72e6bf8a2aff2075ec5ba37f8c

PR #3205, opened 2018-07-08
Added a "Interlink wiki page" toolbar-button

olekrisek:
A simple toolbar-button to convert marked text to link for another wiki
page. "marked text" => [marked text](marked text). + switched icon on
the old link button to "fa-external-link".

### DIFF
--- a/static/js/humhub/humhub.ui.markdown.js
+++ b/static/js/humhub/humhub.ui.markdown.js
@@ -123,9 +123,23 @@ humhub.module('ui.markdown', function (module, require, $) {
                 [{
                     name: "groupCustom",
                     data: [{
+                        name: "cmdInterLinkWiki",
+                        title: "InterlinkWiki",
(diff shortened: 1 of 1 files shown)


luke-:
Hmm, this would appear also on non wiki editors?

In the upcoming 1.3+ there will be a new markdown editor, which should also have a better wiki page link support.

olekrisek:
OK - Where do you have Markdown in non-wiki editors ? I am a bit new to HumHub, but there is maybe other plugins ? - Have seen a couple of different versions of this JavaScript in the source code.

buddh4:
The problem is, that the wiki is a custom module, not everyone installed this module, so the core should not depend on any non core modules.

As luke said, lets wait for the wiki 1.3 integration. The new richtext is rather flexible and is able to add plugins only for specific presets e.g. wiki.

PR #3215, opened 2018-07-15
Small clean

acs-ferreira:


protected/humhub/modules/notification/controllers/OverviewController.php
```diff
@@ -8,10 +8,11 @@
 
 namespace humhub\modules\notification\controllers;
 
-use Yii;
 use humhub\components\Controller;
+use humhub\components\behaviors\AccessControl;;
```
Felli: Added one too many `;`

protected/humhub/modules/comment/Events.php
```diff
@@ -89,7 +88,7 @@ public static function onWallEntryLinksInit($event)
             return;
         }
 
-        if (\Yii::$app->getModule('comment')->canComment($event->sender->object->content)) {
+        if (Yii::$app->getModule('comment')->canComment($event->sender->object->content)) {
```
Felli: Wouldn't it be better to use `!Yii` here?

PR #3225, opened 2018-07-25
Simple clean

acs-ferreira:


protected/humhub/modules/search/behaviors/Searchable.php
```diff
@@ -12,6 +12,7 @@
 use yii\db\ActiveRecord;
 use yii\base\Behavior;
 use yii\console\Exception;
+use humhub\modules\search\interfaces\Searchable;
```
luke-: Conflict?

luke-:
Seems there is a conflict with the searchable import?

protected/humhub/modules/search/behaviors/Searchable.php
Felli: Does changing it to `use \humhub\modules\search\interfaces\Searchable;` fix the conflict?

danielkesselberg: Conflict here is class Searchable and use Searchable. Just import Searchable like this:
use humhub\modules\search\interfaces\Searchable as SearchableInterface and change if to $this->owner instanceof SearchableInterface.

PR #3234, opened 2018-08-05
Improve moduleloader

danielkesselberg:
**Dangerous change:** Removed conditional caching from ModuleAutoLoader

### DIFF
--- a/protected/humhub/components/bootstrap/ModuleAutoLoader.php
+++ b/protected/humhub/components/bootstrap/ModuleAutoLoader.php
@@ -10,6 +10,7 @@ namespace humhub\components\bootstrap;
 use humhub\components\Application;
 use Yii;
 use yii\base\BootstrapInterface;
+use yii\helpers\FileHelper;
 
 /**
  * ModuleAutoLoader automatically searches for config.php files in module folder an executes them.
@@ -19,6 +20,7 @@ use yii\base\BootstrapInterface;
 class ModuleAutoLoader implements BootstrapInterface
 {
     const CACHE_ID = 'module_configs';
+    const CONFIGURATION_FILE = 'config.php';
 
     /**
      * Bootstrap method to be called during application bootstrap stage.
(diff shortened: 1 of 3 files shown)


protected/humhub/components/bootstrap/ModuleAutoLoader.php
```diff
@@ -43,11 +62,52 @@ public function bootstrap($app)
...
+     * @return array
+     */
+    public static function findModulesByPath($path)
+    {
+        $hasConfigurationFile = function ($path) {
+            return is_file($path . DIRECTORY_SEPARATOR . self::CONFIGURATION_FILE);
+        };
+
+        try {
+            return FileHelper::findDirectories(Yii::getAlias($path, true), ['filter' => $hasConfigurationFile]);
```
githubjeka: https://www.yiiframework.com/doc/api/2.0/yii-helpers-basefilehelper#findDirectories()-detail

What about the case when

- moduleDir
   - tests
      - config.php
   - config.php

protected/humhub/components/bootstrap/ModuleAutoLoader.php
```diff
@@ -43,11 +62,52 @@ public function bootstrap($app)
                     }
                 }
             }
-            if (!YII_DEBUG) {
```
githubjeka: Can you describe what's wrong with this line?

githubjeka: > Conditional caching could lead to unexpected results when YII_DEBUG is
enabled while cache is not DummyCache

Ah, ok

protected/humhub/components/bootstrap/ModuleAutoLoader.php
danielkesselberg: Good catch. I will look into recursive true/false tomorrow.

protected/humhub/components/bootstrap/ModuleAutoLoader.php
```diff
@@ -19,8 +20,26 @@
...
     public function bootstrap($app)
+    {
+        $modules = self::locateModules();
+
+        Yii::$app->moduleManager->registerBulk($modules);
+    }
+
+    /**
+     * Find available modules
+     * @deprecated 1.3 replace call for locateModules with findModules and handle caching outside of method (e.g. in boostrap)
```
githubjeka: How about remove `@deprecated` and rewrite it by `findModules()`, now.

danielkesselberg:
@luke- Code Climate not happy but build is ok.

acs-ferreira:
@danielkesselberg I believe Codeclimate can be fixed using `getOrSet()` here.

danielkesselberg:
@acs-ferreira CodeClimate is unhappy because locateModules is way to complex. There is nothing we can do right now. I added `@depcreated` and added a cleaner version with some unit tests.

PR #3239, opened 2018-08-07
Catch IntegrityException in Notification

danielkesselberg:
This fix needs more testing. Source of problem still unknown.


luke-:
I've added also this on IntegrityExceptions:
https://github.com/humhub/humhub/commit/57e1d65157dda088463b917bd27bbb77ebc4cb40#diff-6789def14471dfa75422a1e4cb2d592cR84

Maybe it's better to delete these notifications and log it?

PR #3242, opened 2018-08-08
Add configuration for humhub (core) coverage

danielkesselberg:
Add configuration for humhub (core) to generate code coverage when running test suite

- Use utf8mb4
- Move logic to .travis/install-dependencies.sh & .travis/setup-humhub.sh like for module tests
- Coverage-Configuration for core
- Upload _output to transfer.sh on failure


danielkesselberg:
Stream-Acceptance-Tests fails when --coverage --coverage-xml passed

PR #3264, opened 2018-08-22
Prepare Yii 2.1

acs-ferreira:
- Replace `::className()` calls with `::class`
- Replace usages of `yii\base\InvalidParamException` with `yii\base\InvalidArgumentException` _(deprecated since 2.0.14)_


buddh4:
Thx I've included the change of `InvalidParamException` in https://github.com/humhub/humhub/commit/d41fd377d06cf5f2705918dd6a68c1d61fe953d4

Issue #3258, opened 2018-08-19
UI: Editor buttons


acsfer:
![image](https://user-images.githubusercontent.com/12234510/44309554-e674c880-a3c8-11e8-943c-91befe7ee21d.png)

Upload and save buttons are hover scroll bar (if post has scroll) when editing or writing.

PR #3268, opened 2018-08-28
Fixed CSS issue #3258 about .comment-buttons on content editor.

caos30:
I simply added a right padding:

imatge

I really am a newbie on LESS css so, i don't have very clear where to change the padding-right attribute. So i changed it on 2 files:

- static/less/comment.less
- themes/HumHub/css/theme.css


buddh4:
Unfortunately this looks a bit broken if no scrollbar is visible.

buddh4:
I'll close for now, hope we find a solution soon.

PR #3281, opened 2018-09-07
Fix: pt_br translation

Felli:
This is to fix an error seen [here](https://community.humhub.com/content/perma?id=162132).


buddh4:
Don't see any problems with the translation syntax, the only thing you should not translate is `plural` or `other`. Or did you notice any problems with this translation?

Felli:
When translating `file` or `files` I found that this caused an error which is why I make this P/R.  :+1:

buddh4:
But this is the purpose of the translation, see https://www.yiiframework.com/doc/guide/2.0/en/tutorial-i18n#plural

PR #3295, opened 2018-09-15
Fix Overflow on long filenames

acs-ferreira:
Should fix these overflows:
image

### DIFF
--- a/static/less/stream.less
+++ b/static/less/stream.less
@@ -44,6 +44,7 @@
             overflow:hidden;
             text-overflow:ellipsis;
             max-width: 100%;
+            display: inline-block;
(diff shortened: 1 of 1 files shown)


buddh4:
This destroyed the alignment of links within texts in my tests:

grafik

Don't think its a good idea to use inline-block on paragraphs either. Where does this occure on all posts or only on specific content types as file module etc. ?

acs-ferreira:
@buddh4 Having this problem with files module, specially on mobile devices.

buddh4:
We should review/compare the files module style rather than messing with the general post style, I think the overflow should work on posts.

PR #3330, opened 2018-10-14
add less options for mail font url/family

rekollekt:


luke-:
Thanks!

FYI I've changed: https://github.com/humhub/humhub/commit/cf2757389a967fc294cc9f8dd1ce24b16a12bc68#diff-3fa287ef0d0b959060a6c5a09ed3d3e7R91

rekollekt:
THX. `https://github.com/humhub/humhub-modules-mail/pull/150` It seems to me that mail font family does not appear in other free modules (but should be investigated). Non-free modules should also be tested investigated for mail font family issue. For example `protected/modules/massuserimport/views/mails/ImportUser.php` and `protected/modules/massuserimport/views/mails/InviteUser.php`.

PR #3360, opened 2018-11-06
Fix: Minor Grammar fix

Felli:
This makes the sentence slightly more correct.


buddh4:
This removed all current translations, we have to find a better way of correcting such texts.

luke-:
@Felli What do you think of this variant:

You can add a statistic code snippet (HTML) - which will be added to all rendered pages.

Felli:
Sounds fine to me.

Issue #3343, opened 2018-10-20
URL Rewriting (nginx)


devcline:
cant find any information how to configure fancy url on nginx
thanks for any help to port it.

Felli:
Sadly I don't think there is any official documentation just yet, but I would check out the following issue, with some work it should work for you. https://github.com/humhub/humhub/issues/257

devcline:
im searching converted from apache those strings 

```
  # ensure permalink when url rewriting was enabled (index.php?r=content/perma&id=6 => /content/perma/?id=6
    RewriteCond %{QUERY_STRING} ^r=content(/|%2)perma&id=([0-9]*)$
    RewriteRule ^index\.php$ %{REQUEST_URI}/content/perma/?id=%2 [R=302,L]

    RewriteCond %{REQUEST_URI}::$1 ^(/.+)/(.*)::\2$
    RewriteRule ^(.*) - [E=BASE:%1]
```

Felli:
The only thing that I could think of would be the following?
```
# nginx configuration

location ~ 403 {
  rewrite ^(.*)$ ^/?\.(?!/well-known/acme-challenge/[\w-]{43}$) redirect;
}

location ~ /\.well-known/acme-challenge/[0-9a-zA-Z_-]+$ {
}

location / {
  if ($query_string ~ "^r=content(/|%2)perma&id=([0-9]*)$"){
    rewrite ^/index\.php$ /$request_uri/content/perma/?id=%2 redirect;
  }
  if (-e $request_filename){
    rewrite .? /$env_base/index.php break;
  }
}
```

reymaro0:
well.. config from here didn’t work, but I tried only adding this lines to my vhost nginx and pretty urls are working well:
```
        location / {
                try_files $uri $uri/ /index.php?$args;
        }
```

fusir:
It's interesting that my nginx doesn't even have an $env_base

moversity:
Here's an excerpt of my nginx config that I crafted to fix the old links by redirection to the `r` query parameter, preserving all other parameters, after setting `enablePrettyUrl` to `true`.

```conf
location / {
    try_files $uri $uri/ /index.php$is_args$args;
}

location ~ ^/(protected|framework|themes/\w+/views|\.|uploads/file) {
    deny all;
}

location ~ ^/assets/.*\.php$ {
    deny all;
}

location ~ ^/(assets|static|themes|uploads) {
    expires 10d;
[...]

PR #3377, opened 2018-11-23
Added: NGINX Documentations

Felli:
See #3343


danielkesselberg:
https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/

Felli:
@danielkesselberg sadly the `if` clause seems a necessary evil in some cases, although using `try_files` instead haven't given me any results back, could you do some test for this?.

acs-ferreira:
It also lacks (at least) the protected folders (actually by `.htaccess` files)

luke-:
FYI: 

We will soon change the documentation to another system/repo. 
Then we will reorganize the documentation and also add details like NGINX like this.

luke-:
https://docs.humhub.org/docs/admin/server-setup#nginx

PR #3381, opened 2018-12-07
Enh: Add to cache schema

Felli:
With this P/R, it adds `ico`, `svg`, and `less` to the cache schema.


.htaccess.dist
```diff
@@ -45,7 +45,7 @@
 
 # Caching schema
 <IfModule mod_headers.c>
-    <FilesMatch "\.(jpg|png|gif|css|js)$">
+    <FilesMatch "\.(jpg|png|gif|ico|svg|css|js|less)$">
         Header set Cache-Control "private, max-age=3600"
```
Felli: @luke- Not sure if we should keep this private or change to `public`, either way, it should still work as intended, the biggest difference would be either allowing proxies to cache or not, changing to `public` might help with CloudFlare a little bit but not much, I'll leave this up to you. :thinking:

Felli: A better description of what I'm talking about can be found [here](https://stackoverflow.com/questions/3492319/private-vs-public-in-cache-control).

PR #3405, opened 2019-01-16
Enh: use LocalStorage in panel menu

helnokaly:
Use LocalStorage to save panel menu's state instead of cookies

**Why?**
- Because that data is needed only on the client side and cookies are sent to the server on each request.

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] When resolving a specific issue, it's referenced in the PR's title (e.g. `Fix #xxx[,#xxx]`, where "xxx" is the Github issue number)
- [ ] All tests are passingp
- [x] Changelog was modified

**Other information:**


helnokaly:
@Felli both Cookies and LocalStorage are client-side storage. But when using cookies to store Panel Menu's state, that data will be sent to the server with every request, wasting data bandwidth unnecessarily. That is why LocalStorage is preferred in that case.

Felli:
@luke- I think this is worth merging when the `CHANGLOG_DEV.md` file is updated.

danielkesselberg:
Looks good to me. Please add another check if localStorage is available (see "known issues" https://caniuse.com/#search=localstorage). Here are some examples: https://stackoverflow.com/questions/16427636/check-if-localstorage-is-available

Felli:
In regards to @danielkesselberg's post the following would cause a lot of issues with both the [Discordapp](https://github.com/GreenMeteor/humhub-discordapp-module) module and the custom pages module for those that still use IE 10 as both use iframe to some degree and both have the collapsing feature implemented as far as I know. So I thank you for bring this to the table and provides more discussion on this. :thinking:

> IE10 : The storage event is fired even on the originating document where it occurred. Causes problems with multiple windows websites, and huge problems with iframes.

danielkesselberg:
IE 10 is no longer relevant (https://caniuse.com/usage-table). For some browsers the `localstorage` object is not available when in private mode. Just cover this case by `try-catch` like suggested.

Felli:
Last question; is localStorage truly a safer option? Meaning is there any loop holes that could be exploited when using such a function? At the moment a number of things use the panel menu which would concern me if the protection of data couldn't be guaranteed with such a function.

> Note: At the moment I've already tested this within a test environment for HumHub v1.3.8 and it defaults to cookie regardless of using the modified code.

helnokaly:
> IE 10 is no longer relevant (https://caniuse.com/usage-table). For some browsers the `localstorage` object is not available when in private mode. Just cover this case by `try-catch` like suggested.

Thanks for bringing this up, so this issue affects iOS Safari up to including version 10.x, can't we consider these versions irrelevant now? :thinking: 
https://caniuse.com/#search=localstorage

danielkesselberg:
> Thanks for bringing this up, so this issue affects iOS Safari up to including version 10.x, can't we consider these versions irrelevant now? thinking
> https://caniuse.com/#search=localstorage

Sorry I didn't get that :thinking: 

> Here are some examples: https://stackoverflow.com/questions/16427636/check-if-localstorage-is-available

Please add `try-catch` around the `getItem`, `setItem` and `removeItem`. It's possible that `localstorage` is not available (e.g. in some private mode, disabled by addon or not supported by browser). 

I'm voting against a cookie fallback here. If localstorage is not available a user cannot use this feature. What do you think @buddh4?

buddh4:
Yes, I think if localStorage is not active we should just ignore this feature. In the future we will probably use the localStorage more in order to keep the UI state of such components.

PR #4148, opened 2020-05-22
Use settings->baseUrl for invitation links. Fix #1316

### LINKED ISSUE #1316: Wrong URL in users invite
Hello,

I just can't fix the URL which is in the user invites. Humhub is installed on my NAS and email invite keep sending a sign up link with a local IP in the URL... Found no entrey in database as humhum can be reach via à subdomain on my NAS...

marcodisarno:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**


luke-:
Thanks for your contribution.

Do you use HumHub under multiple URLs or how does the problem occur?

Maybe we should simply move the task of sending the invitation URL to an ActiveJob in the queue, then the BaseURL is automatically used.

marcodisarno:
Hey!

> Do you use HumHub under multiple URLs or how does the problem occur?
I'm using it in a docker (inspired by https://github.com/mriedmann/humhub-docker) that I then deploy automatically to different hosts, set by `php yii installer/set-base-url <HOST>`. The containers then are proxy-passed to by a front-door nginx.

Note: in my particular case I could also just add configurations for the target hosts and switch those during deploy:

```
'urlManager' => [
    'hostInfo' => 'http://example.com/',
    // ...
]
```

But it would feel like double-configuring the same thing. And this open issue #1316 convinced me this needs something a little more general.

> Maybe we should simply move the task of sending the invitation URL to an ActiveJob in the queue, then the BaseURL is automatically used.

I'm not very familiar with Yii so I don't know how this solves the problem, but I trust you do. Plus, I think it is a very good idea to have these invitations in a queue. Especially since you can enter comma-seperated emails in the Form, which (haven't checked) might not be restricted.

marcodisarno:
Hi again, it happens to other Email notifications as well, e.g. have I just tried to change the top admin (id=1) e-mail address, which summons a confirmation mail to the new address.

So we need something really "general" here.

luke-:
Did you configured the FrontEnd proxy in the config:
https://docs.humhub.org/docs/admin/reverse-proxy#trusted-hosts

You can also use the word `any` instead of IP for testing.

PR #4164, opened 2020-06-10
WIP: Create content title 

luke-:
Add a common ``title`` attribute for all types of content in HumHub.

Additionally it tries to simplify the usage of ``ContentActiveRecord`` for module developers.

Targets:

- Access to the common ``Content model`` attributes like title, container, visibility can be done directly via the ContentActiveRecord

- The content attribute (``$post->content->...``)  should not longer used and is not longer required for standard use cases.

- Validation and Error Handling of the Content attributes (e.g. title, visibiltiy) can be done without any double handling (MyContentType & Content)
e.g. https://github.com/humhub/humhub/blob/master/protected/humhub/modules/content/components/ContentActiveRecord.php#L373-L377


buddh4:
Some issues to discuss.

### Define use-cases

Before we add a central content title field and furthermore force content types to support a title, we should define actual use-cases describing where we'll plan to use this feature. In my opinion the ContentActiveRecord class already supports too much default features which do not make sense or are not supported by some content types, e.g. pining/archiving of custom pages and snippets etc. This either restricts possible use-cases of content types or is prone to data inconistencies and pollutes content type classes with features and functions not supported by a type.

So what do we try to achieve/enhance with content titles.

### Benefits of central title field in content table?

What are the actual advantages of a central title field in the content table?

- Content types which do already support titles use a custom title field (instead of content table title field). 
- Some content types may generate a translatable title instead of a user generated title.
- Some content types may not support content titles.

### Does this feature fit for all existing (or future) content types?

 - Activities - A rather special content type, does not have a real title
 - Cfiles - has a filename, we'd need to sync the filename and title
 - Translation Log - Translation module, is created automatically and does not support title
 - Gallery - has optional title, and needs to be added after upload
 - Auto generated content
 - Sub content types

- Furthermore most content types which do already have title support use a custom title field (instead of content table title field). 
- Some content types may generate a translatable title instead of requiring a user generated title.

### Other solution

We may should only implement a content title for posts and think about the mentioned issues before forcing a content
title. Depending on the use cases we could think about a interface for this feature, this way we could support custom content titles as translatable titles and custom title fields.

Issue #4003, opened 2020-04-16
Show user email address in profile (Option) 


luke-:
### Is your feature request related to a problem? Please describe.

Add option to show users e-mail address on the profile page

### Describe the solution you'd like

- Add profile field type: email
- Other place?

PR #4165, opened 2020-06-12
Show users email in profile (#4003)

luke-:

- [ ] Bugfix
- [X] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**


buddh4:
The virtual email field is only available for new installation isn't it? Do we need a migration for existing installations?

protected/humhub/modules/user/models/fieldtype/UserEmail.php
```diff
@@ -0,0 +1,37 @@
...
+     * @inheritDoc
+     */
+    public function getUserValue($user, $raw = true)
+    {
+        if (empty($user->email)) {
+            return '';
+        }
+
+        if ($raw) {
+            return Html::encode($user->email);
```
buddh4, line 32: Shouldn't raw return an unencoded value?

protected/humhub/modules/user/models/fieldtype/BaseTypeVirtual.php
```diff
@@ -0,0 +1,66 @@
...
+     */
+    public function rules()
+    {
+        return [];
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function getUserValue($user, $raw = true)
```
buddh4, line 38: Why not abstract?

protected/humhub/modules/user/models/fieldtype/BaseTypeVirtual.php
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * @inheritdoc
+     */
+    public $isVirtual = true;
+
+    /**
+     * @inheritdoc
+     */
+    public function rules()
```
buddh4, line 30: Is the same as parent `yii\base\Model`

buddh4:
When creating a virtual field you can still select **editable**, **show at registration** etc. even though those settings are not allowed and do not show up when editing such a field.

protected/humhub/modules/user/models/fieldtype/BaseTypeVirtual.php
buddh4: Ok, unfortunately the parent function is not abstract. Maybe we could use a template method to force subclasses to implement a value getter e.g. `abstract getVirtualUserValue()`

luke-:
> When creating a virtual field you can still select **editable**, **show at registration** etc. even though those settings are not allowed and do not show up when editing such a field.
> 
> +See other comments

Unfortunately, this cannot be avoided during creation, but will otherwise be hidden.

protected/humhub/modules/user/models/fieldtype/BaseTypeVirtual.php
luke-: Good idea!

protected/humhub/modules/user/models/fieldtype/UserEmail.php
luke-: It's always encoded but without any extras like Link tags.
https://github.com/humhub/humhub/blob/master/protected/humhub/modules/user/models/fieldtype/Text.php#L203

Issue #3990, opened 2020-04-10
Implement StrictController


buddh4:
### Is your feature request related to a problem? Please describe.

Beside ContentContainerController there currently is only the base Controller class available, which does not use StrictAccess by default. StrictAccess handles guest user control if the guest mode is not active and currently needs to be manually set as follows:

```php
public $access = StrictAccess::class
```

Since the base Controller class is also in use in cases where non protected external access is required e.g. for sharing documents or jitsi meetings with non members. We should add a StrictAccessController which should be used for most non container related actions and adds strict access control by default and only use the base Controller if we need non protected access from non members.

luke-:
Strict should be default in 1.6

buddh4:
So we'll set strict access in the base humhub controller class. Should only require a small migration in rare cases.

luke-:
Case Overview:

Case 1: Everybody (Not logged in)                                             (Login, Sharing)
Case 2: Logged in                                                                       (Primarily user actions)
Case 3: Logged in & Guests (if enabled)                                    (View stuff)

PR #4176, opened 2020-06-22
Enh #3990: Use StrictAccess by default in humhub\components\Controller

buddh4:
Implements: https://github.com/humhub/humhub/issues/3990

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

Controllers requiring guest access even though guest mode is not active need to explicitly configure:

```
public $access = ControllerAccess::class;
```

- [x] All tests are passing
- [x] Changelog was modified

### DIFF
--- a/protected/humhub/components/Controller.php
+++ b/protected/humhub/components/Controller.php
@@ -9,6 +9,7 @@
 namespace humhub\components;
 
 use humhub\components\access\ControllerAccess;
+use humhub\components\access\StrictAccess;
 use humhub\components\behaviors\AccessControl;
 use Yii;
 use yii\helpers\Url;
@@ -56,7 +57,7 @@ class Controller extends \yii\web\Controller
      * @var string defines the ControllerAccess class for this controller responsible for managing access rules
      * @see self::getAccess()
      */
-    protected $access = ControllerAccess::class;
+    protected $access = StrictAccess::class;
 
     /**
      * Returns access rules for the standard access control behavior.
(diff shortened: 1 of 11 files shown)


luke-:
Just to recap:

There are 3 common modes:
1.) No access control
2.) No access control when guest mode is activated
3.) Only logged in users 

StrictMode is case 2, isn't it?

What is the current default if nothing is explicitly stated?  
(In my opinion, case 3 should be the default because it is the safest mode.)

buddh4:
Currently, a limited set of access control checks are always active. There are two base ControllerAccess types in use which dictate a different set of fixed rules, which are always active:

The base [ControllerAccess](https://github.com/humhub/humhub/blob/master/protected/humhub/components/access/ControllerAccess.php) which was the default prior to this change, only has the following fixed access validations:

 - Deny access for disabled users
 - Deny access for unapproved users

Without any additional rules this access check allows guest users even though guest mode is not active, which is required for controllers as Auth or Registration controllers as well as Controller used to externally share resources as documents etc.

The [StrictAccess](https://github.com/humhub/humhub/blob/master/protected/humhub/components/access/StrictAccess.php) adds the following fixed rule:

 - Deny access for guests if guest mode is not active

After this PR, in order to grant guest access even with deactivated guest mode, a controller needs to manually switch to the base ControllerAccess:

```
public class AllAccessController extends Controller
{
    public $access = ControllerAccess::class;
}
```

Thats the only migration case of this PR.

PR #4179, opened 2020-06-23
May CHARACTER SET be utf8mb4?

tomofu74:
and COLLATE utf8mb4_unicode_ci?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**

### DIFF
--- a/protected/humhub/modules/admin/migrations/m141106_185632_log_init.php
+++ b/protected/humhub/modules/admin/migrations/m141106_185632_log_init.php
@@ -63,12 +63,6 @@ class m141106_185632_log_init extends Migration
         foreach ($this->getDbTargets() as $target) {
             $this->db = $target->db;
 
-            $tableOptions = null;
-            if ($this->db->driverName === 'mysql') {
-                // http://stackoverflow.com/questions/766809/whats-the-difference-between-utf8-general-ci-and-utf8-unicode-ci
-                $tableOptions = 'CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci';
-            }
-
             $this->createTable($target->logTable, [
                 'id' => $this->bigPrimaryKey(),
                 'level' => $this->integer(),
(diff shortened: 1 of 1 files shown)


buddh4:
Could you please elaborate, which bug or issue this PR is exactly fixing? Do you encounter any specific problems regarding the log table and used character set?

ArchBlood:
I believe from looking at the Stack Overflow line comment in the code itself speaks of why he changed the character set and log table, not really needed but wouldn't really hurt for newer versions of MySQL.

> https://stackoverflow.com/questions/766809/whats-the-difference-between-utf8-general-ci-and-utf8-unicode-ci

luke-:
Is there any reason to not switch to `utf8mb4_unicode_ci` here?
This collation is changed later anyway.

But this PR should rather go to the `develop` branch if it is not a fix.

buddh4:
Why is this even part of the log table migration? I don't see any problem with switching the character set, but ideally a bugfix should describe usecases which are fixed by the change. Furthermore interesting is the following of the stack overflow answer:

> Note: Newer versions of MySQL have updated Unicode sorting rules, available under names such as utf8mb4_0900_ai_ci for equivalent rules based on Unicode 9.0 - and with no equivalent _general variant. People reading this now should probably use one of these newer collations instead of either _unicode or _general. Much of what's written below is not of much interest anymore if you can use one of the newer collations instead.

https://stackoverflow.com/questions/54885178/whats-the-difference-between-utf8-unicode-ci-and-utf8mb4-0900-ai-ci

Not sure about performance indications when switching to `utf8mb4_0900_ai_ci`

ArchBlood:
@buddh4 general performance is around the same just seems faster, overall the answer is pretty accurate.

What I took a note of from the answer of that Stack Overflow question was this.

> However: The speed of collation is usually the least of the performance issues in queries. INDEXes, JOINs, subqueries, table scans, etc are much more critical to performance.

Overall `utf8mb4_0900_ai_ci` would be good for posts/comments/activities/notifications as well as emojis but not for much of anything else.

luke-:
@buddh4 You're right, we should just remove the character set addition.

luke- (approved):

PR #4193, opened 2020-06-29
Nested Comments

luke-:

- [ ] Bugfix
- [X] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**


ArchBlood:
Works really well, although I don't think it should open in an model action popup, would defeat the purpose of the comment section, not only this but it breaks the display of the Richtext editor by forcing to show the full editor buttons and on fullscreen it breaks the display completely.

Screenshot_1
.

ArchBlood:
MUCH better! Although some issues that could be fix as shown in the screenshot. :smile:
Screenshot_1

### Error on delete
```log
yii\base\Exception: Invalid class type! () in /home/USERNAME/public_html/protected/humhub/libs/Helpers.php:210 Stack trace:
#0 /home/USERNAME/public_html/protected/humhub/modules/comment/controllers/CommentController.php(76): humhub\libs\Helpers::CheckClassType('', Array)
#1 /home/USERNAME/public_html/protected/vendor/yiisoft/yii2/base/Controller.php(155): humhub\modules\comment\controllers\CommentController->beforeAction(Object(yii\base\InlineAction))
#2 /home/USERNAME/public_html/protected/vendor/yiisoft/yii2/base/Module.php(528): yii\base\Controller->runAction('delete', Array)
#3 /home/USERNAME/public_html/protected/vendor/yiisoft/yii2/web/Application.php(103): yii\base\Module->runAction('comment/comment...', Array)
#4 /home/USERNAME/public_html/protected/vendor/yiisoft/yii2/base/Application.php(386): yii\web\Application->handleRequest(Object(humhub\components\Request))
#5 /home/USERNAME/public_html/index.php(25): yii\base\Application->run()
#6 {main}
```

buddh4:
The richtext issue is rather a theming problem, the problem is that the fullscreen richtext uses absolute position and width/height of the viewport, which does not work when rendered within relative positioned elements.

protected/humhub/modules/comment/widgets/views/form.php
```diff
@@ -7,18 +7,25 @@
...
+/* @var $objectModel string */
+/* @var $objectId integer */
+/* @var $id string unique object id */
 
+/** @var \humhub\modules\content\Module $contentModule */
+$contentModule = Yii::$app->getModule('content');
 $submitUrl = Url::to(['/comment/comment/post']);
 
+// Hide the comment form for sub comments until the button is clicked
+$isHidden = ($objectModel === \humhub\modules\comment\models\Comment::class);
```
ArchBlood, line 20: Should be else it hides on click.
```php
$isHidden = ($objectModel !== Comment::class);
```

ArchBlood: Example of it working with above modification.
Screenshot_4

ArchBlood: This completely fixed the issue where comment input disappears.

```php
$isHidden = ($objectId === Comment::class);
```

But as is if you use the `$objectModel` it will force you to only comment and the input form is completely hidden because of `===` and only shown with `!==`.
If `$objectId` is used with `===` then the issue disappears completely and you can use both inputs.

luke-: Did you empty the cache (Browser & Assets) and rebuild the assets if necessary? Normally only the form of nested comments should be hidden.

ArchBlood: Yes did this and the problem stayed the same till modified.

luke-: The current expected behavior is.

After initial stream load:
image

After click on "Comment":
image

buddh4:
The notification of sub comments mentions the post and not the main comment:

grafik

grafik

Maybe a text like: 'xy replied to your comment "blabla"' would be more significant.

But not a big issue in my opinion.

protected/humhub/libs/Helpers.php
```diff
@@ -190,12 +197,12 @@ public static function CheckClassType($className, $type = '')
 
         if (is_array($type)) {
             foreach ($type as $t) {
-                if (class_exists($className) && is_subclass_of($className, $t)) {
+                if (class_exists($className) && (is_subclass_of($className, $t) || $className === $t)) {
```
buddh4, line 200: Why not using [is_a](https://www.php.net/manual/en/function.is-a.php)

buddh4:
When clicking on `edit` on the main comment, the sub comment edit will also trigger:

grafik

buddh4:
You can not attach files when editing a comment or sub comment:

Edit:

grafik

After submit:

grafik

buddh4:
Should test this also in calendar + task module + comment modal.

buddh4:
I'd suggest changing the text from `comment` to `reply` in case of sub comments.

protected/humhub/modules/comment/widgets/views/form.php
ArchBlood: Yes, only issue is with this method the input box does not appear at all, even on flushing cache. :thinking:

protected/humhub/libs/Helpers.php
```diff
@@ -197,12 +197,12 @@ public static function CheckClassType($className, $type = '')
```
ArchBlood, line 197: Gives a `PHP Notice:  Undefined variable: t`

Below fixes this

```php
    /**
     * Checks if the class has this class as one of its parents
     *
     * @param string $className
     * @param string $type
     * @param string $t
     * @return boolean
     */
    public static function CheckClassType($className, $t, $type = '')
    {
        $className = preg_replace('/[^a-z0-9_\-\\\]/i', '', $className);

        if (is_array($type)) {
            foreach ($type as $t) {
                if (class_exists($className) && is_a($className, $t, true)) {
                    return true;
                }
            }
        } else {
            if (class_exists($className) && is_a($className, $t, true)) {
                return true;
            }
        }

        throw new Exception("Invalid class type! (" . $className . ")");
    }
```

ArchBlood: But then you'll get this error once this is fixed
`is_a() expects parameter 2 to be string, array given `

buddh4: The second is_a check is wrong and should use type parameter. This function also needs some proper doc.

luke-: Thanks!

PR #4197, opened 2020-07-04
Why missing 'en-US', 'en-GB'?

tomofu74:
We need to translate options... don't we?

- [ ] Bugfix
- [x ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

system messages required to rebuild.

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**

### DIFF
--- a/protected/humhub/config/i18n.php
+++ b/protected/humhub/config/i18n.php
@@ -12,6 +12,8 @@ return [
     // array, required, list of language codes that the extracted messages
     // should be translated to. For example, ['zh-CN', 'de'].
     'languages' => [
+        'en-US',
+        'en-GB',
         'de',
         'fr',
         'nl',


luke-:
The main difference between "en-US" and "en-GB" are currently Locale specific details like the date format.

However, it is possible to overwrite individual terms in the different variants using a local translation file.
I would currently rather not create all language files for this automatically. Also to keep the number of files as low as possible.

I'm also not sure that our translation community software supports this.

ArchBlood:
> The main difference between "en-US" and "en-GB" are currently Locale specific details like the date format.
> 
> However, it is possible to overwrite individual terms in the different variants using a local translation file.
> I would currently rather not create all language files for this automatically. Also to keep the number of files as low as possible.
> 
> I'm also not sure that our translation community software supports this.

Not only this but English is already a default language that's used throughout the whole project so adding it wouldn't really matter. :thinking:

tomofu74:
Yes, English is already default and it wouldn't really matter to use system.
But, Humhub announces Available translations list shows "en-UK", though it is not appeared on Language selection list on default system, isn't it ? https://docs.humhub.org/docs/admin/translations/
By the way, we just only want to use specific wording for specific member with this English choice. So it is useful specific English member.

ArchBlood:
To be honest `en-UK` should be changed to `en-GB` because this is the actual locale code for the UK, so that is a typo in the docs, but you can already use the translation overwrite & limited allowed languages even without them being added to `i18n.php` so I'm still not seeing actual use of it being added. If something in English needs overwritten then do it through the overwrite option, if you see a grammar issue or typo then you'll have to find the exact file that has the English wording needed changed, as there are no `/messages/en-US` or `/messages/en-GB` source file directories to these languages. :thinking:

### Profile Languages
Screenshot_2

### Global Languages
Screenshot_1

### Limited Allowed Languages
Screenshot_3

luke-:
If there are several wording differences between en-US (default) and en-GB, we can also create 'en-GB' language files.
But if possible I would try to find a consistent wording for both language variants.

ArchBlood:
> If there are several wording differences between en-US (default) and en-GB, we can also create 'en-GB' language files.
> But if possible I would try to find a consistent wording for both language variants.

@luke- I believe this [here](https://tinyurl.com/y5wkuete) hits the nail on the head when differentiating the two English languages

luke-:
@ArchBlood thanks!

I think the most important are the date & time formats.  

We should try to find a wording that fit well for both language variants (US & GB). If this is not possible, we should create these particular language files manually.

@tomofu74 Would that solve your issue?

tomofu74:
That wil be OK and We'll try to translate each time according to the customer's request.
Thank you ArchBlood, loke-.

luke-:
@tomofu74 Great, please feel free to create a PR with the changed wordings.  Thanks!

PR #4270, opened 2020-07-27
Fix #4263: New ContentVisibilitySelect is missing a CreatePublicConte…

### LINKED ISSUE #4263: New ContentVisibilitySelect is missing a CreatePublicContent permission check
The ContentVisibilitySelect form widget introduced with 1.6 does not check for CreatePublicContent permission on the container.

luke-:
…nt permission check

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

- [ ] All tests are passing
- [X] Changelog was modified

**Other information:**

### DIFF
--- a/protected/humhub/modules/ui/form/widgets/ContentVisibilitySelect.php
+++ b/protected/humhub/modules/ui/form/widgets/ContentVisibilitySelect.php
@@ -42,7 +42,7 @@ use yii\bootstrap\Html;
 class ContentVisibilitySelect extends InputWidget
 {
     /**
-     * @var bool Automatically hide the field when no only one visibility option is available
+     * @var bool Hide the field when only one visibility option is available
      */
     public $autoHide = true;
 
--- a/protected/humhub/modules/ui/form/widgets/ContentVisibilitySelect.php
+++ b/protected/humhub/modules/ui/form/widgets/ContentVisibilitySelect.php
@@ -11,6 +11,7 @@ namespace humhub\modules\ui\form\widgets;
 use humhub\modules\content\components\ContentActiveRecord;
 use humhub\modules\content\components\ContentContainerActiveRecord;
 use humhub\modules\content\models\Content;
+use humhub\modules\content\permissions\CreatePublicContent;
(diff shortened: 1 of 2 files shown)


protected/humhub/modules/ui/form/widgets/ContentVisibilitySelect.php
```diff
@@ -41,7 +42,7 @@
 class ContentVisibilitySelect extends InputWidget
 {
     /**
-     * @var bool Automatically hide the field when no mulitple visibility modes available
+     * @var bool Automatically hide the field when no only one visibility option is available
```
ArchBlood, line 45: Probably could change this to something simpler
`Hide the field when only one visibility option is available` or
`Automatically hide the field when only one visibility option is available`

luke-: Ok, good point!

buddh4:
When using this we need to be aware that this field may be not be set, could this have any sideeffects, perhaps there we should suggest setting a default value? Also we should add some default validation within the content layer or force a default visibility in `Content::beforeSave` in such cases?

Currently ther only is:

```php
if (!$this->visibility) {
   $this->visibility = self::VISIBILITY_PRIVATE;
}
```

luke-:
@buddh4 I'm not sure I get your point. We could always render at least one hidden input. This should avoid empty values. Additionally, a validation in the model must of course take place.

buddh4:
My point is, that each form model which loads the visibility value needs to manually implement a validation of the value against the same visibility rules implemented in this widget, which leads to redundancy. Therefore there should be a validator or helper in the core. And we should either block invalid visibility settings or overwrite the visibility if needed on content class level to avoid inconsistent data as for example public content on private spaces.

luke-:
Okay, that makes sense. Then we drop the check/validation in the form field until we have a validator that we can use in both the model and the form field.

PR #4339, opened 2020-08-29
canLike permissions added for owner, member, moderator, admin

MasterMindNET:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**
Should we show the likes if the user does not have permission for this module(like)?


MasterMindNET:
in last commits check-in controller added; CSS file added with the disabled class(which is shown on the screenshot down below)
image
Why is the first post is inactive, but the other is active. 

Thank you!

protected/humhub/modules/like/permissions/CanLike.php
```diff
@@ -0,0 +1,59 @@
...
+    /**
+     * @inheritdoc
+     */
+    protected $moduleId = 'like';
+
+    public function __construct($config = [])
+    {
+        parent::__construct($config);
+
+        $this->title = \Yii::t('LikeModule.permissions', $this->title);
```
luke-, line 55: Such call with the message as variable are not possible. 
Otherwise the Yii2 message command cannot extract the message.

luke- (changes requested):
Attached are a few remarks, otherwise really good work! Thanks for the contribution!

protected/humhub/modules/like/permissions/CanLike.php
MasterMindNET: Sorry for some extra mistakes and a lack of attention to some new details.
I would take this into my account.
Thank you!

protected/humhub/modules/like/permissions/CanLike.php
```diff
@@ -0,0 +1,59 @@
+<?php
+
+/**
+ * @link https://www.humhub.org/
+ * @copyright Copyright (c) 2017 HumHub GmbH & Co. KG
```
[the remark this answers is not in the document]
MasterMindNET: I would take this into my account; thank you!

protected/humhub/modules/like/controllers/LikeController.php
```diff
@@ -78,6 +83,9 @@ public function actionLike()
      */
     public function actionUnlike()
     {
+        if (!$this->module->canLike($this->parentContent)) {
+            throw new ForbiddenHttpException(Yii::t('LikeModule.base', 'You are not allowed to like.'));
```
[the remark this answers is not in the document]
MasterMindNET: Yeah, I will remove it in the next commit.

protected/humhub/modules/like/Module.php
```diff
@@ -8,7 +8,11 @@
 
 namespace humhub\modules\like;
 
+use humhub\debug\RDebug;
```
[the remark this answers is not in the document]
MasterMindNET: I am using it locally for debugging.

I will remove it. These commits are subject for testings.

Could you please take a look at the images, and try to give some hint or ideas what should I check to make sure it works as intended?

I mean I am not able to like some of my own posts [here](https://user-images.githubusercontent.com/22347676/91640473-7fae2f00-ea26-11ea-98f6-c65afc6f026f.png)
and image

protected/humhub/modules/like/Module.php
```diff
@@ -58,4 +76,25 @@ public function getNotifications()
         ];
     }
 
+    /**
+     * Checks if given content object can be liked
+     *
+     * @param Like|ContentActiveRecord $object
+     * @return boolean can like
+     */
+    public function canLike($object)
```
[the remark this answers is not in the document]
MasterMindNET: [nested example in space and on the profile page](https://user-images.githubusercontent.com/22347676/91993679-56e7ab80-ed3e-11ea-9dc5-cdb11d3c0abc.png)
I need some help with it ( and the comment [above ](https://github.com/humhub/humhub/pull/4339#discussion_r482045174))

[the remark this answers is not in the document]
buddh4: All other can* methods are part of the content class, not sure if its a good idea to implement this one in the module class. Think we should keep them together.

buddh4 (commented):
Doesn't it make more sense to use a global user group permission instead of a container group permission?

protected/humhub/modules/like/assets/LikeAsset.php
```diff
@@ -23,6 +23,13 @@ class LikeAsset extends AssetBundle
      */
     public $sourcePath = '@like/resources';
 
+    /**
+     * @inheritdoc
+     */
+    public $css = [
+        'css/humhub.like.css'
```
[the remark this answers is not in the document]
MasterMindNET: Good, I will move this CSS code to `static/less/theme` and I will these extra lines.

Thank you for your help!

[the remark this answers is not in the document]
MasterMindNET: [All other can* methods are part of the content class](https://github.com/humhub/humhub/pull/4339#discussion_r483680644)
Could you please specify the path in which file I should move this method, please?
I did it by `protected/humhub/modules/comment/Module.php` example

[Example](https://github.com/humhub/humhub/blob/master/protected/humhub/modules/comment/Module.php#L69-#L91)


> Doesn't it make more sense to use a global user group permission instead of a container group permission?
Something like this:
`Yii::$app->user->can(ManageSpaces::class)`? or how?
Thank you! Hope you may understand me.

protected/humhub/modules/like/Module.php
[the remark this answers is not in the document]
luke-: @buddh4 I think this approach is ok. On the one hand it is solved identical to the comments module (canComment) and on the other hand not only content can be commented.

protected/humhub/modules/like/permissions/CanLike.php
```diff
@@ -0,0 +1,59 @@
...
+        Space::USERGROUP_ADMIN,
+        Space::USERGROUP_MODERATOR,
+        Space::USERGROUP_MEMBER,
+    ];
+
+    /**
+     * @inheritdoc
+     */
+    protected $fixedGroups = [
+        Space::USERGROUP_GUEST,
```
luke-, line 33: I think also non space members should be able to like.

The standard behavior should not change after the PR. It should only be an additional option to deny likes.

luke- (changes requested):
From my side we can merge the PR afterwards.

Is there anything else that prevents this?

protected/humhub/modules/like/permissions/CanLike.php
MasterMindNET: Please take a look at the[ comment above](https://github.com/humhub/humhub/pull/4339#discussion_r482096038).
I need some help with this. what could be  wrong  with it (Why I am not able to like my posts and in some other cases)

luke-: @MasterMindNET  This should solves the problem with the user profile related likes:

```
+++ b/protected/humhub/modules/like/permissions/CanLike.php
@@ -9,6 +9,7 @@
 namespace humhub\modules\like\permissions;

 use humhub\modules\space\models\Space;
+use humhub\modules\user\models\User;

 /**
  * CanLike Permission
@@ -20,6 +21,9 @@ class CanLike extends \humhub\libs\BasePermission
      * @inheritdoc
      */
     public $defaultAllowedGroups = [
+        User::USERGROUP_SELF,
+        User::USERGROUP_FRIEND,
+        User::USERGROUP_USER,
         Space::USERGROUP_OWNER,
         Space::USERGROUP_ADMIN,
         Space::USERGROUP_MODERATOR,
```

MasterMindNET: Oh, Thank you so much for this, I could not find it before 😃 
Please advise me which lines I should remove and which I should leave as a default?

I am really happy that everything is working just fine, but these lines were a real solution.

MasterMindNET: Oh, Thank you so much for this, I could not find it before 😃 
Please advise me which lines I should remove and which I should leave as a default?

I am really happy that everything is working just fine, but these lines were a real solution.

I am looking forward to hearing from you follow-up instructions

luke-:
Very good, you can adjust the CHANGELOG_DEV.md and then we can merge the PR.

protected/humhub/modules/like/controllers/LikeController.php
```diff
@@ -78,6 +83,9 @@ public function actionLike()
      */
     public function actionUnlike()
     {
+        if (!$this->module->canLike($this->parentContent)) {
```
buddh4, line 86: Maybe unlike should always be possible, a user should always be able to revert such actions? But I think this would only be an edge case.

static/less/humhub.less
```diff
@@ -250,3 +250,10 @@ ul.tag_input {
   line-height: 8px;
 }
 
+
+span.likeLinkContainer .like.disabled, span.likeLinkContainer .unlike.disabled {
```
buddh4, line 254: humhub.less not the best place for this, rather use base.less, which is also not ideal but I did not find a better place either.

protected/humhub/modules/like/controllers/LikeController.php
MasterMindNET: > Maybe unlike should always be possible, a user should always be able to revert such actions? But I think this would only be an edge case.

All right, I will remove this check for `actionUnlike`. Thank you for this note

static/less/humhub.less
MasterMindNET: As you please, I have moved it.
If you would find a better place to store it, just let me know, and I will move it to another place if it is needed.

luke- (approved):

Issue #4265, opened 2020-07-22
Using deprecated View class


ArchBlood:
Looking through some of the files I've noticed that some files are still using `humhub\components\View` which was deprecated in v1.3, is there any reason to still be using it?
https://github.com/humhub/humhub/blob/40bcad188bd3a7c270b645c74afe90db53457baa/protected/humhub/components/View.php#L16

#### Example
This file `profileHeader.php` still uses it despite it needing to use `humhub\modules\ui\view\components\View`

https://github.com/humhub/humhub/blob/40bcad188bd3a7c270b645c74afe90db53457baa/protected/humhub/modules/user/widgets/views/profileHeader.php#L5

PR #4340, opened 2020-09-02
replace old deprecated "\humhub\components\View " reference 

sahartak2025:
https://github.com/humhub/humhub/issues/4265 
replace old deprecated "\humhub\components\View " reference to the new \humhub\modules\ui\view\components\View reference


luke- (approved):

ArchBlood:
Glad that this was finally taken care of, of course, we'll also have to do the same in a few modules as well but that can be done at a later date

luke-:
@ArchBlood I also changed this for most modules.

PR #4344, opened 2020-09-03
Enh: Added unit tests in live/activity/comment module.

buddh4:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No


luke-:
I think we should only hotfixes into the `master` Branch.
Can we merge this into the `develop` branch?  Or is there something against this?

ArchBlood:
Normally you'd only want live branches to have tests, and seeing as develop branch is more lively it should take priority on tests than master

buddh4:
Those tests were written prior to 1.6 release. As long as the tests pass I don't see any issues adding them to master. Would rather argue why such tests did not exist prior 1.6 release. But I can also push them to dev.

buddh4:
The tests in this pr do not really add much coverage or are very important, but generally I think tests should be part as early as possible. Adding tests to master assures that the code works in v1.6.2+ while adding them to dev assures it works in v1.7.0+. New tests do not break existing code or compatibility. Also hotfixes should be come with tests assuring the fix works...

luke-:
I'm mainly interested in a clear policy with the branches. This PR can of course merged into master without any problems.

Regarding the policy, I would prefer to have only important fixes (incl. corresponding tests) and no other new code in this branch.
Of course, we can also generally allow new test code here.

luke-:
@buddh4 You can easily rebase in GitHub, just edit the PR and change the target.

PR #4345, opened 2020-09-03
Console User/Admin Controller (for Docker)

### LINKED ISSUE #3621: [Feature] Command line test
### Describe the solution you'd like
The `InstallController` is currently used for integration tests, but it could be used for command line install of HumHub (already the case in [this docker image)](https://github.com/mriedmann/humhub-docker#humhub_auto_install).

To do that, the installer would need some external parameters (with the same default value as now) for the following fields:
* instance settings
https://github.com/humhub/humhub/blob/f072f95321e1fab18e8533bf4e80860cc24866f6/protected/humhub/modules/installer/commands/InstallController.php#L36-L39
* admin username
https://github.com/humhub/humhub/blob/f072f95321e1fab18e8533bf4e80860cc24866f6/protected/humhub/modules/installer/commands/InstallController.php#L43
* admin email
https://github.com/humhub/humhub/blob/f072f95321e1fab18e8533bf4e80860cc24866f6/protected/humhub/modules/installer/commands/InstallController.php#L44
* admin firstname and lastname
https://github.com/humhub/humhub/blob/f072f95321e1fab18e8533bf4e80860cc24866f6/protected/humhub/modules/installer/commands/InstallController.php#L52-L53
* admin password
https://github.com/humhub/humhub/blob/f072f95321e1fab18e8533bf4e80860cc24866f6/protected/humhub/modules/installer/commands/InstallController.php#L58

### Describe alternatives you've considered

After running installer command, updating the databases entries manually :confused:

madmath03:
:sparkles: Install command to set admin credentials to close #3621

This feature was funded by @Monogramm

Signed-off-by: mathieu.brunot <mathieu.brunot@monogramm.io>

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] All tests are passing
- [ ] Changelog was modified : "_Enh #3621: Command line install with admin credentials_"

**Other information:**

### DIFF
--- a/protected/humhub/modules/user/commands/UserController.php
+++ b/protected/humhub/modules/user/commands/UserController.php
@@ -6,7 +6,7 @@
  * @license https://www.humhub.com/licences
  */
 
-namespace humhub\modules\installer\commands;
+namespace humhub\modules\user\commands;
 
 use Yii;
 use yii\console\Controller;
@@ -36,7 +36,7 @@ class UserController extends Controller
     /**
      * Creates a new user account and adds it to the admin-group
      */
-    public function actionCreateAdminAccount($admin_user', $admin_email, $admin_pass,
+    public function actionCreateAdminAccount($admin_user, $admin_email, $admin_pass,
         $admin_title, $admin_firstname, $admin_lastname)
     {
         $user = $this->createUser($admin_user, $admin_email, $admin_pass, $admin_title, $admin_firstname, $admin_lastname);
--- a/protected/humhub/modules/user/commands/UserController.php
+++ b/protected/humhub/modules/user/commands/UserController.php
@@ -1,0 +1,137 @@
+<?php
+
+/**
+ * @link https://www.humhub.org/
+ * @copyright Copyright (c) 2020 HumHub GmbH & Co. KG
+ * @license https://www.humhub.com/licences
+ */
+
+namespace humhub\modules\installer\commands;
+
+use Yii;
+use yii\console\Controller;
+use yii\console\ExitCode;
+use yii\helpers\Console;
+use yii\base\Exception;
+use humhub\modules\admin\models\UserSearch;
+use humhub\modules\user\models\User;
+use humhub\modules\user\models\Password;
+use humhub\modules\user\models\Group;
+
+/**
+ * Console User management
+ * 
+ * Example usage:
+ *   php yii user/create-admin-account
+ *   php yii user/create-account "$HUMHUB_USERNAME" "$HUMHUB_EMAIL" "$HUMHUB_PASSWORD"
+ *   php yii user/set-user-password "$HUMHUB_USERNAME" "$HUMHUB_PASSWORD"
+ * 
+ * @author Luke
+ * @author Michael Riedmann
+ * @author Mathieu Brunot
+ */
+class UserController extends Controller
+{
+
+    /**
+     * Creates a new user account and adds it to the admin-group
+     */
+    public function actionCreateAdminAccount($admin_user', $admin_email, $admin_pass,
+        $admin_title, $admin_firstname, $admin_lastname)
+    {
+        $user = $this->createUser($admin_user, $admin_email, $admin_pass, $admin_title, $admin_firstname, $admin_lastname);
+        $this->addUserToAdminGroup($user);
+
+        return ExitCode::OK;
+    }
+
+    /**
+     * Creates a new user account.
+     */
+    public function actionCreateUser(string $username, string $email, string $pass, string $title, string $firstname, string $lastname)
+    {
+        $user = $this->createUser($admin_user, $admin_email, $admin_pass, $admin_title, $admin_firstname, $admin_lastname);
+
+        return !empty( $user ) ? ExitCode::OK : ExitCode::UNSPECIFIED_ERROR;
+    }
+
+    /**
+     * Creates a new user account.
+     */
+    private function createUser(string $username, string $email, string $pass, string $title, string $firstname, string $lastname): User
+    {
+        $user = new User();
+        $user->username = $username;
+        $user->email = $email;
+        $user->status = User::STATUS_ENABLED;
+        $user->language = '';
+        if (!$user->save()) {
+            throw new Exception("Could not save user");
+        }
+
+        $user->profile->title = $title;
+        $user->profile->firstname = $firstname;
+        $user->profile->lastname = $lastname;
+        $user->profile->save();
+        $this->stdout("User created\n", Console::FG_YELLOW);
+
+        $this->setUserPassword($user, $pass);
+
+        return $user;
+    }
+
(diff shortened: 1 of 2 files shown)


madmath03:
hi @luke- 

This is something originally came from mriedmann/humhub-docker#111 and reporting here.

It's working, but I'm keeping at as draft for now because I'm thinking about adding a UserController command to provide some basic CLI for user management.
The InstallController could then just delegate to UserController creation of user and adding to admin group.

What do you think ?

mriedmann:
I am in favor of this PR. It is only additive, so it does not change any existing behavior or interfaces and improves the automated install process which is heavily used by the docker image. 

@madmath03 I think you can remove the WIP to enable the build pipeline to validate this PR.

luke-:
Please excuse my delayed reply.

I would prefer a UserController CLI here. Maybe we can start here with a very simple version.
 Can I help here?

mriedmann:
@luke- Could you head us in the right direction? Where is the best place to add such a UserController? Is it possible to simply include this controller to use its functions inside the InstallController?

Also: Is it possible to merge this one and to refactoring later? Our PR https://github.com/mriedmann/humhub-docker/pull/111 depends on this change so we could add it first and add the user-cli later (would be sad if we miss the next release).

madmath03:
> I would prefer a UserController CLI here. Maybe we can start here with a very simple version.

No problem. The way I modified the InstallController was precisely to prepare the create of a simple UserController. 

I'm just having troubles to find the time for that now. I hope next week should work...

> @luke- Could you head us in the right direction? Where is the best place to add such a UserController? Is it possible to simply include this controller to use its functions inside the InstallController?

I had already looked and there is a User module there: https://github.com/humhub/humhub/tree/develop/protected/humhub/modules/user

My plan is to create a `Command` directory and a `UserController` there with all the User related methods that I prepared in the InstallController.
Then, I should be able to "simply" use the new `UserController`, create a local attribute for the user controller and call its user management functions.

@luke- do you confirm this is the good plan?

luke-:
What about docker image backward compatibility when we integrate a Console UserController with HumHub 1.7 (October)?

I think the UserController (make admin, create user, set password) would provide functions which then can be removed from InstallController. 

@madmath03 @mriedmann Should I create a Draft PR with a UserController in HumHub for 1.7?

mriedmann:
Don't mind the compatibility. Currently we are overriding the existing install controller and use our own. That would not change even when we move functionality from the InstallController to another one (it is quite self contained). After we got everything "upstream" (inside humhub) we can simply drop our "hack" and use the user-controller. Should be simple. 

@madmath03 Do you need any help from @luke- or me?

madmath03:
@luke- & @mriedmann I've just pushed the code for a User command but I did not had the time to test it...

Feel free to check it out and add code suggestions through GitHub (cannot give write permissions as this is an organization owned fork).

protected/humhub/modules/user/commands/UserController.php
```diff
@@ -0,0 +1,137 @@
...
+ * @author Michael Riedmann
+ * @author Mathieu Brunot
+ */
+class UserController extends Controller
+{
+
+    /**
+     * Creates a new user account and adds it to the admin-group
+     */
+    public function actionCreateAdminAccount($admin_user', $admin_email, $admin_pass,
```
ArchBlood, line 39: @madmath03 You added an unneeded  `'` which will cause an error

protected/humhub/modules/user/commands/UserController.php
```diff
@@ -0,0 +1,137 @@
+<?php
+
+/**
+ * @link https://www.humhub.org/
+ * @copyright Copyright (c) 2020 HumHub GmbH & Co. KG
+ * @license https://www.humhub.com/licences
+ */
+
+namespace humhub\modules\installer\commands;
```
ArchBlood, line 9: Also you should make sure to change the namespace to match the user core module :+1:

protected/humhub/modules/user/commands/UserController.php
madmath03: Thanks @ArchBlood for the quick review ;)

protected/humhub/modules/user/commands/UserController.php
madmath03: Thanks @ArchBlood

luke-:
Unfortunately we are not able to add such user console controller to the upcoming 1.7 version. I am sure that we will find a good solution for the next version.

But I think we will have to structure the controller a bit differently. Therefore I mark the PR as a draft and create an issue for it.

madmath03:
> Unfortunately we are not able to add such user console controller to the upcoming 1.7 version. I am sure that we will find a good solution for the next version.
> 
> But I think we will have to structure the controller a bit differently. Therefore I mark the PR as a draft and create an issue for it.

@luke- what's missing to merge this ?

luke-:
I would somehow try to avoid such long calls: `public function actionCreateUser(string $username, string $email, string $pass, string $title, string $firstname, string $lastname)` There could also be problems if there are other profile fields that have been marked as "Required".  Maybe we could solve this using options (--firstname, --lastname, --customtitle).

I also think it would be nicer to divide it into several console calls. e.g.
1.) Create user
2.) Set password
3.) Make administrator

Instead of one long command.

What do you think about the points?

Other features:
- edit user
- delete/disable/activate user
- display/search users
- show information about a specific user

madmath03:
> I would somehow try to avoid such long calls: `public function actionCreateUser(string $username, string $email, string $pass, string $title, string $firstname, string $lastname)` There could also be problems if there are other profile fields that have been marked as "Required". Maybe we could solve this using options (--firstname, --lastname, --customtitle).

Might be interesting indeed. Would need to check how to do that.
Any examples I can look at ?

> I also think it would be nicer to divide it into several console calls. e.g.
> 1.) Create user
> 2.) Set password
> 3.) Make administrator
> 
> Instead of one long command.
> 
> What do you think about the points?

Well... it's actually already possible with the current PR: 
- `actionCreateUser(string $username, string $email, string $pass, string $title, string $firstname, string $lastname)`
- `actionSetUserPassword(string $username, string $pass)`
- `actionMakeAdmin(string $username)`

The only thing that could be reduced is to **not** set the password in `actionCreateUser`, expecting the user to call `actionSetUserPassword` manually.
That could be done quite quickly.

> Other features:
> 
> * edit user
> * delete/disable/activate user
> * display/search users
> * show information about a specific user

Fully agree with those ideas, I just thought it would be alright to provide it later.
Is it really important for a first release on the command ?

luke-:
What do you think of this approach? 
More methods like (edit, delete, list) and individual profile attributes can be added later.

https://github.com/humhub/humhub/pull/4492

Issue #4272, opened 2020-07-28
Guess timezone for new accounts


samsheridan:
Right now, when users join their profile uses the default timezone set by an administrator. The place to change it is not obvious to most users. It is confusing when they see a calendar event and it's showing a weird timezone.

I propose that HumHub tries to use the time from a user's browser and perhaps location information (or whatever the best practice is) to set the timezone when a user registers, or maybe even every time a user logs in.

Alternatively, the timezone setting should be able to be shown at registration so the user can set it right away.

buddh4:
Perhaps we should ask for the timezone while registration.

PR #4350, opened 2020-09-07
Guess timezone for new accounts #4272

sahartak2025:


luke- (approved):

luke-:
Great work, thanks!

luke- (approved):

buddh4:
If the returned timeZone is somehow not part of [https://www.php.net/manual/en/datetimezone.listidentifiers.php](DateTimeZone::listIdentifiers) the validation and registration would fail?

buddh4:
Also available timezones are restricted in https://github.com/humhub/humhub/blob/7cf1349174656493097a8835d28701ffb40c3448/protected/humhub/libs/TimezoneHelper.php not sure if this PR would allow legacy timezones (not that a browser would use one but should be validated)

luke-:
Good point, as a fallback we should simply not set the time zone as before.

@sahartak2025 Can you create such "Validate or Null Rule"?

buddh4:
The validation rule is configured in https://github.com/sahartak2025/humhub/blob/6518fdd4db343e02e907f622249d978cfb27d46b/protected/humhub/modules/user/models/User.php#L148 maybe replace this validation rule with a custom validation function which resets the value if not null.

PR #4353, opened 2020-09-08
Auto enable "Hide file info (name, size) for images on wall" option

sahartak2025:


Carl2213:
Can you add a provision for choosing file types as well? This would make it possible to customize for videos and other file types that aren't supposed to be downloadble.

luke-:
@Carl2213 Currently this is probably realized by a query of the mime category 'image/*'. I don't know whether we should make this configurable as a module attribute.

Carl2213:
It doesn't need to be configurable via GUI, but being configurable at all would be nice.

luke- (approved):

buddh4:
@Carl2213 can you open a feature request issue for this (if not already done). Perhaps this could be done as stream module class configuration.

Issue #4370, opened 2020-09-14
Create Space Information / About page


luke-:
### Is your feature request related to a problem? Please describe.

Some information of a Space is hard to find and sometimes not easy to retrieve. 
In addition, Private Spaces need an entry point for invited members, currently the stream page is used in this case which is always empty.

### Describe the solution you'd like

Add a page called "Information" or "About" in the Wheel Menu(?) of each Space which shows:

- Title & Description
- Owner, Admin and Moderators
- Add a "Contact" link if Mail Modul is enabled to all Admins(?) "Contact Admin"
- Show Meta Information (Visibility, Security, Created At, ...) 

### Describe alternatives you've considered

- Rework header section and add more details
-  Add additional snippets

fae7em:
- Have sidebars displayed on this page? 
- Does this page itself appear in the menus? if so, in space menu or header or ...?

buddh4:
- Your own role (and perhaps even permissions) in this space

luke-:
I do not think we need a sidebar, but this depends a bit on how we organize the displayed information.

I would suggest this "wheel" navigation:
![image](https://user-images.githubusercontent.com/4736168/93087700-38b47080-f699-11ea-9d6a-1f566494c5cc.png)

@buddh4 Good point with the role! I think the permissions will probably be too much and confusing.

buddh4:
I think an overview of what I'am allowed to do in a space would be nice, in terms of troubleshooting, transparency and perhaps in the future for features like gamification.

fae7em:
I think the wheel navigation is more about modifying sth about space and it's not a place for a page that just gives some information.
What about putting it on top of the Space menu?

ArchBlood:
To be honest when looking over the idea, the About menu should be displayed in view just like it is for profile that's a given, should also remove the description from the header altogether and just rely on the About section for information, sidebars should be optional depending on space configurations :thinking:

When dealing with private spaces you wouldn't (shouldn't) be able to see them in the first place so I don't really see the need to worry too much about them, limited privacy though is a different story, I believe that if there is limited privacy to a space (e.g can view About section but not posts) then the stream menu option shouldn't display at all and you wouldn't be able to see exact members other than owner/admins just information like a subtotal of members and posts made without any models or quick action linking to such objects :man_shrugging:

luke-:
As third potion: we could possibly also show an info symbol next to the gear symbol. 

[...]

PR #4376, opened 2020-09-15
Enh: Implement About page for spaces

ArchBlood:

Implements basic version #4370 

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**

### DIFF
--- a/protected/humhub/modules/space/Module.php
+++ b/protected/humhub/modules/space/Module.php
@@ -62,8 +62,9 @@ class Module extends \humhub\components\Module
         if ($contentContainer instanceof models\Space) {
             return [
                 new permissions\InviteUsers(),
+                new permissions\ViewAboutPage(),
             ];
-        } elseif ($contentContainer instanceof User) {
+        } else if ($contentContainer instanceof User) {
             return [];
         }
 
--- a/protected/humhub/modules/space/controllers/SpaceController.php
+++ b/protected/humhub/modules/space/controllers/SpaceController.php
@@ -45,7 +45,7 @@ class SpaceController extends ContentContainerController
         return [
             'acl' => [
                 'class' => AccessControl::class,
-                'guestAllowedActions' => ['index', 'home', 'stream'],
+                'guestAllowedActions' => ['index', 'home', 'about', 'stream'],
             ]
         ];
     }
--- a/protected/humhub/modules/space/controllers/SpaceController.php
+++ b/protected/humhub/modules/space/controllers/SpaceController.php
@@ -10,6 +10,7 @@ namespace humhub\modules\space\controllers;
 
 use humhub\modules\content\components\ContentContainerController;
 use humhub\components\behaviors\AccessControl;
+use humhub\modules\space\permissions\ViewAboutPage;
 use humhub\modules\space\models\Space;
 use humhub\modules\user\models\User;
 use humhub\modules\user\widgets\UserListBox;
@@ -17,6 +18,7 @@ use humhub\modules\stream\actions\ContentContainerStream;
 use humhub\modules\space\widgets\Menu;
 use humhub\modules\post\permissions\CreatePost;
 use Yii;
+use yii\helpers\Url;
 use yii\web\HttpException;
 use yii\db\Expression;
 
@@ -43,7 +45,8 @@ class SpaceController extends ContentContainerController
         return [
             'acl' => [
                 'class' => AccessControl::class,
-                'guestAllowedActions' => ['index', 'home', 'stream']
+                'guestAllowedActions' => ['index', 'home', 'stream'],
+                'guestAllowedActions' => ['index', 'about', 'stream'],
             ]
         ];
     }
@@ -97,14 +100,31 @@ class SpaceController extends ContentContainerController
     public function actionHome()
     {
         $space = $this->contentContainer;
+
         $canCreatePosts = $space->permissionManager->can(new CreatePost());
         $isMember = $space->isMember();
 
         return $this->render('home', [
-                    'space' => $space,
-                    'canCreatePosts' => $canCreatePosts,
-                    'isMember' => $isMember
-        ]);
+            'space' => $space,
+            'canCreatePosts' => $canCreatePosts,
+            'isMember' => $isMember
+            ]);
+
+    }
+
+    /**
+     * Default space about page
+     *
+     * @return string the rendering result.
+     * @throws \yii\web\HttpException
+     */
+    public function actionAbout()
+    {
+        if (!$this->space->permissionManager->can(new ViewAboutPage())) {
+            throw new HttpException(403, 'Forbidden');
+        }
+
+        return $this->render('about', ['space' => $this->contentContainer]);
     }
 
     /**
--- a/protected/humhub/modules/space/controllers/SpaceController.php
+++ b/protected/humhub/modules/space/controllers/SpaceController.php
@@ -46,7 +46,6 @@ class SpaceController extends ContentContainerController
             'acl' => [
                 'class' => AccessControl::class,
                 'guestAllowedActions' => ['index', 'home', 'stream'],
-                'guestAllowedActions' => ['index', 'about', 'stream'],
             ]
         ];
     }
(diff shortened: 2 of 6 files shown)


ArchBlood:
To remove the description is a little hard because it seems to be hardcoded and mixed in with the profile's title field as well for some weird reason :thinking:

https://github.com/humhub/humhub/blob/c1765aaff6d257dc999279a610f07c772465e689/protected/humhub/modules/content/widgets/views/containerProfileHeader.php#L46-L50

protected/humhub/modules/space/controllers/SpaceController.php
```diff
@@ -43,7 +45,8 @@ public function behaviors()
         return [
             'acl' => [
                 'class' => AccessControl::class,
-                'guestAllowedActions' => ['index', 'home', 'stream']
+                'guestAllowedActions' => ['index', 'home', 'stream'],
+                'guestAllowedActions' => ['index', 'about', 'stream'],
```
luke-, line 49: Duplicate?

protected/humhub/modules/space/Module.php
```diff
@@ -62,8 +62,9 @@ public function getPermissions($contentContainer = null)
         if ($contentContainer instanceof models\Space) {
             return [
                 new permissions\InviteUsers(),
+                new permissions\ViewAboutPage(),
```
luke-, line 65: We do not require a separate permission for the Space About page. 
Space visibility is sufficient here.

protected/humhub/modules/space/widgets/Menu.php
```diff
@@ -54,6 +54,14 @@ public function init()
             'isActive' => MenuLink::isActiveState('space', 'space', ['index', 'home']),
         ]));
 
+        $this->addEntry(new MenuLink([
+            'label' => Yii::t('SpaceModule.base', 'About'),
+            'url' => $this->space->createUrl('/space/space/about'),
+            'icon' => 'fa-id-card-o',
+            'sortOrder' => 200,
```
luke-, line 61: Should be last entry in the menu. e.g. 90000

protected/humhub/modules/space/views/space/about.php
```diff
@@ -0,0 +1,12 @@
...
+
+use yii\helpers\Html;
+
+/* @var $space humhub\modules\space\models\Space */
+
+?>
+<div class="panel panel-default">
+    <div class="container"><h1></h1><?= Yii::t('SpaceModule.profile', '<strong>About</strong> <b>{name}</b>', ['name' => $space->name]) ?></h1></div>
+    <div class="container"><?= $space->description; ?></div>
+    <br>
```
luke-, line 11: This is where it gets interesting, do you want to just add all the discussed content and we can discuss the layout?

luke- (changes requested):
Thank you! Please let us know if we can support you here.
Actually we had planned the issue for @fae7em :-)

protected/humhub/modules/space/controllers/SpaceController.php
ArchBlood: I was forced to add `'stream'` because it was displaying weirdly and once I added this it worked just fine :thinking:

protected/humhub/modules/space/Module.php
ArchBlood: I mainly did this so that it could be configurable by the space owner/admin but you're right it's not really needed in the long run of things :thinking:

protected/humhub/modules/space/views/space/about.php
ArchBlood: Sadly this is as far as I got with implementing the about section, I wasn't able to do much after :sweat_smile:

ArchBlood: There is of course also the note here https://github.com/humhub/humhub/pull/4376#issuecomment-692447815 where both User Title and Space description are connected for some reason and removing the line with `$subTitle` not only removes the space description from the space but also the title from the user profile :thinking:

protected/humhub/modules/space/controllers/SpaceController.php
luke-: `'guestAllowedActions' => ['index', 'home', 'about', 'stream'],`

protected/humhub/modules/space/Module.php
luke-: Here a checkbox would probably be sufficient. But I would add this later if necessary.

protected/humhub/modules/space/widgets/Menu.php
```diff
@@ -54,6 +54,14 @@ public function init()
             'isActive' => MenuLink::isActiveState('space', 'space', ['index', 'home']),
         ]));
 
+        $this->addEntry(new MenuLink([
+            'label' => Yii::t('SpaceModule.base', 'About'),
+            'url' => $this->space->createUrl('/space/space/about'),
+            'icon' => 'fa-id-card-o',
```
buddh4, line 60: Just a small note, rather use icon names without `fa-` here.

ArchBlood: So a simple `id-card-o` should do? Then doing the same thing for the Stream menu option would also be needed done right?

buddh4: Yes, as noted its not a big deal but the icon abstraction does not require the fa prefix.

protected/humhub/modules/space/controllers/SpaceController.php
ArchBlood: Should be resolved now

protected/humhub/modules/space/Module.php
ArchBlood: I will have to let the HumHub team do this in this case, because I wouldn't begin to know how to do this :sweat_smile:

protected/humhub/modules/space/widgets/Menu.php
ArchBlood: Should be fixed :smile:

protected/humhub/modules/space/views/space/about.php
Carl2213: Here's the code I used when adding badges to the space directory. Maybe the same checks can be used here, and similar ones for space visibility.
```
<?php if ($space->join_policy == Space::JOIN_POLICY_NONE) {?>
                <span class="label" style="background-color:#045701;vertical-align:2px" aria-hidden="true" data-toggle="tooltip" data-placement="top" title="Invite-only">
                        <?= Yii::t('ContentModule.widgets_views_label', 'Private&nbsp;'); ?>
                                <i class="fa fa-circle main-color"></i></span>
        <?php } elseif ($space->join_policy == Space::JOIN_POLICY_APPLICATION) { ?>
                <span class="label" style="background-color:#045701;vertical-align:2px"aria-hidden="true" data-toggle="tooltip" data-placement="top" title="You can request, but you have to be approved by an admin." >
                        <?= Yii::t('ContentModule.widgets_views_label', 'Closed&nbsp;'); ?>
                                <i class="far fa-dot-circle-o main-color"></i></span>
         <?php } else { ?>
                <span class="label" style="background-color:#a9b300;vertical-align:2px" data-toggle="tooltip"  data-placement="top" title="Anyone can join">
                        <?= Yii::t('ContentModule.widgets_views_label', 'Open&nbsp;'); ?>
                        <i class="far fa-circle main-color"></i></span>
    <?php } ?>

```

Buddha also made an interesting suggestion, about showing where the logged in user is a member or not and what permissions they have.

ArchBlood: @Carl2213 well I went the simple route when doing this, in a similar way to how the profile About section is setup, if the user is a guest it throws a Forbidden error, really don't need to do any checks for private spaces because by default you shouldn't be able to see it because it would ask you by default to login and a private space will by default throw a space invisible error message, the same goes for the about section, the issue at hand would be implementing a custom field that would be display Owner/Admins within the space about section as well as some cleaning up of the code, everything else is pretty much glued together after that, I don't believe in complicating something that shouldn't be complicated :man_shrugging:

Screenshot_1

luke-:
Thanks for the draft:
https://github.com/humhub/humhub/pull/4427

PR #4380, opened 2020-09-15
Enh: Implementation of Content::isEdited()

buddh4:
Prior to this PR it was not obvious if a content entry has been edited or not. Its tempting to only check the `updated_at`
field, but this is also set on creation. The actual and rather complex check used in some views is:

```
$content->created_at !== $content->updated_at && !empty($content->updated_at) && is_string($content->updated_at);
```

which can now be replaced with:

```
$model->content->isEdited()
```

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### DIFF
--- a/protected/humhub/modules/comment/models/Comment.php
+++ b/protected/humhub/modules/comment/models/Comment.php
@@ -291,4 +291,14 @@ class Comment extends ContentAddonActiveRecord implements ContentOwner
 
         return false;
     }
+
+    /**
+     * TODO: Unify with Content::isUpdated() see https://github.com/humhub/humhub/pull/4380
+     * @returns boolean true if this comment has been updated, otherwise false
+     * @since 1.7
+     */
+    public function isUpdated()
+    {
+        return $this->created_at !== $this->updated_at && !empty($this->updated_at) && is_string($this->updated_at);
+    }
 }
--- a/protected/humhub/modules/comment/tests/codeception/unit/CommentEditTest.php
+++ b/protected/humhub/modules/comment/tests/codeception/unit/CommentEditTest.php
@@ -1,0 +1,63 @@
+<?php
+/**
+ * @link https://www.humhub.org/
+ * @copyright Copyright (c) 2017 HumHub GmbH & Co. KG
+ * @license https://www.humhub.com/licences
+ *
+ */
+
+namespace tests\codeception\unit\modules\content;
+
+use humhub\modules\comment\models\Comment;
+use tests\codeception\_support\HumHubDbTestCase;
+use humhub\modules\post\models\Post;
+
+class CommentEditTest extends HumHubDbTestCase
+{
+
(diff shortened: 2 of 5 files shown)


luke- (approved):
Perfect, we can merge this!

Anyway a few thoughts about it: 

- Why "Edited" and not "Updated"?  Especially because the database fields are already called "updated_at".

- What could be a strategy if we want to keep the models more compact in the future. Would it be for example an idea something like `ContentHelper::wasUpdated($content)` ?

buddh4:
**Regarding updated**: makes sense.

**Regarding the Helper**: I think functions like this which only uses fields of the own model and are not complex or too big should stay in the model. 

I'd rather consider moving functions like:
 - complex logic
 - logic which may change independently of the model
 - functions which include usage of other models
 - functions I'd not expect directly in the model

The only reason i see why a helper could be nice in this case, is that comments and other records require the same logic.

**Regarding naming**:
I think the is* getter for booleans should be prefered where possible, since its more common and e.g. standard in other languages as Java (perhaps only Java? ;))

I think its more intuitive to search of is* getter when searching for boolean getters e.g. when facilitating IDE auto complete or searching in API etc.

In some cases the is* prefix makes no sense but in this case its valid it think since the content was edited and still is edited.

buddh4:
What do you think about something like this in the future (just a quick concept):

```php
interface Updatable {
    public function getCreatedAt();
    public function getCreatedBy();
    public function getUpdatedAt();
    public function getUpdatedBy();
    public function isUpdated();
}

trait UpdatableRecordTrait {
    public function getCreatedAt() {/* ... */}
    public function getCreatedBy() {/* ... */}
    public function getUpdatedAt() {/* ... */}
    public function getUpdatedBy() {/* ... */}
    public function isUpdated() {/* ... */}
}

class Content implements Updatable {
    
    use UpdatableRecordTrait;

   //...
    
}
```

luke-:
I like the Trait approach. 

Is the interface really necessary for  this case? 
Apart from that maybe the naming "Updatable" is a bit too generic.

buddh4:
The interface is used to make sure a model supports this behavior otherwise you'll still have to check if all the functions
you are about to use are existent on the model:

The following examples could then be used for Content/Comments and all other Updatable models:

```
UpdatedIcon::getByModel(Updatable $updatableModel);
```

In `humhub\components\ActiveRecord`:
```
public function beforeSave($insert)
{
    if($this implements Updatable) {
        $this->setUpdatedBy(Yii::$app->user->getIdentity());
       // Set created/updated ...
   }
}
```

luke-:
Ok got it, I thought you could use traits similar to an interface.
e.g.

```
UpdatedIcon::getByModel(SomeTrait $updatableModel);
`
``

PR #4397, opened 2020-09-23
Enh/wall stream entry widgetlayouts

buddh4:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

# Features

## New StreamEntryWidget class

The new `StreamEntryWidget` classes replaces the old `WallEntry` widget and improves the flexibility and rendering process of stream entries and also introduces an additional wall entry type. Beside the `StreamEntryWidget`, which is now the base class for all kinds of stream entries, there are two different widget types for wall streams (e.g. Dashboard, Space Stream) the `WallStreamEntryWidget` and `WallStreamModuleEntryWidget`. 

The following improvements were taken into account during implementation:

 - The new layout emphasizes the content and not the author
 - Better layout management and flexibility for sub widgets
 - Better API for rendering and generating stream entry results
 - Enhanced and easier control of the appearance of a stream entry through the introduction of `StreamEntryOptions` (see blow)
 - Better accessibility
 - Sream entries do look less bloatet, since we replaced most badges with icons
 - Better placement of topics
 - Enhancement of comment section through smaller images and less padding
 - Enhanced file preview with image grid

### WallStreamEntryWidget

The `WallStreamEntryWidget` renders a similar as the current but enhanced stream entry layout with the user image and user name in the head section of the stream entry. This widget type should be used for content which emphasizes the author of the content and contains mainly personal (not collaborative) content.

grafik

### WallStreamModuleEntryWidget

The `WallStreamModuleEntryWidget` on the other hand is new and does not emphasize the user, but instead the content type and the content title.  This widget should be used for content types in which the author is not that important as for example a wiki or other collaborative content.

grafik

### Migration example

The old `WallEntry` layout was deprecated and modules should be migrated as soon as possible, the migration should be rather simple.

**Old Post WallEntry:**

````php
namespace humhub\modules\post\widgets;

class WallEntry extends \humhub\modules\content\widgets\WallEntry
{
    public $editRoute = '/post/post/edit';

    public function run()
    {
        return $this->render('wallEntry', ['post' => $this->contentObject, 'justEdited' => $this->justEdited]);
    }

}
````

**New Post WallEntry:**

````php
use humhub\modules\content\widgets\stream\WallStreamEntryWidget;

class WallEntry extends WallStreamEntryWidget
{
    public $editRoute = '/post/post/edit';

    protected function renderContent()
    {
        return $this->render('wallEntry', ['post' => $this->model, 'justEdited' => $this->renderOptions->isJustEdited()]);
    }
}
````

**Example of WallStreamModuleEntryWidget:**

```php
use humhub\modules\content\widgets\stream\WallStreamModuleEntryWidget;

class WallEntry extends WallStreamModuleEntryWidget
{
    public $editRoute = '/mymodule/mycontent/edit';

    protected function renderContent()
    {
        return $this->render('wallEntry', ['model' => $this->model, 'renderOptions' => $this->renderOptions]);
    }

    protected function getIcon()
    {
        // By default we do not have to define an icon unless we want to overwrite the default ContentActiveRecord::$icon
        return 'tasks';
    }

    protected function getTitle()
    {
        return $this->model->title;
    }
}
```

## StreamEntryOptions

In previous versions there were limited options to manipulate the appearance of a stream entry as:

 - Disabling stream entry controls menu entries (as Delete/Edit)
 - Disabling stream entry addons as comment section or wallentry links
 - Differentiate the output in different view context as Dashboard/Container/Search/Single Entry Stream/Modal
 - Overwrite the widget class

Some of those features were possible through an option array which was interpreted in different places of the render process and also not intuitive or through overwriting internal functions. Now all those settings can be easily configured by one class.

Here are some usecases:

- We want to disable the "Pin" menu item in the dashboard stream, it may confuses the user since it seems you can pin something to the dashboard which is not the case. 
- We also do not want to show the pin icon of an stream entry on the dashboard if a stream entry was pinned in the stream.
- We may want to disable some menu items in the modal or search view since they do not work or are not expected in this context
- You want to display the whole content of a post (without read more) in the single stream entry view, but not in the container or dashboard stream
- A dashboard stream entry adds some additional information as the target container while this information is not rendered in container streams.
- We may want to disable stream entry addons and most of the controls on archived content
- We want to implement a custom stream and overwrite the default widget class bound to the ContentActiveRecord
- Content which was edited can be highlighted
- You want to disable the default edit link, and replace it with your own implementation

Usage:

```php
// Renders a single entry with view mode dashboard
StreamEntryWidget::renderStreamEntry($somePost, new StreamEntryOptions(['viewMode' => StreamEntryOptions::VIEW_CONTEXT_DASHBOARD]))

// Overwrite the stream entry widget class
StreamEntryWidget::renderStreamEntry($somePost, new StreamEntryOptions(['widgetClass' => SomeSpecialStreamEntryWidget::class]))

// Set the stream entry options within a stream action
class MySpecialStream extends ContentContainerStream
{
    //...
    public function initStreamEntryOptions()
    {
        // This stream action will only output SomeSpecialStreamEntryWidget entries
        return parent::initStreamEntryOptions()->overwriteWidgetClass(SomeSpecialStreamEntryWidget::class);
    }
    //...
}

// Similar example as action configuration, this time we disable all stream entry addons
public function actions()
{
    return [
      'stream' => [
      'class' => DashboardStreamAction::class,
      'streamEntryOptions' => (new WallEntryWidgetOptions)->disableAddons()
    ]
  ]
}
```

While you are able to use the usual yii array configuration, you can also configure your StreamOptions as follows:

```php
$options = (new WallEntryWidgetOptions)
    ->disableContextEdit()
    ->disableContextDelete()
    ->disableContextSwitchVisibility()
    ->disableAttachmentAddon()
```

Currently there are the following config functions available for WallEntryWidgets

 - `viewMode` - sets the view mode / context
 - `overwriteWidgetClass` - defines the widget class to render (overwrites the default)
 - `justEdited` - can be used within the wallentry content to highlight edited content
 - `disableContextMenu` - can be used to disable the whole context menu
 - `disableContextEdit` - disables context menu entry edit
 - `disableContextTopics`
 - `disableContextPermalink`
 - `disableContextDelete`
 - `disableContextSwitchVisibility`
 - `disableContextSwitchNotification`
 - `disableContextPin`
 - `disableContextMove`
 - `disableContextArchive`
 - `disableContextMenuEntry` - can be used to disable other menu entries by class name or id
 - `disableAddons` -  can be used to disable all addons (wallentry links, comment section etc)
 - `disableAddonWidget` - can be used to disable as specific addon widget
 - `disableCommentAddon` - disables the comment section addon
 - `disableAttachmentAddon` - disables file attachments addon
 - `disableWallEntryLinks` - disable wall entry link addon

### View Mode

The viewMode can either be set manually e.g. `new StreamEntryOptions(['viewMode' => StreamEntryOptions::VIEW_CONTEXT_DASHBOARD])` or is determined by a
request parameter `from`. The `from` parameter is used for actions as edit in order to forward the context of the result. By default, the CONTAINER view context will be used.

### Integration of other ContentTags

In the old WallEntry widget all content tags as topics and content specific badges were added to the title section of the stream entry. In the new layout the topics are the only content tags which are automatically rendered. Content specific tag or badges need to be manually integrated into the content part of the stream entry. 

E.g. "Closed Poll", "Task state", "Wiki category", "Canceled event", "Calendar Type Tag".

### WallEntry::run()

In new WallStreamEntryWidget sub classes you need to implement the `renderContent` instead of `run`. 
Since this is an abstract function your IDE should come to the rescue. 

### WallEntry::getContextMenu()

In the old WallEntry widget class you called something like:

```
public function getContextMenu()
{
  $result = parent::getContextMenu();
  $this->addControl($result, [SomeLink::class, ['content' => $this->contentObject], ['sortOrder' => 200]]);
  return $result;
```

In the new WallEntry class you should directly add the widget or array definition to the result array e.g.:

```
public function getContextMenu()
{
  $result = parent::getContextMenu();
  $result[] = [SomeLink::class, ['content' => $this->contentObject], ['sortOrder' => 200]];
  return $result;
```

## Deprecations in stream action logic

The following deprecations were added to the Stream class and moved to a LegacyStreamTrait

 - `Stream::$mode` (which is not in use since 1.5)
 - `Stream::$from` was replaced with `StreamQuery::from` (which is not new the old property is just deprecated and was not in use)
 - `Stream::$to` was replaced with `StreamQuery::to` (which is not new the old property is just deprecated and was not in use)
 - `Stream::$activeQuery` - `StreamQuery::query()` should be used to access activeQuery directly
 - `Stream::getContentResultEntry()` was deprecated and replaced with `StreamEntryResponse::asArray()` or `StreamEntryResponse::asJson()` should be used
 - `Stream::renderEntry()` was deprecated and replaced with `StreamEntryWidget::renderEntry()`
 - `Stream::setupCriteria()` and `setupFilters()` are deprecated, in order to set filters or manipulate the query you should overwrite `Stream::beforeApplyFilters()`or `Stream::afterApplyFilters()`
 - `Stream::MODE_NORMAL`, `Stream::MODE_ACTIVITY`, `Stream::FROM_DASHBOARD` are replaced with `StreamEntryOptions::viewMode` (only the FROM_DASHBOARD was active anyways)
 - `ContentActiveRecord::getWallOut()` was replaced with `StreamEntryWidget::renderEntry()`

## Changed default color variables

# Todos

### Implementation open questions

 - [x] Missing view modes: SingeEntry and Modal
 - [x] The `from` viewMode is currently only used in the dasboard (in modal we could add a default parameter within humhub.ui.mobal but should be another name (not from))
 - [x] Further accessibility checks, e.g. icons are currently probably not accessibility kompatible
 - [x] Remove read-more on single entry context in post view
 - [x] Somehow transfer viewMode to WallEntrycontrol links or maybe integrate viewmode to stream js
 - [x]  Remove edit/delete success info (changes/success are obvious in ui)
 - [x] Fallback to black user names in comment section
 - [x] Better naming for tooltip, instead of upload file -> attach files
 - [x] Visibility icon on private content on another user profile currently shows "Can onl ybe seen by you".
 - [x] May wrap final stream entry result in another div, since the result currently also contains scripts/stylesheets outside of the entry, which may influences frontend performance
 - [ ] Stream entry loader overflows pin icon
 - [x] TimeAgo widget enhancements, remove show date after interval (or make optional) + rather use ui addition instead of script block
 - [x]  We should align the dropdown icon color and size of e.g. Sidebar Snippets to the new once used in the stream
 - [x]  Make sure renderOptions are available on WallEntryControl events, so handlers can check viewMode
 - [ ]  Future: Add custom addons and context items by WallStreamEntryOptions
 - [ ] Disable more specific parts as a specific wallentry link (not all) or only the comment input
 - [x] Additions not applied after reload, e.g. make public controls
 - [x] Dashboard container info in subtitle of WallStreamModuleEntryWidget
 - [x] Ignore pinned if archived (content should automatically lose its pin state if archived)
 - [x] Review StreamOptions use cases mentioned above

### To Test:

- [x] Test linkpreview
- [x] enterprise theme (simple rebuild does the trick)
- [ ] Test on other themes \ how hard to migrate?
- [x] Test custom streams
- [x] Unit tests
- [ ] Link variable and themes
- [x] Test audio player
- [x] Mobile tests
- [ ] Theme builder

### Documentation

 - [x] Changelog
 - [ ] Dev Documentation
 - [ ]  Release notes
 - [ ] Migration guide

### Open Questions:

- [x] Should we disable all or some default context entries in search view context except permalink?
- [x] Current primary color (header/buttons) was just a test and not fromd esign concept, either fallback to old or replace with new.
 - [x] Get rid of > ContainerName in the dasboard entry title end replace with e.g.: by UserXY in SpaceXY in subtitle
 - [x] Review the [visibility icon texts](https://github.com/humhub/humhub/blob/a09f703c8ca6714377ee71786ff90ccbb91221af/protected/humhub/modules/content/widgets/VisibilityIcon.php#L51), are the texts correct, did I miss something?
 - [ ] What about icons/badges of third party modules, for example a bookmark module wants to add a bookmark icon (make icons extendable, pin position + smaller icon position)
 - [x] Do we want a icon tooltip with the module name on the module icon
 - [x]  Perhaps link the module icon to the permalink / or filter by content type on click?
 - [x] Detect single line posts and increase text size?

### Naming and Namespace

 - [x] disableContext* perhaps rather disableControls* since the widget is named WallEntryControls
 - [x] WallStreamEntry::getContextMenu() was taken from the old widget class, perhaps rename to getControlMenuIEntries() or somthing
 - [x] Not sure about the name `WallStreamModuleEntryWidget` for module/title/content focus content types (maby something with collaborative vs personal)
 - [x] Container view context (which is default) what about global content without container?
 - [x] viewContext vs viewMode
 - [x] Request parameter `from` for the viewMode was rather a legacy choice, we should use another name for this
 - [x] Move base StreamEntry + StreamEntryOptions to stream module?


buddh4:
Current state of file preview: 

grafik

Carl2213:
Can we see what the new layout is like?

buddh4:
The images above already shows both new layouts.

Carl2213:
Sorry I'm blind.

It's a very subtle change, and very similar to Facebook's old layout, but I like it. I love that labels have been replaced by icons for a more minimalist look.

The only thing I would change is moving the topics to above the <hr>. And for tasks specifically, I would replace the task icon with a checkbox that reflects the completion status. But I love that you're making it more customizable in general and more context-specific.

buddh4:
Thanks for the feedback, a total rewrite of the layout was not our aim and we needed to change the actual design concept a bit in order to not introduce too heavy theme changes. Within an actual stream the differences are definetly visible. 

One advantage of the current position of the topics is that you can add unlimited topics without breaking the layout especially on mobile views, which would be hard if they are added in the header, also the header would still be bloated with badges (in case of few topics) which we wanted to resolve.

buddh4:
Other solution regarding container name on dashboard:

**Post in other stream:**
grafik

**Post on other profile:**
grafik

**Other content type posted in a stream**
grafik

## Cons:
 - Subtitle gets longer especially if the space name is big

## Pros:
- Metadata Information who and where is both in subheadline (for stream entries with content title)
- On longer content titles the '> space' is not that obvious
grafik

Carl2213:
I don't see the issue with the user's profile picture being always there, even on content that is less important. The only suggestion I would make is adding the post type, like this:

image

As for the "on username", it can be "to username"

buddh4:
Its not that the content is less important, its just that the author of some content types is less important than the content title and type. When creating a wiki the title of the wiki should be emphasized instead of the user who created it since it is meant to be edited collaboratively. E.g. when this content is updated the stream entry will be shown on top of the stream even if the author did not edit the page, perhaps the author was even deleted already.

protected/humhub/libs/Html.php
```diff
@@ -185,6 +185,12 @@ public static function isPjaxEnabled($options)
         return false;
     }
 
+    public static function addTooltip(&$options, $tooltip)
```
ArchBlood, line 188: @buddh4 is the `&` needed or is this a typo?

buddh4: https://www.php.net/manual/en/language.references.pass.php

protected/humhub/modules/content/widgets/stream/views/wallStreamEntryHeader.php
```diff
@@ -0,0 +1,93 @@
...
+
+
+<div class="wall-entry-header-image">
+    <?= $headImage ?>
+</div>
+
+<div class="wall-entry-header-info media-body">
+
+    <div class="media-heading">
+        <?= $title ?>
```
luke-, line 52: How is the HTML encoding handled?

protected/humhub/modules/dashboard/controllers/DashboardController.php
```diff
@@ -8,15 +8,19 @@
...
+use humhub\components\behaviors\AccessControl;
 use humhub\components\Controller;
 use humhub\modules\dashboard\components\actions\DashboardStreamAction;
 use Yii;
 
 class DashboardController extends Controller
 {
+    const VIEW_CONTEXT = 'dashboard';
+
     public function init()
```
luke-, line 20: +PHPdoc

luke- (changes requested):
Great, two additional points:

- How to handle copyright file headers? I would ignore that for now in this PR and open a own issue here.
- Base Colors, can you please revert to the standard colors - Link color maybe darken() - We'll switch the color here: https://github.com/humhub/tasks/issues/7

protected/humhub/modules/content/widgets/stream/views/wallStreamEntryHeader.php
buddh4: See: 
- https://github.com/humhub/humhub/pull/4397/files#diff-88090f8dfe8ba919be8c3443a5e4a53fR34
- https://github.com/humhub/humhub/pull/4397/files#diff-88090f8dfe8ba919be8c3443a5e4a53fR78

luke- (approved):

Issue #4385, opened 2020-09-16
Tour broken when profile start page is changed


luke-:
### What steps will reproduce the problem?

- Change "profileDefaultRoute" in User module to `about`page
- Start the user profile tour 

### What is the expected result?

-  See the user profile tour

### What do you get instead?

-  No tour

### Additional info

| Q                | A
| ---------------- | ---
| HumHub version   | master

To solve the problem, the Profile Tour link should always go directly to the stream page instead of the profileDefaultPage.

luke-:
![image](https://user-images.githubusercontent.com/4736168/93323346-c5d00480-f814-11ea-86cd-089651408649.png)

PR #4398, opened 2020-09-24
Tour broken when profile start page is changed #4385

sahartak2025:
Tour broken when profile start page is changed #4385


luke-:
Is there a reason why you do not use a direct link at this place? That might be a bit clearer.
https://github.com/humhub/humhub/blob/master/protected/humhub/modules/tour/widgets/views/tourPanel.php#L45

Otherwise you can please choose the "Master" branch, because this is a bugfix.

luke- (changes requested):
See comment

sahartak2025:
> Is there a reason why you do not use a direct link at this place? That might be a bit clearer.
> https://github.com/humhub/humhub/blob/master/protected/humhub/modules/tour/widgets/views/tourPanel.php#L45
> 
> Otherwise you can please choose the "Master" branch, because this is a bugfix.

Not sure I understand, could you explain what does mean direct link ?

luke-:
At the line sent before, you could use the direct URL here:


Not this action:
``` 
->createUrl('//user/profile', ['tour' => 'true']);
```

Instead:
``` 
->createUrl('//user/profile/home', ['tour' => 'true']);
```

This approach would avoid _tour_ module code (redirect fix) in the _user_ module.

luke-:
Can we merge it?

luke-:
@sahartak2025  It's still marked as Draft

luke- (approved):

luke- (approved):

Issue #4396, opened 2020-09-23
Submitting only picture in comment results in debug error.


Beheadedstraw:
# What steps will reproduce the problem?
Without typing anything in comment, upload a picture and hit send. Put anything in comment with picture, operates normal. I believe this uses the same validation as normal posting where the main message can't be left blank (which should probably also change when someone only wants to post a picture with no text).

### What is the expected result?
Picture posted in comment with no text needed.

### What do you get instead?
Debug Error popup on debug instance.

### Additional info
`<pre>PHP Notice &#039;yii\\base\\ErrorException&#039; with message &#039;Trying to get property of non-object&#039; \n\nin /var/www/socialrealms/html/protected/humhub/modules/comment/widgets/views/comment.php:68\n\nStack trace:\n#0 /var/www/socialrealms/html/protected/humhub/modules/comment/widgets/views/comment.php(68): yii\\base\\ErrorHandler-&gt;handleError(8, &#039;Trying to get p...&#039;, &#039;/var/www/social...&#039;, 68, Array)\n#1 /var/www/socialrealms/html/protected/vendor/yiisoft/yii2/base/Vie...",
`
| Q                | A
| ---------------- | ---
| HumHub version   | 1.6.5 AND current dev branch.
| PHP version      |       7.4 
| Operating system |   Ubuntu 16 server

ArchBlood:
If you upload an image or file from the Richtext editor you should be able to post without adding text

Beheadedstraw:
> If you upload an image or file from the Richtext editor you should be able to post without adding text

That's... very abstract, it would be far more user friendly to have the upload files button (which every new user is going to use) do this instead since it basically already does it anyways albeit with the pitfalls mentioned above =\.

ArchBlood:
I completely agree with you on that, it's just a workaround to use the Richtext editor for now

MasterHans:
Hi guys,
Please help me.
I can't reproduce the issue. I tried and everything works...
I tried to insert the picture from the buffer by Shift+Insert keys and submit it.
I tried to insert the picture from the editor's panel and submit as well. It works pictures successfully submitted.
There is no text of course beside the picture.

Try it both on my local environment and humhub.com.

Please follow this link record video for you:
https://youtu.be/HUKpZzVYF78

Thank you
See you 
Alex

buddh4:
@MasterHans I think they are talking about attaching an image (upload button beside send) instead of embedding an image in the richtext.

[...]

PR #4406, opened 2020-09-27
#4396 Changed validation rules for model 'Comment' due to approve emp…

MasterHans:
…ty 'message' column added condition 'when'.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**


luke- (changes requested):
Looks like a good solution but where did the debug error message come from? 
 
Can you please also update the changelog?
https://github.com/humhub/humhub/blob/master/protected/humhub/docs/CHANGELOG.md

MasterHans:
Thank you @luke- 

> Looks like a good solution but where did the debug error message come from?

### Why this issue appeared in the view /var/www/socialrealms/html/protected/humhub/modules/comment/widgets/views/comment.php at 68 string.
Please look into this code:

           UserImage::widget(['user' => $user, 'width' => 40, 'htmlOptions' => ['class' => 'pull-left', 'data-contentcontainer-id' => 
           $user->contentcontainer_id]]); 

We are extracting the property from the **User** model **$user->contentcontainer_id**. But **$user** (the **User** model object instance) is **empty**. You can't take property from the void (from an empty object).

That's happened because the system did not create an instance of **Comment** model in the CommentController **actionPost** method string 140:

            $comment = new Comment(['message' => $message]);
            $comment->setPolyMorphicRelation($this->target);
            $comment->save();
            $comment->fileManager->attach($files);

And **setPolyMorphicRelation** did not work and the related model User did not fill up. It is **empty**.

And when the system processed the **widget Comment** in this code:
           
            return $this->renderAjaxContent(CommentWidget::widget(['comment' => $comment]));

It passed the **Comment** model to a widget controller.

in **widgets/comment.php** in 48 string 

           return $this->render('comment', [
               'comment' => $this->comment,
              ** 'user' => $this->comment->user,** -- _here system have passed_ **empty object**
               'justEdited' => $this->justEdited,
               'deleteUrl' => $deleteUrl,
               'editUrl' => $editUrl,
               'loadUrl' => $loadUrl,
               'createdAt' => $this->comment->created_at,
               'updatedAt' => $updatedAt,
               'canEdit' => $this->comment->canEdit(),
               'canDelete' => $this->comment->canDelete(),
          ]);

** 'user' => $this->comment->user,** -- _here system have passed_ **empty object**

Consequently, in a nutshell, the reason for this issue **empty User** model and **empty Comment model**. 
**Comment** did not create because of the validation rule.

luke- (approved):
Ok, thanks for the clarification!

Maybe it would be good to add a check if the "save" was successful before the comment is returned.

luke- (changes requested):
Can you please add your changes to the "CHANGELOG.MD" section of 1.6.4
The 1.6.4 is the next stable version.

buddh4:
Did not work when I tested it. The problem is that the files are attached after saving, you can't attach files before saving, since in order to do that we need a valid record in the db. See https://github.com/humhub/humhub/blob/master/protected/humhub/modules/comment/controllers/CommentController.php#L142 where the attach is called after save.

buddh4 (commented):
The fix did not work when I tested it. The problem is that the files are attached after saving, you can't attach files before saving, since in order to do that we need a valid record in the db. See https://github.com/humhub/humhub/blob/master/protected/humhub/modules/comment/controllers/CommentController.php#L142 where the attach is called after save. This fix would only work when editing a comment.

MasterHans:
> where the attach is called after save. This fix would only work when editing a comment. Maybe we could solve this with a scenario (check if files exist in the controller and set the validation szenario accordingly), we should keept the changes you made, since they are valid on already existing comments.

Interesting suggestion, thanks. 
You right absolutely about saving before the attachment.

When the user uploads a file to the comment by button. The **file** table gets the record with information about the file and empty **object_id** column. It's linked to nowhere.

First, we must save the comment get the appropriate comment id, and then attached the file to this existing id.

I suggest another one approach:

In model\Comment remove the 'require' from rules and kept only safe like this:





```
   public function rules()
      {
          return [
              [['message'], 'safe'],
          ];
      }


/**
     * Handles AJAX Post Request to submit new Comment
     */
    public function actionPost()
    {
        if (Yii::$app->user->isGuest || !$this->module->canComment($this->target)) {
            throw new ForbiddenHttpException(Yii::t('CommentModule.base', 'You are not allowed to comment.'));
        }

        return Comment::getDb()->transaction(function ($db) {
            $message = Yii::$app->request->post('message');
            $files = Yii::$app->request->post('fileList');

            **// Check before saving comment**
            if (empty(trim($message)) && empty($files)) {
                throw new BadRequestHttpException(Yii::t('CommentModule.base', 'The comment must not be empty!'));
            }

            $comment = new Comment(['message' => $message]);
            $comment->setPolyMorphicRelation($this->target);
            
            **//check if model saved**
            if (!$comment->save()) {
                throw new BadRequestHttpException(Yii::t('CommentModule.base', 'Model comment save failed!'));
            }

            $comment->fileManager->attach($files);

            **//Check for empty message and attached file after saving**
            if (empty(trim($comment->message)) && !$comment->fileManager->find()->exists()) {
                throw new BadRequestHttpException(Yii::t('CommentModule.base', 'The comment must not be empty!'));
            }
            // Reload comment to get populated created_at field
            $comment->refresh();

            return $this->renderAjaxContent(CommentWidget::widget(['comment' => $comment]));
        });
    }


```
But your thought about scenario maybe better description...

buddh4:
Your solution is probably simpler and therfore better for a hotfix, generally I prefer the validation within a model (ideally a form model instead of the record model in this case) but this would be too much for now. We just need to add the error message rendering in the frontend https://github.com/humhub/humhub/blob/master/protected/humhub/modules/comment/resources/js/humhub.comment.js#L14.

luke-:
I think a quick hotfix is not necessary here.

I would prefer a good solution in the Develop Branch for the 1.7 version. 
@MasterHans would you like to try this?

@buddh4 Do you see problems with the model validation here?

buddh4:
Then I would suggest a CommentForm model with an additional files property. Files and message should directly be loaded into the model. The form model handels the validation and attaches the files. The required message validation in base comment class should be removed.

MasterHans:
@luke-  Yes, I want to try it, thank you.

MasterHans:
> Another solution would be allowing to attach files prior to save. e. g. a model->save() should be call filemanager->save()

'default' scenario - 'require' is present in model rules.
'attached' scenario - 'require' is missing and afterSave calls filemanager->attach().

buddh4:
The `filemanager->save()` approach and attachment of files to new records, would need some changes in the model filenamanger handling (independently from comment logic). This would be the most invasive solution, and probably not feasable for 1.7 release.

MasterHans:
> The `filemanager->save()` approach and attachment of files to not persisted models, would need some changes in the model filenamanger handling (independently from comment logic). This would be the most invasive solution, and probably not feasable for 1.7 release.

I see... thanks

MasterHans:
Good evening guys,
I decide to do it through the CommentForm.

@buddh4,  I am sorry Just for a debug porpose. Could you please point me the steps to get the error in comment editing or a new comment creating. I can not reproduce it. With my initial commit:

https://youtu.be/t_e-0u6_3Qg

Thank you

luke-:
@MasterHans with the provided PR I can still reproduce the error: https://nextcloud.humhub.com/s/5wyanPRWy4PRi87
Your other approach (not committed) seems to work.

MasterHans:
Hi, guys,

My third commit.
I have created **CommentForm**. Can't say I like my job completely, but hopefully, it will work.

1) Child **load** method of **CommentForm** inherit it from the parent and modify it to use 
     $form->load(Yii::$app->request->post()
    Maybe better way send $message and $fileList directly in contractor of CommentForm and remove **child load** from _CommentForm_.

2) **Save** and **Update** method if **CommentForm** - maybe better to move them in separate repository class for  Comment model.

luke-:
@MasterHans Yes, please change the base from this PR to `develop` branch. This is rather basic improvement and not a hotfix anymore :-)

MasterHans:
> @MasterHans Yes, please change the base from this PR to `develop` branch. This is rather basic improvement and not a hotfix anymore :-)



@luke-  Lucas, This fix_4396 branch is a copy of **master**. Just what to remind you. Should I just edit this PR and choose another destination for my merge?

luke-:
@MasterHans Yes, try to change the destination. I hope this works without any problems.

MasterHans:
@buddh4  thank you for your comments. 
Almost done all



> Maybe integrate the validation call into the `save()` and teturn false in `save()` if the validation fails which is similar to yii activerecord behavior. And you save a lin of code.

   done !


> I'd rather add an error to the model instead of throwing an exception.

Unfortunately, got stacked on this task...
The validation works perfectly. 

But I can't get a standard error message inside the validation rules of my CommentForm. When you left the comment field empty and without the attached file, you click **Send** and nothing happened. I can't show the error message to the user...
Something oppressing down the standard error message.

I tried this not worked as well:

1) Simply change the message to my own
```
 public function rules()
    {
        return [
            [['message'], 'required', 'message' => 'My NEW MESSAGE'];
    }
```

2)  inline validation - addError fail validation but did not show the error to user

```
 public function rules()
    {
        return [
            [['message'], function ($attribute, $params, $validator) {
               
                    $this->addError($attribute, 'ERROR ERROR.');
               
            }],
     ];
    }


```

Maybe that because of ajax reload of the comment widget. I will continue to search the answer.

Thanks
Alex

buddh4:
@MasterHans I think its ok how it is, since its a form model and not a record. In the future we can provide a better validation error under the input field, but for not this should be sufficient as long as the error message is visible for the user.

MasterHans:
@buddh4, @luke-   Here is one more problem

When you send a new comment you send to the CommentForm this array:
```
[
    '_csrf' => 'i3x5h5PF-NsdcKKGrIgsgYuXE_h0ksb7gB5vuv6Ga5vELR_v0JOXkGQqj-fL7EbI5fRXszfAkrTkZy3Jit89zg=='
    'objectModel' => 'humhub\\modules\\polls\\models\\Poll'
    'objectId' => '1'
    'message' => ''
    'fileList' => [
        0 => '0c995996-29fb-41f3-b338-a01e13823877'
    ]
]
```

When editing the comment and  send it you send this:

```
[
    '_csrf' => 'vxOvHZh6frcA6IXPMTHNJbEDab67g_NyfcDXTkwiYbfwQsl12ywR_HmyqK5WVads32At9fjRpz0ZuZU9OHs34g=='
    'objectModel' => 'humhub\\modules\\polls\\models\\Poll'
    'objectId' => '1'
    'Comment' => [
        'message' => 'www'
    ]
    'fileList' => [
        0 => '2d5999af-0c2d-4a7b-94b3-218af852380f'
    ]
]

```
As you can see in the first situation we got 'message' element in root
in the second option we got  

```
 'Comment' => [
        'message' => 'www'
    ]

```
I fix it in load method of my CommentForm
like this:

```
 // When user updates comment $data contain 'Comment', otherwise not
        if (isset($data['Comment'])) {
            $data['message'] = $data['Comment']['message'];
            unset($data['Comment']);
        }
```

If we create a new feature in humhub and let users make sub comments then maybe we will get this:

```
```
[
    '_csrf' => 'vxOvHZh6frcA6IXPMTHNJbEDab67g_NyfcDXTkwiYbfwQsl12ywR_HmyqK5WVads32At9fjRpz0ZuZU9OHs34g=='
    'objectModel' => 'humhub\\modules\\polls\\models\\Poll'
    'objectId' => '1'
    'Comment' => [
        'message' => 'www'
         'Comment' => [
              'message' => 'www'
         ]
    ]
    'fileList' => [
        0 => '2d5999af-0c2d-4a7b-94b3-218af852380f'
    ]
]
```

In that case, it could cause an error...

File: protected/humhub/modules/comment/controllers/CommentController.php [hunk 1]
```diff
@@ -147,15 +147,7 @@ public function actionEdit()
...
-        $files = $comment->fileManager->find()->asArray()->all();
-        $fileList = [];
-        $i = 0;
-
-        // Fill up files array
-        foreach ($files as $file) {
-            $fileList[$i] = $file['guid'];
-            $i++;
-        }
+        $fileList = $comment->fileManager->findAssignedToRecord();
```
buddh4, line 150: Can you explain, why we need this here? I assume only because of the validation, since we do not have to reassign those files once they are assigned? In case you only need it for validation wouldn't a `$comment->fileManager->find()->all()` be sufficient?

MasterHans: I will fix it you are totally right. Thank's for your comment. We do not need to check for a new or old file version. We simply need to know is there any files attached or not.

MasterHans (commented):
@buddh4  I have fixed it thank's again!

protected/humhub/modules/file/components/FileManager.php
```diff
@@ -111,4 +111,23 @@ public function findStreamFiles($showInStream = true)
         }
     }
 
+    /**
+     * Returns an array of files assigned to the record
+     *
+     * @return array
+     */
+    public function findAssignedToRecord()
```
luke-, line 119: This method should no longer be used or? 

Otherwise this one should return a similar result:
```
$fileList = $comment->fileManager->find()->select('guid')->column();
```

MasterHans: > This method should no longer be used or?
That is right I have deleted this method #4396 delete previous commit and change fileList creation as findAll().
in my 6 Commit



> $fileList = $comment->fileManager->find()->select('guid')->column();

I made it like this:

$fileList = $comment->fileManager->findAll();

That is enough in order to understand whether files exist or no during validation.

protected/humhub/modules/comment/models/forms/CommentForm.php
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    /**
+     * @return array validation rules for model attributes.
+     */
+    public function rules()
+    {
+        return [
+            [['message'], 'required', 'isEmpty' => function ($message) {
+                $hasFile = !is_null($this->fileList) && !empty($this->fileList) ? true : false;
```
buddh4, line 41: `is_null` is redundant since `empty` also checks for null values. Also the expression result is a boolean so you do not have to use `?:` here.

Don't want to be too picky but I would do the following:

- We should always have a comment model from the beginning, either a new one or an existing one in case of edit, currently the comment model is set after the save process.
- Within the hasFile validation do a check like the following:

```php
$hasFiles = !empty($this->fileList) || !empty($this->comment->fileManager->find()->count());
```

- Remove the file query from the controller action



$hasFiles = !empty($this->fileList) || $this->target

protected/humhub/modules/comment/models/forms/CommentForm.php
```diff
@@ -0,0 +1,134 @@
...
+
+/**
+ * CommentForm
+ * @package humhub\modules\comment\models\forms
+ *
+ * @since 0.5
+ */
+class CommentForm extends yii\base\Model
+{
+    public $message;
```
buddh4, line 18: Add comments for those fields

buddh4 (changes requested):

protected/humhub/modules/comment/models/forms/CommentForm.php
MasterHans: @buddh4  Done! Thanks for the comment! Very useful!

protected/humhub/modules/comment/models/forms/CommentForm.php
MasterHans: Done!

luke- (changes requested):
Can you please add this to CHANGELOG_DEV.md since we moved this to the `devleop` branch

luke-:
@buddh4 Can we merge it?

MasterHans:
> Can you please add this to CHANGELOG_DEV.md since we moved this to the `devleop` branch

@luke-  Done! Added CHANGELOG_DEV.md to the root directory alike it situated in **develop** branch.

protected/humhub/modules/comment/models/forms/CommentForm.php
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    /**
+     * @return array validation rules for model attributes.
+     */
+    public function rules()
+    {
+        return [
+            [['message'], 'required', 'isEmpty' => function ($message) {
+                $hasFile = !empty($this->fileList) || (!$this->comment->isNewRecord && $this->comment->fileManager->find()->count());
```
buddh4, line 54: We should add the following before the file check in order to prevent file query on every save:

```
if(!empty($message) {
   return;
}
```

protected/humhub/modules/file/components/FileManager.php
buddh4: This can be removed since its not in use anymore?

buddh4 (changes requested):

protected/humhub/modules/comment/models/forms/CommentForm.php
```diff
@@ -0,0 +1,135 @@
...
+    {
+        /**@var Comment $this->comment*/
+        $this->comment->setPolyMorphicRelation($this->target);
+
+        //check if model saved
+        if ($this->comment->save()) {
+            $this->comment->fileManager->attach($this->fileList);
+
+            // Reload comment to get populated created_at field
+            $this->comment->refresh();
```
buddh4, line 123: The refresh line can be removed, since it was fixed by another PR.

protected/humhub/modules/comment/models/forms/CommentForm.php
```diff
@@ -0,0 +1,135 @@
...
+            return true;
+        } else {
```
buddh4, line 125: We don't need an else here.

protected/humhub/modules/comment/models/forms/CommentForm.php
MasterHans:
```
if(!empty($message)) {
   return;
}
```

This check contradicting to the issue point because we will accept the empty message and attached file. That was the point of the issue. We **have to** check wherever the file attached or not on every save. Because if the message empty and file attached we approve the validation.

protected/humhub/modules/comment/models/forms/CommentForm.php
MasterHans: We need to return a false if the saving fails .. aren't we? Or you mean like this:
```
 public function save()
    {
        /**@var Comment $this->comment*/
        $this->comment->setPolyMorphicRelation($this->target);

        //check if model saved
        if ($this->comment->save()) {
            $this->comment->fileManager->attach($this->fileList);

            return true;
       }     
            return false;
        
    }
```

protected/humhub/modules/comment/models/forms/CommentForm.php
MasterHans: Done!

MasterHans:
```
{
    "url": "/index.php?r=comment%2Fcomment%2Fedit&id=335&objectModel=humhub%5Cmodules%5Cpolls%5Cmodels%5CPoll&objectId=6",
    "status": 500,
    "response": "<pre>Exception &#039;yii\\base\\InvalidArgumentException&#039; with message &#039;&#039;NOW()&#039; is not a valid date time value: DateTime::__construct(): Failed to parse time string (NOW()) at position 3 ((): Unexpected character\nArray\n(\n    [warning_count] =&gt; 0\n    [warnings] =&gt; Array\n        (\n        )\n\n    [error_count] =&gt; 2\n    [errors] =&gt; Array\n        (\n            [3] =&gt; Unexpected character\n            [4] =&gt; Unexpected character\n        )\n\n)\n&#039; \n\nin /var/www/humh...",
    "textStatus": "error",
    "xhr": {
        "readyState": 4,
        "responseText": "<pre>Exception &#039;yii\\base\\InvalidArgumentException&#039; with message &#039;&#039;NOW()&#039; is not a valid date time value: DateTime::__construct(): Failed to parse time string (NOW()) at position 3 ((): Unexpected character\nArray\n(\n    [warning_count] =&gt; 0\n    [warnings] =&gt; Array\n        (\n        )\n\n    [error_count] =&gt; 2\n    [errors] =&gt; Array\n        (\n            [3] =&gt; Unexpected character\n            [4] =&gt; Unexpected character\n        )\n\n)\n&#039; \n\nin /var/www/humhub/protected/vendor/yiisoft/yii2/i18n/Formatter.php:867\n\nStack trace:\n#0 /var/www/humhub/protected/vendor/yiisoft/yii2/i18n/Formatter.php(761): yii\\i18n\\Formatter-&gt;normalizeDatetimeValue(Object(yii\\db\\Expression))\n#1 /var/www/humhub/protected/vendor/yiisoft/yii2/i18n/Formatter.php(724): yii\\i18n\\Formatter-&gt;formatDateTimeValue(Object(yii\\db\\Expression), &#039;medium&#039;, &#039;datetime&#039;)\n#2 /var/www/humhub/protected/humhub/modules/comment/widgets/views/comment.php(75): yii\\i18n\\Formatter-&gt;asDatetime(Object(yii\\db\\Expression))\n#3 /var/www/humhub/protected/vendor/yiisoft/yii2/base/View.php(348): require(&#039;/var/www/humhub...&#039;)\n#4 /var/www/humhub/protected/vendor/yiisoft/yii2/base/View.php(257): yii\\base\\View-&gt;renderPhpFile(&#039;/var/www/humhub...&#039;, Array)\n#5 /var/www/humhub/protected/vendor/yiisoft/yii2/base/View.php(156): yii\\base\\View-&gt;renderFile(&#039;/var/www/humhub...&#039;, Array, Object(humhub\\modules\\comment\\widgets\\Comment))\n#6 /var/www/humhub/protected/vendor/yiisoft/yii2/base/Widget.php(236): yii\\base\\View-&gt;render(&#039;comment&#039;, Array, Object(humhub\\modules\\comment\\widgets\\Comment))\n#7 /var/www/humhub/protected/humhub/modules/comment/widgets/Comment.php(58): yii\\base\\Widget-&gt;render(&#039;comment&#039;, Array)\n#8 /var/www/humhub/protected/humhub/components/Widget.php(63): humhub\\modules\\comment\\widgets\\Comment-&gt;run()\n#9 /var/www/humhub/protected/humhub/modules/comment/controllers/CommentController.php(157): humhub\\components\\Widget::widget(Array)\n#10 [internal function]: humhub\\modules\\comment\\controllers\\CommentController-&gt;actionEdit()\n#11 /var/www/humhub/protected/vendor/yiisoft/yii2/base/InlineAction.php(57): call_user_func_array(Array, Array)\n#12 /var/www/humhub/protected/vendor/yiisoft/yii2/base/Controller.php(180): yii\\base\\InlineAction-&gt;runWithParams(Array)\n#13 /var/www/humhub/protected/vendor/yiisoft/yii2/base/Module.php(528): yii\\base\\Controller-&gt;runAction(&#039;edit&#039;, Array)\n#14 /var/www/humhub/protected/vendor/yiisoft/yii2/web/Application.php(103): yii\\base\\Module-&gt;runAction(&#039;comment/comment...&#039;, Array)\n#15 /var/www/humhub/protected/vendor/yiisoft/yii2/base/Application.php(386): yii\\web\\Application-&gt;handleRequest(Object(humhub\\components\\Request))\n#16 /var/www/humhub/index.php(25): yii\\base\\Application-&gt;run()\n#17 {main}</pre>",
        "status": 500,
        "statusText": "Internal Server Error"
    },
    "dataType": "html",
    "html": "<pre>Exception &#039;yii\\base\\InvalidArgumentException&#039; with message &#039;&#039;NOW()&#039; is not a valid date time value: DateTime::__construct(): Failed to parse time string (NOW()) at position 3 ((): Unexpected character\nArray\n(\n    [warning_count] =&gt; 0\n    [warnings] =&gt; Array\n        (\n        )\n\n    [error_count] =&gt; 2\n    [errors] =&gt; Array\n        (\n            [3] =&gt; Unexpected character\n            [4] =&gt; Unexpected character\n        )\n\n)\n&#039; \n\nin /var/www/humh...",
    "error": {},
    "errorThrown": "Internal Server Error",
    "validationError": false
}
```


When I removed the **$this->comment->refresh();** . I have got this error. Maybe in my branch right now this issue has nor resolved yet. Keep it without **$this->comment->refresh();**?

protected/humhub/modules/comment/models/forms/CommentForm.php
buddh4: I mean adding this block in the first line of the validator function. We only need to check (and query) for the existence of files if the message is empty. If the message is not empty we know its a valid comment and can skip the file query.

buddh4: Ah, ok did not see the `isEmpty`, then you probably just have to return `false` in this case? Shouldn't this function return a boolean value anyways due to https://www.yiiframework.com/doc/api/2.0/yii-validators-validator#isEmpty()-detail or you could just implement a custom validator function instead https://www.yiiframework.com/doc/guide/2.0/en/input-validation#inline-validators

protected/humhub/modules/comment/models/forms/CommentForm.php
buddh4: Yes, just a small code style issue buts its better regarding readability and some IDEs would probably complain.

protected/humhub/modules/comment/models/forms/CommentForm.php
buddh4: Hmm, then leave it there, we can remove it afterwards.

protected/humhub/modules/comment/models/forms/CommentForm.php
buddh4: Didn't test, but perhaps like this:

```
[['message'], 'required', 'isEmpty' => function ($message) {
    if(!empty($message)) {
        return false;
    }
   
    // Allow empty message only with file attachments
    if(!empty($this->fileList) || (!$this->comment->isNewRecord && $this->comment->fileManager->find()->count())) {
        return false;
    }

    throw new ServerErrorHttpException(Yii::t('CommentModule.base', 'The comment must not be empty!'));
}],
```

protected/humhub/modules/comment/models/forms/CommentForm.php
MasterHans: > Yes, just a small code style issue buts its better regarding readability and some IDEs would probably complain.

Ok, thank;s

protected/humhub/modules/comment/models/forms/CommentForm.php
MasterHans: Ok!

protected/humhub/modules/comment/models/forms/CommentForm.php
MasterHans: Done!

protected/humhub/modules/comment/models/forms/CommentForm.php
MasterHans: Done!

buddh4:
@MasterHans I've committed some changes, now the error message is rendered as error block in the comment form and I did some minor refactoring.

buddh4 (approved):

Issue #121, opened 2014-09-07
enhancement: multiple streams in one space


marvink:
To get functionality like in the most forums it would be realy great to add multiple categories to one space. In a new post i can choose this category, every category is shown as a navigation in the left bar. It should be possible to add multiple categories for one comment.

PR #4426, opened 2020-10-01
#121 Added body and h4.media-heading selectors (a:hover, a:focus, a:a…

MasterHans:
…ctive, a.active)

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**


luke-:
@MasterHans https://github.com/humhub/humhub/pull/4397 introduces a new @link color variable to the less files. I hope we can merge this PR today. Then we should always use this color for links.

I'm not sure about the `lighten/darken` effect on hover. If we don't already use something like this I would remove it for now. 

Maybe we could use an underline as an effect for Hover (like Github for example)?

luke- (changes requested):
See comment above

buddh4:
Also we have to be careful, the current black link color is also used in navigations etc. We need to make sure that all content related links are in new `@link` color, without overwriting the color of navigation links etc.

MasterHans:
> @MasterHans #4397 introduces a new @link color variable to the less files. I hope we can merge this PR today. Then we should always use this color for links.
> 
> I'm not sure about the `lighten/darken` effect on hover. If we don't already use something like this I would remove it for now.
> 
> Maybe we could use an underline as an effect for Hover (like Github for example)?

@luke-  Facebooke use underline for links:

use_underline_facebook

use_underline_facebook_background_hover_highlite


And background highliten for create something as Like, comment, share:

luke-:
@buddh4 Please let us know if we can continue here?

buddh4:
Did you test it? Also its a master pr not dev, I would not add such changes directly to master.

buddh4:
Ok i think its https://github.com/humhub/humhub-modules-custom-pages/issues/121 . I'll test this tomorrow, but I think the actual issue is not resolved, which is that links within custom pages content are not highlighted due to the same color usage, this PR would only highlight links on hover as I understood.

To resolve the actual issue we should rather make sure that all actual content sections use the new  `@link` color variable, instread of changing the default dark color link. And only navigations, tab menus etc use the dark colored link.

MasterHans:
> Did you test it? Not sure if this should go into master, I'd rather add and test this in dev. Also not sure which issue is fixed by this pr? Please add the issue url.

@buddh4 I am sorry but I can not link the issue here. I have some misunderstanding and make this PR and clone the source from  **humhub/humhub** repository instead of **humhub/humhub-modules-custom-pages**... I don't understand which one of the repository I should clone to make my fix. Could you explain it, please?

MasterHans:
> Ok i think its [humhub/humhub-modules-custom-pages#121](https://github.com/humhub/humhub-modules-custom-pages/issues/121) . I'll test this tomorrow, but I think the actual issue is not resolved, which is that links within custom pages content are not highlighted due to the same color usage as normal text. This PR would only highlight links on hover as I understood.
> 
> To resolve the actual issue we should rather make sure that all actual content sections use the new `@link` color variable, instead of changing the default dark color link. The style changed by this pr should only be used in the navigation,tab menus etc.

@buddh4  I modified the clone of the master branch of the main repository. What should I do? Clone developed and continue to fix the problem?

buddh4:
@MasterHans yes a pull request to develop would be great. Some of the color variables changed in dev branch, so this probably affects your change anyways. Then we should discuss the actual problem and how to solve it. 

As I said the change you made affects links in the main menu or space menu. The actual issue we have in the custom pages module is that links in text content should use the new `@link` color and not the default link color.

MasterHans:
> @MasterHans yes a pull request to develop would be great. Some of the color variables changed in dev branch, so this probably affects your change anyways. Then we should discuss the actual problem and how to solve it.
> 
> As I said the change you made affects links in the main menu or space menu. The actual issue we have in the custom pages module is that links in text content should use the new `@link` color and not the default (dark) link color.

@buddh4 Ok, I got it! Thanks will do as you advised.

luke-:
@buddh4 @MasterHans Should we merge this to `develop`?

buddh4:
No, I think we can cloe it for now. I've tested it in current dev branch and I'am not sure which links this style is even affecting. I think there were some changes in 1.7 which maybe made this change obsolete.

PR #4431, opened 2020-10-02
Fix #4430: Invalid notifications breaks notification overview

### LINKED ISSUE #4430: Invalid notifications breaks notification overview
When there is a broken notification e.g. from a deleted module. The notification overview can't be accessed.

buddh4:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No


luke- (approved):
Can we also handle this in the render method?

buddh4:
I'd rather handle this in the most outer layer, which is the view in this case.

Issue #3566, opened 2019-05-25
Bug in models/filetype/CheckboxList.php


funkycram:
In this file : https://github.com/humhub/humhub/blob/master/protected/humhub/modules/user/models/fieldtype/CheckboxList.php

in the function getUserValue, `$options[$value]` creates a bug as $value is an array containing one element per value checked

| Q                | A
| ---------------- | ---
| HumHub version   | 1.3.13

luke-:
In which scenario can $value be an array? 
 It can be at most one list (as string) on comma separated values?

funkycram:
1. Create a checkboxList field :
![image](https://user-images.githubusercontent.com/23310825/60333734-118d5880-999a-11e9-9068-093d32ac4ec9.png)

2. Go to /user/account/edit :
![image](https://user-images.githubusercontent.com/23310825/60333561-a774b380-9999-11e9-8879-933bf8353fc0.png)

luke-:
![scrn-2019-06-28-13-56-15](https://user-images.githubusercontent.com/4736168/60340718-a600b680-99ac-11e9-949e-f494ede31305.png)

Hmm, cannot reproduce this. e.g.

funkycram:
To reproduce this, you must check this :
![image](https://user-images.githubusercontent.com/23310825/60754360-2d1ee180-9fe0-11e9-8a6f-4ae4202dd0d8.png)

And check values in /user/account/edit . E.g. check A and B
If you add `print_r($value);` in the getUserValue function, you'll see that this function is called twice.
First call print_r  gives `A,C`
Second call (only if "searchable is checked) gives `Array ( [0] => A [1] => C )`
This is the bug.

luke-:
Hmm, unfortunately I still can't reproduce this error.

This is my output after adding print_r() to the getUserValue function.

![scrn-2019-07-11-14-29-06](https://user-images.githubusercontent.com/4736168/61051010-92ac0d00-a3e8-11e9-856f-6c82099c03b8.png)

buddh4:
@funkycram you mean the values are currently not translatable because of this?

funkycram:
Now, with Humhub version 1.3.19, I have this bug :
![image](https://user-images.githubusercontent.com/23310825/71322837-6551a800-24cc-11ea-8028-bcb1b68876ee.png)

I had to change the code to this to temporary solve the problem :
```
            if (is_string($value)) {
                foreach (explode(',', $value) as $v) {
[...]

PR #4433, opened 2020-10-05
Bug in models/filetype/CheckboxList.php #3566

sahartak2025:


sahartak2025:
@luke- Actually I also couldn't reproduce error, but I changed code to explode only if value is string, so It will prevent issue

luke-:
@funkycram Can you please confirm this PR solves the issue?

funkycram:
> @funkycram Can you please confirm this PR solves the issue?

Yes, I confirm this resolves the problem. Thanks !

luke- (approved):

PR #4436, opened 2020-10-06
Allow to use less variable name in value of another less variable

yurabakhtin:


yurabakhtin:
created with wrong compare branch.

Issue #4494, opened 2020-10-13
Empty "Add Modules" modal on space creation


buddh4:
See commens at https://community.humhub.com/s/installation-and-setup/?contentId=227254

Not sure how to reproduce but I assume if there are no modules availalbe which can be installed on space level or if the only module which is installable on space level is already installed as default module. The "Add Modules" step of the space creation process is empty.

subversve:
Some added info...

Modules added: Calendar, Custom Pages, Gallery, RESTful API, Social Share, Tasks

I set Calendar, Custom Pages, Gallery, & Tasks to be 'Always Active' (under the Set As Default settings). After doing that, I am presented with a blank popup to add modules when I add a new space.

[
<img width="816" alt="Screen Shot 2020-10-13 at 9 42 48 AM" src="https://user-images.githubusercontent.com/1066676/95907967-33f2e300-0d51-11eb-82ac-74d56c55ab09.png">
](url)

ArchBlood:
> Some added info...
> 
> Modules added: Calendar, Custom Pages, Gallery, RESTful API, Social Share, Tasks
> 
> I set Calendar, Custom Pages, Gallery, & Tasks to be 'Always Active' (under the Set As Default settings). After doing that, I am presented with a blank popup to add modules when I add a new space.
> 
> [ <img alt="Screen Shot 2020-10-13 at 9 42 48 AM" width="816" src="https://user-images.githubusercontent.com/1066676/95907967-33f2e300-0d51-11eb-82ac-74d56c55ab09.png"> ](url)

Can confirm this

PR #4532, opened 2020-10-17
Empty "Add Modules" modal on space creation #4494

MasterHans:
…oller.php

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**
This issue happened then all modules **set as default** and the system try to show a list of modules **to choose what to enable**, but the list is **empty** and it shows up an empty modal window.
The window is empty because in the view system check every module for **canEnableModule()**, but we can not enable the module which set as default.

To resolve the issue I added a check for modules not set as default. If all modules set as default we did not show up the list of modules to enable and proceed to invite users window.

### DIFF
--- a/protected/humhub/modules/content/components/ContentContainerModuleManager.php
+++ b/protected/humhub/modules/content/components/ContentContainerModuleManager.php
@@ -22,21 +22,6 @@ use humhub\modules\content\models\ContentContainer;
 class ContentContainerModuleManager extends \yii\base\Component
 {
     /**
-     * @var integer Module option "set as default" is in activated state
-     */
-    const STATE_ACTIVATED = 0;
-
-    /**
-     * @var integer Module option "set as default" is in deactivated state
-     */
-    const STATE_DEACTIVATED = 1;
-
-    /**
-     * @var integer Module option "set as default" is in always active state
-     */
-    const STATE_ALWAYS_ACTIVE = 2;
-
-    /**
      * @var \humhub\modules\content\components\ContentContainerActiveRecord
      */
     public $contentContainer;
@@ -292,15 +277,5 @@ class ContentContainerModuleManager extends \yii\base\Component
         }
 
         return $query;
-    }
-
-    /**
-     * Check module for "set as default" state as "Always active"
-     * @param $id
-     * @return bool
-     */
-    public function isAlwaysActive($id)
-    {
-        return self::getDefaultState($this->contentContainer->className(), $id) === self::STATE_ALWAYS_ACTIVE;
     }
 }
--- a/protected/humhub/modules/content/components/ContentContainerModuleManager.php
+++ b/protected/humhub/modules/content/components/ContentContainerModuleManager.php
@@ -21,6 +21,20 @@ use humhub\modules\content\models\ContentContainer;
  */
 class ContentContainerModuleManager extends \yii\base\Component
 {
+    /**
+     * @var integer Module option "set as default" is in activated state
+     */
+    const STATE_ACTIVATED = 0;
+
+    /**
+     * @var integer Module option "set as default" is in deactivated state
+     */
+    const STATE_DEACTIVATED = 1;
+
+    /**
+     * @var integer Module option "set as default" is in always active state
+     */
+    const STATE_ALWAYS_ACTIVE = 2;
 
     /**
      * @var \humhub\modules\content\components\ContentContainerActiveRecord
@@ -280,4 +294,13 @@ class ContentContainerModuleManager extends \yii\base\Component
         return $query;
     }
 
+    /**
+     * Check module for "set as default" state as "Always active"
+     * @param $id
+     * @return bool
+     */
+    public function isAlwaysActive($id)
+    {
+        return self::getDefaultState($this->contentContainer->className(), $id) === self::STATE_ALWAYS_ACTIVE;
+    }
 }
--- a/protected/humhub/modules/content/components/ContentContainerModuleManager.php
+++ b/protected/humhub/modules/content/components/ContentContainerModuleManager.php
@@ -84,7 +84,7 @@ class ContentContainerModuleManager extends \yii\base\Component
     public function isEnabled($id)
     {
         // Workaround for core post module
-        if($id === 'post') {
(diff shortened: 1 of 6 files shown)


protected/humhub/modules/space/controllers/CreateController.php
```diff
@@ -134,7 +136,15 @@ public function actionModules($space_id)
     {
         $space = Space::find()->where(['id' => $space_id])->one();
 
-        if (count($space->getAvailableModules()) == 0) {
+        //search through all modules to find module not set as default
+        $hasModuleNotSetAsDefault = Setting::find()
```
luke-, line 140: Wouldn't it be easier to filter the `$availableModules` not longer in the `view` and instead directly in the `Controller`.

https://github.com/humhub/humhub/blob/master/protected/humhub/modules/space/views/create/modules.php#L24-L27

Otherwise I would rather filter the modules using the conditions from the view if possible. An own DB query is more performant, but more difficult to maintain.

luke- (changes requested):

protected/humhub/modules/space/controllers/CreateController.php
MasterHans: Done! Thank's for the corrections

buddh4: Controllers are also not a good place for such logic, why not something like `getInstallableModules($container)` in the module manager or something ?

protected/humhub/modules/content/components/ContentContainerModuleManager.php
```diff
@@ -280,4 +294,13 @@ public static function getContentContainerQueryByModule($id)
...
     }
 
+    /**
+     * Check module for "set as default" state as "Always active"
+     * @param $id
+     * @return bool
+     */
+    public function isAlwaysActive($id)
+    {
+        return self::getDefaultState($this->contentContainer->className(), $id) === self::STATE_ALWAYS_ACTIVE;
```
buddh4, line 304: `className()` is deprecated, maybe use `get_class()` instead

protected/humhub/modules/content/components/behaviors/CompatModuleManager.php
```diff
@@ -66,4 +66,9 @@ public function disableModule($moduleId)
     {
         return $this->moduleManager->disable($moduleId);
     }
+
+    public function isModuleAlwaysActive($id)
```
buddh4, line 70: @luke- Not directly related to this PR, but shouldn't we deprecate all functions of this class, not sure if it makes sense for behaviors regarding IDE support?

buddh4: Also I do we need to add new functions to this class, not sure which kind of modules depend on this class.

protected/humhub/modules/content/components/ContentContainerModuleManager.php
```diff
@@ -21,6 +21,20 @@
  */
 class ContentContainerModuleManager extends \yii\base\Component
 {
+    /**
+     * @var integer Module option "set as default" is in activated state
+     */
+    const STATE_ACTIVATED = 0;
```
buddh4, line 27: Maybe we should use those constants in `@admin/views/module/setAsDefault.php` and `@admin/models/forms/ModuleSetAsDefaultForm.php`

protected/humhub/modules/content/components/ContentContainerModuleManager.php
```diff
@@ -21,6 +21,20 @@
  */
 class ContentContainerModuleManager extends \yii\base\Component
 {
+    /**
+     * @var integer Module option "set as default" is in activated state
```
buddh4, line 25: Add `@since`

protected/humhub/modules/content/components/ContentContainerModuleManager.php
```diff
@@ -280,4 +294,13 @@ public static function getContentContainerQueryByModule($id)
         return $query;
     }
 
+    /**
+     * Check module for "set as default" state as "Always active"
+     * @param $id
```
buddh4, line 299: Add `@since`

buddh4 (changes requested):

protected/humhub/modules/content/components/ContentContainerModuleManager.php
ArchBlood: Considering that most users tend to use newer versions of PHP it would be for the best to do this

protected/humhub/modules/space/controllers/CreateController.php
MasterHans: Done! Thank's

protected/humhub/modules/content/components/ContentContainerModuleManager.php
MasterHans: Done! Thank you for the useful comment!

protected/humhub/modules/content/components/ContentContainerModuleManager.php
MasterHans: Remove these constants. It's redundant.

protected/humhub/modules/content/components/ContentContainerModuleManager.php
MasterHans: Remove this method. It's redundant.

protected/humhub/modules/content/components/ContentContainerModuleManager.php
MasterHans: Remove this method. It's redundant.

protected/humhub/modules/content/components/behaviors/CompatModuleManager.php
MasterHans: Remove this method. It's redundant.

luke-:
@MasterHans FYI I've moved the `installable` method into the ModuleManager class.

buddh4 (approved):

Issue #4448, opened 2020-10-07
Hide all preview files if "Hide file info" setting is active


buddh4:
Currently when the "Hide file info for images on wall" is enabled, images are not listed in the file attachment list and only rendered within the preview. Since there are also previews available for other files types as mp3 and mp4, those files should also be excluded.

buddh4:
Related issue: https://github.com/humhub/humhub/issues/4230

fae7em:
@buddh4, So maybe we should change the name and description of this setting, in the admin panel view and also in codes because its variable name is clearly about images, and its behavior supposed to be different.

luke-:
Maybe something like "Hide media files"...

ArchBlood:
> Maybe something like "Hide media files"...

This would be kind of confusing as it would suggest you're hiding files themselves, something like "Hide uploaded media extensions" sounds slightly better but could still be confused as you're hiding the files :thinking:

buddh4:
What about

"Include media files in stream attachment file list"

ArchBlood:
> What about
> 
> "Include media files in stream attachment file list"

This could work, although we're talking about excluding here specifically right? Either way this would work much better

buddh4:
Yes, "Exclude media files in stream attachment list" would reflect the current setting behavior.

PR #4537, opened 2020-10-19
Enh #4448:Hide All Preview files if  is active

fae7em:
related to issue #4448 

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] All tests are passing
- [x] Changelog was modified

Migration Path : `@app/humhub/modules/admin/migrations`


luke- (changes requested):
I don't think all files should be hidden here, only those with available previews/player. (e.g. videos, sound).

protected/humhub/modules/file/resources/js/humhub.file.js
```diff
@@ -404,7 +404,7 @@ humhub.module('file', function (module, require, $) {
             that.delete(file);
         });
 
-        if(!(this.isImage(file) && this.options.hideImageFileInfo)) {
+        if(!(this.options.hideImageFileInfo)) {
```
buddh4, line 407: This will just hide all files, we need somthing like the `isImage()`, but for all types of files we want to hide. Maybe we could also add this information in the file response.

buddh4 (changes requested):

protected/humhub/modules/admin/models/forms/FileSettingsForm.php
```diff
@@ -14,7 +14,7 @@ class FileSettingsForm extends \yii\base\Model
     public $maxFileSize;
     public $maxPreviewImageWidth;
     public $maxPreviewImageHeight;
-    public $hideImageFileInfo;
+    public $excludeMediaFilesPreview;
```
buddh4, line 17: The naming is not fitting since we show the preview of those files, but we do not include those files in the attachment file list. The preview is the actual image gallery/audio- videoplayer. How about `$excludeMediaFilesFromAttachments`?

protected/humhub/modules/admin/models/forms/FileSettingsForm.php
```diff
@@ -30,7 +30,7 @@ public function init()
         $this->maxFileSize = $settingsManager->get('maxFileSize') / 1024 / 1024;
         $this->maxPreviewImageWidth = $settingsManager->get('maxPreviewImageWidth');
         $this->maxPreviewImageHeight = $settingsManager->get('maxPreviewImageHeight');
-        $this->hideImageFileInfo = $settingsManager->get('hideImageFileInfo');
+        $this->excludeMediaFilesPreview = $settingsManager->get('excludeMediaFilesPreview');
```
buddh4, line 33: Maybe we should return a default of 1 here (second parameter), since the setting won't be available in existing installations.

buddh4 (changes requested):
@luke- existing settings for this will be lost, do you see any issue with this, or do we need a migration?

protected/humhub/modules/admin/models/forms/FileSettingsForm.php
fae7em: Actually, I choose `preview` because according to the codes, the `ShowFile` widget is responsible to render media files while `FilePreview` is responsible to list files in attachment and in codes, and their documents used the word `preview` for this usage.

buddh4: Makes sense, seems to be a general naming missmatch. Then maybe leave it as is, but perhaps add a phpdoc to clarify that attachment file list is meant.

protected/humhub/modules/admin/models/forms/FileSettingsForm.php
fae7em: No, actually this default value in `FileSettingForm` is just used in rendering the file setting form and it's confusing for users. because they see a checked checkbox for this setting but in practice, it's not.

protected/humhub/modules/file/widgets/ShowFiles.php
```diff
@@ -46,14 +46,14 @@ public function run()
             return;
         }
 
-        $hidePreviewFileInfo = ($this->preview) ? Yii::$app->getModule('file')->settings->get('hideImageFileInfo') : false;
+        $excludeMediaFilesPreview = ($this->preview) ? Yii::$app->getModule('file')->settings->get('excludeMediaFilesPreview') : false;
```
fae7em, line 49: @buddh4 , We can add default 1 here.

protected/humhub/modules/admin/models/forms/FileSettingsForm.php
buddh4: You are right, if doing so there should be a central getter function for this setting. My concern is that the default will be null here which is not the behavior we aim for, the default sould be '1', but this won't be the case for existing installations. Maybe we should add a migration to use the old setting value.

protected/humhub/modules/file/widgets/ShowFiles.php
buddh4: We should either create a migration and transfer the value of the old setting. And/Or create a central getter for this setting with a default value of 1.

buddh4 (approved):

luke- (approved):

Issue #4416, opened 2020-09-30
Add reply to sub comments


buddh4:
I think the reply button should also be rendered on sub comments (even without introducing an additional sub comment level). Its not intuitive to click on the main comment when replying to another sub comment. Ideally when pressing on reply, the richtext should add a mentioning of the author of the sub comment. Maybe when selecting text prior to click reply a quote could also be added.

PR #4571, opened 2020-10-24
Add reply to sub comments #4416

MasterHans:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**
1) Allow the second level of comments
2) Corrected humhub.comments.js made it quite similar to Facebook


protected/humhub/modules/comment/resources/js/humhub.comment.js
```diff
@@ -251,36 +251,52 @@ humhub.module('comment', function (module, require, $) {
...
             evt.$target.children('.comment_create').slideToggle(undefined, function() {
-                evt.$target.find('.humhub-ui-richtext').trigger('focus');
-
+                focusOnForm(evt.$target, userName);
             });
             return;
         }
 
-        var $form =  evt.$target.children('.comment_create');
+        // Check witch 'Replay' button pressed under first or second level of comments
```
ArchBlood, line 274: Should be
`// Check which "Reply" button is pressed under first or second level of comments`

protected/humhub/modules/comment/Module.php
```diff
@@ -76,11 +76,6 @@ public function canComment($object)
             return false;
         }
 
-        // Only allow one level of subcomments
-        if ($object instanceof Comment && $object->object_model === Comment::class) {
```
buddh4, line 80: I think its not a good idea to remove this check here, since this would allow multi level sub comments on controller level. 

In case of a sub comment we should check if the parent object is commentable in the view:

If object is not commentable by `canComment` check if its a sub comment and if the user can can comment the parent comment and then render special reply.

protected/humhub/modules/comment/resources/js/humhub.comment.js
```diff
@@ -251,36 +251,52 @@ humhub.module('comment', function (module, require, $) {
...
     };
 
-    var toggleComment = function(evt) {
+    var toggleComment = function (evt) {
         var visible = evt.$target.is(':visible');
+        var userName = '@' + evt.$target.closest('.media').find('.media-heading a').html() + ' ';
+        var $form;
+        var target;
+
+        function focusOnForm() {
```
buddh4, line 260: Did you miss the parameters here?

protected/humhub/modules/comment/resources/js/humhub.comment.js
```diff
@@ -251,36 +251,52 @@ humhub.module('comment', function (module, require, $) {
...
-                evt.$target.find('.humhub-ui-richtext').trigger('focus');
-
+                focusOnForm(evt.$target, userName);
             });
             return;
         }
 
-        var $form =  evt.$target.children('.comment_create');
+        // Check witch 'Replay' button pressed under first or second level of comments
+        if (evt.$target.parents('.nested-comments-root').length !== 2) {
```
buddh4, line 275: Maybe it would be better to implement a custom action beside toggleComment in case of sub comment reply? Something like `toggleParentComment` which just calls toggleComment of the parent.

protected/humhub/modules/comment/resources/js/humhub.comment.js
MasterHans: Thank's for your comment, but no **userName** variable is global in this context... did not need to pass parameter

ArchBlood:
Is `public $publishOptions` really needed here? From my understanding this is used for development mode right?

MasterHans (commented):
> Is `public $publishOptions` really needed here? From my understanding this is used for development mode right?

Fixed!  You right! Thank you!

protected/humhub/modules/comment/Module.php
```diff
@@ -97,4 +97,29 @@ public function canComment($object)
         return true;
     }
 
+    /**
+     * Checks if given content object is a subcomment
+     *
+     * @param $object
+     * @return bool
```
buddh4, line 104: I'd rather move this as static function to the comment class. Also always add `@since` (in this case 1.8) to new functions. You can also refactor it as follows:

```
public function isSubComment($object)
{
    return $object instanceof Comment && $object->object_model === Comment::class
}
```

protected/humhub/modules/comment/Module.php
```diff
@@ -97,4 +97,29 @@ public function canComment($object)
...
+            return true;
+        }
+        return false;
+    }
+
+    /**
+     * Return parent of given content object
+     *
+     * @param $object
+     * @return ContentActiveRecord
```
buddh4, line 118: We can remove this function you can just call `$object->getPolymorphicRelation()`. Also I'd suggest always to use `findOne(['id' => $id])` since its more safe, probably not a problem in this case but we should stick to it.

buddh4 (changes requested):

protected/humhub/modules/comment/Module.php
MasterHans: Done! Thank you

protected/humhub/modules/comment/Module.php
MasterHans: Done! Thank you for your suggestion.

MasterHans:
I find one misfit problem. 
When I reply to the comment and insert the name of the author of the previous comment to the comment form the cursor only the first time shows up at the end of the comment. The second time when I replay to comment or sub comment the cursor appears at the beginning of the string. End that is not convenient for the client.
https://youtu.be/YnzYzpe25dc

buddh4:
@MasterHans You should never manually change the dom of a richtext field since it manages its state in the editor logic.
 I've added a mentioning link once we click on sub comment reply. One problem ist, that cursor is at the beginning of the richtext, but this can't be changed at the moment. I've planned some richtext enhancements for 1.8 perhaps we could solve this issue later.

MasterHans:
> @MasterHans You should never manually change the dom of a richtext field since it manages its state internally.
> I've added a mentioning link once we click on sub comment reply. One problem ist, that the cursor is at the beginning of of the line before the mentioning, but this can't be easily changed at the moment. I've planned some richtext enhancements for 1.8 perhaps we could solve this issue later.

Ok, I will know about richtext dom. Thank you!

buddh4:
@MasterHans sorry did miss to commit the guid addition for the container link, with latest commit it should work

MasterHans:
> @MasterHans sorry did miss to commit the guid addition for the container link, with latest commit it should work

Thanks! Everything ok)) I am observing **data-guid**. And waiting for instructions. All changes request was satisfied for now.
image

buddh4:
@MasterHans maybe do some further tests with the sub comment reply and mentioning link. I'll do a last review soon.

MasterHans:
> @MasterHans maybe do some further tests with the sub comment reply and mentioning link. I'll do a last review soon.

@buddh4 did you mean to create codeception test on this?

buddh4:
@MasterHans no, I mean just some further manual testing.

luke-:
Here is message from the my browser console. @buddh4 is this ok?
image

Otherwise it works for me and I would merge it!

@buddh4 It would be really cool if with 1.8 we could also set the cursor correctly.

buddh4:
@luke- Can you expand the message in the console? Does this occure every time you click on reply?

luke-:
@buddh4 the "unhandled" warning seems to appear on the first click of any like/comment links inside a wall entry.
image

But I don't know if it's a problem at all.

buddh4:
@luke- Ah ok, this is a normal debug output and is logged if the action is not directly handled by a handler of the element, but by the global action event hanlder.

protected/humhub/libs/Html.php
```diff
@@ -145,6 +145,7 @@ public static function backButton($url = '', $options = [])
     public static function containerLink(ContentContainerActiveRecord $container, $options = [])
     {
         $options['data-contentcontainer-id'] = $container->contentcontainer_id;
+        $options['data-guid'] = $container->guid;
```
buddh4, line 148: Maybe we should mention this in the changelog:

- Enh #4571: `humhub/libs/Html::containerLink()` now adds a  `data-guid` attribute

protected/humhub/modules/comment/widgets/views/comment.php
```diff
@@ -91,6 +92,8 @@
         <div class="wall-entry-controls">
             <?php if ($module->canComment($comment)): ?>
                 <?= CommentLink::widget(['object' => $comment]); ?>&nbsp;&nbsp;&middot;&nbsp;
+            <?php elseif (Comment::isSubComment($comment) && $module->canComment($comment->content->getPolymorphicRelation())): ?>
```
buddh4, line 95: Maybe it make sense to move this into the `CommentLink` widget.

protected/humhub/modules/comment/widgets/CommentLink.php
```diff
@@ -55,10 +55,6 @@ public function run()
             $this->mode = self::MODE_INLINE;
         }
 
-        if (!$module->canComment($this->object)) {
```
buddh4, line 58: The `CommentLink` is used in other places and modules too. We need to make sure that this widget is hidden if we are not allowed to comment. Maybe add the check in `comment.php` here.

luke-:
@buddh4 Ok great I wasn't sure. Is there anything on your side against a merge?

protected/humhub/modules/content/resources/js/humhub.ui.richtext.prosemirror.js
```diff
@@ -141,6 +141,18 @@ humhub.module('ui.richtext.prosemirror', function(module, require, $) {
 
     };
 
+    /**
+     * Builds mentioning string from container link
+     * @param $containerLink
+     * @returns {string}
```
buddh4, line 147: Can you add a `since 1.8` here

buddh4 (changes requested):

protected/humhub/libs/Html.php
MasterHans: Done ! But just want you to know  
This code is not in my branch
$options['data-guid'] = $container->guid;

protected/humhub/modules/comment/widgets/views/comment.php
MasterHans: Done!

protected/humhub/modules/comment/widgets/CommentLink.php
MasterHans: Done!

protected/humhub/modules/content/resources/js/humhub.ui.richtext.prosemirror.js
MasterHans: Done!

protected/config/common.php
```diff
@@ -6,4 +6,10 @@
  * @see http://docs.humhub.org/dev-environment.html
  */
 return [
+    'components' => [
```
luke-, line 9: Can you please revert this commit

luke- (changes requested):

protected/config/common.php
MasterHans: > Can you please revert this commit

Done!

Issue #4225, opened 2020-07-13
Delete pending registrations automatically


Semir1212:
Entries of pending registrations/invitations should be deleted automatically after a period of time or at least the admin should be able to 'select all' and delete them. The list gets quite messy after a while otherwise.

MasterHans:
@buddh4 Did you mean this list?
![invite-list](https://user-images.githubusercontent.com/17376109/96353711-51b8a300-10d7-11eb-82ed-109e1552ab0f.png)

luke-:
@MasterHans It's about this entries: https://github.com/humhub/humhub/blob/master/protected/humhub/modules/admin/controllers/PendingRegistrationsController.php#L44

siwa-tkapplmueller:
Wouldn't it make a lot more sense for admins to be able to disable / configure this feature in the admin settings rather than just automatically deleting all invites after a fixed period of time without any possibility to disable that or change the amount of days before deleting?

PR #4595, opened 2020-10-30
Delete pending registrations automatically #4225

MasterHans:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

1) Added **DeleteAll** button which has two actions depending on does any selected rows in the grid exists. First simply deletes all records in the **Invite** model. And second, delete selected rows in the grid. If the user selects at least one row in the grid the button text description changes from **"Delete All"** to **"Delete selected invitations"**.
2) Added **checkboxColumn** in grid
3) Created **humhub.admin.pending.registration.js** module. Put there all handling of deleting selected rows.


protected/humhub/modules/admin/resources/js/humhub.admin.pending.registrations.js
```diff
@@ -0,0 +1,51 @@
...
+                result = true;
+                return false;
+            }
+        });
+        return result;
+    }
+
+    var init = function (isPjax) {
+
+        $("input").change(function () {
```
buddh4, line 17: Note, once loaded a module will be initialized on every following pjax load. In the init function you should make sure you are residing in the right view, either by checking the existence of certain dom elements (selectors) or checking the [view state](https://github.com/humhub/humhub/blob/master/static/js/humhub/humhub.ui.view.js#L167). In this case you could also consider implementing a [JsWidget](https://docs.humhub.org/docs/develop/javascript-components#widgets)

Also its never a good idea to add such handlers to all `input` elements on the page, you should be more specific about which elements you want to listen to (e.g. only within a specific root element).

For example when implementing a JsWidget on the root element of the form, you could just call `this.$.find('input')`.

protected/humhub/modules/admin/views/pending-registrations/index.php
```diff
@@ -20,6 +24,9 @@
     </div>
 
     <div class="pull-right">
+        <?php if ($dataProvider->totalCount > 0): ?>
+            <?= Html::a(Yii::t('AdminModule.user', 'Delete All'), Url::toRoute(['/admin/pending-registrations/delete-all']), ['class' => 'btn btn-danger btn-delete']); ?>
```
buddh4, line 28: Rather use a js action with confirm e.g. `Button::danger('Yii::t('AdminModule.user', 'Delete All')')->action('admin.pending.registrations.deleteAll', Url::toRoute(['/admin/pending-registrations/delete-all'])->confirm()`

protected/humhub/modules/admin/resources/js/humhub.admin.pending.registrations.js
```diff
@@ -0,0 +1,51 @@
...
+            }
+        });
+
+        $(".btn-delete").on("click", function (e) {
+            var keys = $("#grid").yiiGridView("getSelectedRows");
+            if (keys.length > 0) {
+                e.preventDefault();
+                if(confirm("Delete selected invitations?")){
+                    $.ajax({
+                        url: "/index.php?r=admin%2Fpending-registrations%2Fdelete-all",
```
buddh4, line 33: Do not use static urls in javascript, this will fail for example when using url rewriting rules. Check https://docs.humhub.org/docs/develop/javascript-index#module-configuration for how to transfer such data to js as module configuration or when using a JsWidget, you can use the widget options which are provided as `data-*` attributes on the widget root.

protected/humhub/modules/admin/views/pending-registrations/index.php
```diff
@@ -6,6 +6,10 @@
 use yii\helpers\Url;
 
 /** @var $searchModel \humhub\modules\admin\models\PendingRegistrationSearch */
+/** @var $dataProvider yii\data\ActiveDataProvider */
+/** @var $types array */
+
+\humhub\modules\admin\assets\AdminPendingRegistrationsAsset::register($this);
```
buddh4, line 12: Prefer importing the namepsace.

buddh4 (changes requested):

protected/humhub/modules/admin/assets/AdminPendingRegistrationsAsset.php
```diff
@@ -0,0 +1,31 @@
...
+    ];
+    public $sourcePath = '@admin/resources';
+
+    /**
+     * @inheritdoc
+     */
+    public $js = [
+        'js/humhub.admin.pending.registrations.js'
+    ];
+
```
buddh4, line 30: You could register the url here e.g:

```
public static function register($view)
    {
        $view->registerJsConfig('admin.pending.registrations', [
            'deleteAllUrl' => Url::to('...')
        ]);

        return parent::register($view); // TODO: Change the autogenerated stub
    }
```

In you js module call:

```
var url = module.config.deleteAllUrl;
```

protected/humhub/modules/admin/resources/js/humhub.admin.pending.registrations.js
MasterHans: Done!

protected/humhub/modules/admin/views/pending-registrations/index.php
MasterHans: Done!

protected/humhub/modules/admin/views/pending-registrations/index.php
MasterHans: Done!

protected/humhub/modules/admin/assets/AdminPendingRegistrationsAsset.php
MasterHans: Done with JsWidget

protected/humhub/modules/admin/resources/js/humhub.admin.PendingRegistrations.js
```diff
@@ -0,0 +1,65 @@
...
+    PendingRegistrations.prototype.init = function() {
+        urlDeleteSelected = this.options.urlDeleteSelected;
+        this.$.find("input").change(function () {
+
+            var $checkBoxes = $('.regular-checkbox');
+
+            if (hasChecked($checkBoxes)) {
+                $('.delete-all').html('Delete selected rows')
+                $('.delete-all').attr('data-action-click','admin.PendingRegistrations.deleteAllSelected');
+                $('.delete-all').attr('data-action-click-url','/index.php?r=admin%2Fpending-registrations%2Fdelete-all-selected');
```
luke-, line 54: You always need to use `Url::to()` otherwise URL Rewriting is not supported.

protected/humhub/modules/admin/resources/js/humhub.admin.PendingRegistrations.js
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    PendingRegistrations.prototype.init = function() {
+        urlDeleteSelected = this.options.urlDeleteSelected;
+        this.$.find("input").change(function () {
+
+            var $checkBoxes = $('.regular-checkbox');
+
+            if (hasChecked($checkBoxes)) {
+                $('.delete-all').html('Delete selected rows')
```
luke-, line 52: Missing `Yii::t`

protected/humhub/modules/admin/controllers/PendingRegistrationsController.php
```diff
@@ -125,6 +125,51 @@ public function actionDelete($id)
...
+        if (Yii::$app->request->isPost) {
+               Invite::deleteAll();
+            $this->view->success(Yii::t(
+                'AdminModule.user',
+                'Deleted all invitations'
+            ));
+        }
+        return $this->redirect(['index']);
+    }
+    /**
```
luke-, line 147: \n

protected/humhub/modules/admin/assets/AdminPendingRegistrationsAsset.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+/**
+ * @link https://www.humhub.org/
+ * @copyright Copyright (c) 2017 HumHub GmbH & Co. KG
```
luke-, line 5: 2020

luke- (changes requested):
I've also added some findings.

If I change the browser scaling slightly, e.g. to 110%, it looks quite strange:
image

luke-:
We should also automatically delete invites that have not yet been registered after 90 days.
E.g. similar as here: https://github.com/humhub/humhub/blob/master/protected/humhub/modules/admin/Events.php#L53

protected/humhub/modules/admin/resources/js/humhub.admin.PendingRegistrations.js
buddh4: See https://docs.humhub.org/docs/develop/javascript-index#module-texts

protected/humhub/modules/admin/assets/AdminPendingRegistrationsAsset.php
```diff
@@ -0,0 +1,32 @@
...
+ * @license https://www.humhub.com/licences
+ */
+
+namespace humhub\modules\admin\assets;
+
+use yii\web\AssetBundle;
```
buddh4, line 11: Use the `humhub\components\assets\AssetBundle` instead

protected/humhub/modules/admin/controllers/PendingRegistrationsController.php
```diff
@@ -125,6 +125,51 @@ public function actionDelete($id)
...
+        if (Yii::$app->request->isPost) {
+            $ids = Yii::$app->request->post('id');
+            if (!empty($ids)) {
+                foreach ($ids as $id) {
+                    $invitation = Invite::findOne(['id' => $id]);
+                    $invitation->delete();
+                }
+                $this->view->success(Yii::t(
+                    'AdminModule.user',
+                    'Deleted all selected invitations'
```
buddh4, line 166: Selected invitations have been deleted

protected/humhub/modules/admin/controllers/PendingRegistrationsController.php
```diff
@@ -125,6 +125,51 @@ public function actionDelete($id)
...
+     * @throws HttpException
+     * @throws \Throwable
+     */
+    public function actionDeleteAll()
+    {
+        if (Yii::$app->request->isPost) {
+               Invite::deleteAll();
+            $this->view->success(Yii::t(
+                'AdminModule.user',
+                'Deleted all invitations'
```
buddh4, line 142: All invitations have been deleted

protected/humhub/modules/admin/resources/js/humhub.admin.PendingRegistrations.js
```diff
@@ -0,0 +1,65 @@
+humhub.module('admin.PendingRegistrations', function(module, require, $) {
+    var object = require('util').object;
+    var Widget = require('ui.widget').Widget;
+    var client = require('client');
+    var urlDeleteSelected;
+
+    var PendingRegistrations = function(node, options) {
```
buddh4, line 7: Just use:

```
var PendingRegistrations = Widget.extend();
```

and remove the following line:

```
object.inherits(PendingRegistrations, Widget);
```

protected/humhub/modules/admin/resources/js/humhub.admin.PendingRegistrations.js
```diff
@@ -0,0 +1,65 @@
...
+                url: urlDeleteSelected,
+                type: "POST",
+                data: {id: keys},
+            })
+        }).catch(function(e) {
+            module.log.error(e, true);
+        })
+    };
+
+    PendingRegistrations.deleteAll = function(evt) {
```
buddh4, line 26: Change functions to:

`PendingRegistrations.prototype.deleteAll` and `PendingRegistrations.prototype.deleteAllSelected` otherwise those function won't be bound to your widget instance and are only available as static functions, also you won't be able to access `this`.

Then change the action buttons to:

`->action('deleteAll', Url::toRoute(['/admin/pending-registrations/delete-all']))`

and

  $('.delete-all').attr('data-action-click','deleteAllSelected');`

As long as those buttons reside in your widget, the widget function will be called.

protected/humhub/modules/admin/resources/js/humhub.admin.PendingRegistrations.js
```diff
@@ -0,0 +1,65 @@
...
+        client.post(evt).then(function() {
+            that.reload();
```
buddh4, line 28: That is currently not initialized. Add `var that = this` prior to the `client.post()` call.

protected/humhub/modules/admin/resources/js/humhub.admin.PendingRegistrations.js
```diff
@@ -0,0 +1,65 @@
...
+            }
+        });
+        return result;
+    }
+
+    PendingRegistrations.prototype.init = function() {
+        urlDeleteSelected = this.options.urlDeleteSelected;
+        this.$.find("input").change(function () {
+
+            var $checkBoxes = $('.regular-checkbox');
```
buddh4, line 49: I think `var $selection = that.$.find(':checked')` should to the trick and we can remove the `hasChecked()` function then just:

```
if($selection.length) {
 // Do stuff
}
```

(of course you'll have to add the `var that = this` prior to ` this.$.find("input").change(function () {`)

protected/humhub/modules/admin/resources/js/humhub.admin.PendingRegistrations.js
```diff
@@ -0,0 +1,65 @@
...
+            if ($(this).prop("checked")) {
+                result = true;
+                return false;
+            }
+        });
+        return result;
+    }
+
+    PendingRegistrations.prototype.init = function() {
+        urlDeleteSelected = this.options.urlDeleteSelected;
```
buddh4, line 46: I'd rather directly check `this.options.urlDeleteSelected` in the `deleteAllSelected ` instead of adding this static variable.

protected/humhub/modules/admin/widgets/views/PendingRegistrationsWidgetView.php
```diff
@@ -0,0 +1,88 @@
...
+                return isset($types[$data->source]) ?: Html::encode($data->source);
+            },
+        ],
+        [
+            'header' => Yii::t('AdminModule.user', 'Actions'),
+            'class' => 'yii\grid\ActionColumn',
+            'template' => '{resend} {delete}',
+            'buttons' => [
+                'resend' => function ($url, $model, $key) {
+                    return Html::a(
```
buddh4, line 70: `Button::primary()->link(Url::to(['resend', 'id' => $model->id]))->icon('envelope')->xs()` we don't need the `tt` class since we don't have a tooltip.

protected/humhub/modules/admin/widgets/views/PendingRegistrationsWidgetView.php
```diff
@@ -0,0 +1,88 @@
...
+                        '<i class="fa fa-envelope"></i>',
+                        Url::to(['resend', 'id' => $model->id]),
+                        ['class' => 'btn btn-primary btn-xs tt']
+                    );
+                },
+                'delete' => function ($url, $model, $key) {
+                    return Html::a(
```
buddh4, line 77: `Button::primary()->link(Url::to(['delete', 'id' => $model->id]))->icon('delete')->xs()` we don't need the `tt` class since we don't have a tooltip.

buddh4 (changes requested):

protected/humhub/modules/admin/widgets/views/PendingRegistrationsWidgetView.php
buddh4: I think those two functions should be `post` links instead.

protected/humhub/modules/admin/resources/js/humhub.admin.pending.registrations.js
MasterHans: Done with Widget!

protected/humhub/modules/admin/resources/js/humhub.admin.PendingRegistrations.js
MasterHans: Done!

protected/humhub/modules/admin/resources/js/humhub.admin.PendingRegistrations.js
MasterHans: Done!

MasterHans: > I've also added some findings.
> 
> If I change the browser scaling slightly, e.g. to 110%, it looks quite strange:
> image

Done!

protected/humhub/modules/admin/controllers/PendingRegistrationsController.php
MasterHans: @luke- I am sorry did not catch what does it mean...

protected/humhub/modules/admin/assets/AdminPendingRegistrationsAsset.php
MasterHans: done!

protected/humhub/modules/admin/assets/AdminPendingRegistrationsAsset.php
MasterHans: Done!

protected/humhub/modules/admin/controllers/PendingRegistrationsController.php
MasterHans: Done!

protected/humhub/modules/admin/resources/js/humhub.admin.PendingRegistrations.js
MasterHans: done!

protected/humhub/modules/admin/resources/js/humhub.admin.PendingRegistrations.js
MasterHans: Done! Thank you!

protected/humhub/modules/admin/resources/js/humhub.admin.PendingRegistrations.js
MasterHans: Done with client.reload()
Redirect in controller also works, but I did not want to leave empty  
client.post(evt).then(function()  like this:

```
        client.post(evt).then(function () {

        }).catch(function (e) {
            module.log.error(e, true);
        })
```


)

protected/humhub/modules/admin/resources/js/humhub.admin.PendingRegistrations.js
MasterHans: Done!

protected/humhub/modules/admin/widgets/views/PendingRegistrationsWidgetView.php
MasterHans: Done!

protected/humhub/modules/admin/widgets/views/PendingRegistrationsWidgetView.php
MasterHans: Done!

protected/humhub/modules/admin/resources/js/humhub.admin.PendingRegistrations.js
buddh4: This would have been an alternative:

```
client.post(evt)..catch(function (e) {
    module.log.error(e, true);
});
```

In this case you could just omit the `then` handler.

protected/humhub/modules/admin/resources/js/humhub.admin.PendingRegistrations.js
MasterHans: Done!

protected/humhub/modules/admin/resources/js/humhub.admin.PendingRegistrations.js
MasterHans: Done!

protected/humhub/modules/admin/controllers/PendingRegistrationsController.php
luke-: @MasterHans There is a missing newline between the end of the method and the next one.

luke-:
Looks quite good for me! Can you also implement an automatic cleanup after 90 days, similar to the logfile cleanup.

Would be great if the auto delete time could be configured as module variable, e.g.:
https://github.com/humhub/humhub/blob/master/protected/humhub/modules/admin/Module.php#L56

MasterHans:
> We should also automatically delete invites that have not yet been registered after 90 days.
> E.g. similar as here: https://github.com/humhub/humhub/blob/master/protected/humhub/modules/admin/Events.php#L53

Done!

protected/humhub/modules/admin/jobs/CleanupPendingRegistrations.php
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * @since 1.8
+ */
+
+class CleanupPendingRegistrations extends ActiveJob
+{
+    /**
+     * @var int seconds before delete old pending registrations messages
+     */
+    public $cleanupInterval = 60 * 60 * 24 * 90;
```
ArchBlood, line 26: @luke- question about the interval, is this related to Cron? Then wouldn't `60 * 60 * 24 * 7` be for the best? 
### See;
https://github.com/humhub/humhub/blob/a5da4c1b0b9eae3c38345f46abfa93ab8d030d42/protected/humhub/modules/admin/jobs/CleanupLog.php#L23-L26

Although I'm thinking something like `59 59 23 * * 7 *` would be better? As some Cron tasks don't calculate the same way :thinking:

protected/humhub/modules/admin/controllers/PendingRegistrationsController.php
MasterHans: Done!

MasterHans:
> We should also automatically delete invites that have not yet been registered after 90 days.
> E.g. similar as here: https://github.com/humhub/humhub/blob/master/protected/humhub/modules/admin/Events.php#L53

Done!

MasterHans:
> We should also automatically delete invites that have not yet been registered after 90 days.
> E.g. similar as here: https://github.com/humhub/humhub/blob/master/protected/humhub/modules/admin/Events.php#L53

I got it about variable, but
I am sorry I did not catch  about  automatically delete invites - I have already implemented your previous task :
 
> > We should also automatically delete invites that have not yet been registered after 90 days.
> > E.g. similar as here: https://github.com/humhub/humhub/blob/master/protected/humhub/modules/admin/Events.php#L53
>

MasterHans:
> Would be great if the auto delete time could be configured as module variable, e.g.:
> https://github.com/humhub/humhub/blob/master/protected/humhub/modules/admin/Module.php#L56

Done!

luke-:
There is should be no difference. You had only created an additional feature to delete invites with checkboxes.

MasterHans:
> There is should be no difference. You had only created an additional feature to delete invites with checkboxes.
I am sorry but I had created this:
Did you saw this commit
image
Please check out this commit

luke- (approved):

protected/humhub/modules/admin/jobs/CleanupPendingRegistrations.php
ArchBlood: When commenting on this, I was wondering if this would work for all systems if it were related to cron jobs :thinking:

Issue #3414, opened 2019-01-22
User admin should be able to set/change user passwords


luke-:

yurabakhtin:
@luke- Please see where I added the admin password edit form. Password is updated only when the fields are filled both without errors, if they are empty then password will not be changed:
<img width="1148" alt="admin_edit_pass_form" src="https://user-images.githubusercontent.com/6995524/97894503-9f721400-1d43-11eb-9b38-264beea6a63d.png">

luke-:
Looks fine, we'll integrate this in 1.8

yurabakhtin:
@luke- I detected a bug related to this task:
<img width="1163" alt="bug" src="https://user-images.githubusercontent.com/6995524/98337917-67f0b980-201a-11eb-8c83-d0830704f182.png">

yurabakhtin:
Fixed in commit: https://github.com/humhub/humhub/pull/4598/commits/d066113e499d15a43690834d7babc1a37b511d89

luke-:
Awesome!

PR #4598, opened 2020-11-02
Allow to change user passwords by admin

yurabakhtin:

- [x] Feature
- [x] Fix undefined `UserEditForm::newPassword` for users without record in DB table `user_password` - https://github.com/humhub/humhub/issues/3414#issuecomment-722924295

**Other information:**

Issue: https://github.com/humhub/humhub/issues/3414

### DIFF
--- a/protected/humhub/modules/admin/controllers/UserController.php
+++ b/protected/humhub/modules/admin/controllers/UserController.php
@@ -11,6 +11,7 @@ use humhub\compat\HForm;
 use humhub\components\export\DateTimeColumn;
 use humhub\components\export\SpreadsheetExport;
 use humhub\modules\admin\components\Controller;
+use humhub\modules\admin\models\forms\PasswordEditForm;
 use humhub\modules\admin\models\forms\UserDeleteForm;
 use humhub\modules\admin\models\forms\UserEditForm;
 use humhub\modules\admin\models\UserSearch;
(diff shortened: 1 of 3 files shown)


luke- (approved):

protected/humhub/modules/admin/models/forms/PasswordEditForm.php
```diff
@@ -0,0 +1,45 @@
...
+{
+    const SCENARIO_EDIT_ADMIN = 'editAdmin';
+
+    /**
+     * @inheritdoc
+     */
+    public function init()
+    {
+        parent::init();
+        $this->scenario = self::SCENARIO_EDIT_ADMIN;
```
buddh4, line 21: Just wondering, why do we need this scenario? I don't see any other use-cases of this class.

buddh4: Ah ok  its derifed from Password model.

buddh4 (approved):

protected/humhub/modules/admin/controllers/UserController.php
```diff
@@ -150,6 +160,26 @@ public function actionEdit()
             ];
         }
 
+        // Change Password Form
+        if ($canEditAdminFields) {
+            $definition['elements']['Password'] = [
+                'type' => 'form',
+                'title' => 'Change Password',
```
luke-, line 167: -  Can you please make sure to use `Yii::t` here.
- Only the word "Password" should be enough here.

luke- (changes requested):

protected/humhub/modules/admin/controllers/UserController.php
yurabakhtin: - done in https://github.com/humhub/humhub/pull/4598/commits/f72fefec9a0950e5d6f4724b0235c0aefcd658de
- it seems I copied that from tab "Account" which was not translatable either, I fixed that too there.

luke- (approved):

Issue #3996, opened 2020-04-15
Force Password Change


Ockenheimer:
### Is your feature request related to a problem? Please describe.

When adding new users, you are able to set a password. These pw are active and given to the user. 
The user itself dont have to change the password. 
When setting up Humhub in a company without LDAP (we are using humhub as a social network in a kindergarden), we have to add mayn users and must provide each an easy to remember but unique password.

### Describe the solution you'd like
When adding a user, it would help to mark the password as deprecated, so that the user is forced to change the password on the first login.

### Describe alternatives you've considered
the alternative would be, to generate passwords via a pw-manager, which is not provided in our environment. Setting up all users with the same password without changing it, would be a huge security-problem.

yurabakhtin:
@luke- Please confirm it should be implemented like this scheme:

<img width="1151" alt="force_password" src="https://user-images.githubusercontent.com/6995524/97959179-d4bc4780-1dbf-11eb-86a3-60d97f1b04d2.png">

> We should add a flag maybe in the "contentcontainer_settings" of the user.

I see the db table `contentcontainer_setting` contains settings per module and per some container, as I understand there we can keep some custom settings per User from the module `user`, right?
If yes, do you mean to store there new flag for the checkbox "[√] Force to change this password on first log in" instead of extend db tables like `user` or `user_password`?

luke-:
@yurabakhtin Yes, but I would use this layout: (When Guest Access is disabled at: Administration -> User -> Settings -> Allow limited access for non-authenticated users (guests)")

![image](https://user-images.githubusercontent.com/4736168/97967230-3c24c800-1dbd-11eb-8b30-591c9cac94d2.png)

Regarding Settings something like:

```php
/** @var Module $module */
$module = Yii::$app->getModule('user');
$module->settings->user()->set('mustChangePassword', true);
```

yurabakhtin:
@luke- Please review the PR https://github.com/humhub/humhub/pull/4602.
<img width="951" alt="test_force_pwd" src="https://user-images.githubusercontent.com/6995524/98273488-620bc180-1fa3-11eb-95df-8f7bbf213ecf.png">

PR #4602, opened 2020-11-04
New setting to force to change password on first log in

yurabakhtin:

- [x] New setting "Force to change password on first log in" on creating user by admin
- [x] Implement new form to force password changing when user has new setting `contentcontainer_setting::mustChangePassword`
- [x] Prevent periodical xhr requests to `/live/poll`

**Other information:**

Issue: https://github.com/humhub/humhub/issues/3996


yurabakhtin:
> Prevent periodical xhr requests to `/live/poll`

prevent_xhr

I think the xhr requests might be prevented here https://github.com/humhub/humhub/blob/enh/3996-force-password-change/protected/humhub/modules/live/resources/js/humhub.live.js#L10 with code like this:

```
        if (user.mustChangePassword()) {
            return;
        }
```

yurabakhtin:
@luke- Also I don't like my this solution https://github.com/humhub/humhub/pull/4602/files#diff-58224309d127ec31027d9beae8cefa081ae86c16bc92f5426d3b8f309ceab9b5R151-R152:

```
            } else if ($this->controllerAccess->code == 403) {
                return $this->mustChangePassword();
            }
```

the 403 code is set here https://github.com/humhub/humhub/pull/4602/files#diff-230df3417b1e6af913d2567248216004090ea2735ad49c67502cc3a2262ad5acR244, but if in future some other validators will use the 403 code too, then this code will not work, e.g. currently there many validators use 401 code which is used for login page restriction.
So I think to change the condition to like:

```
            if (isset($this->controllerAccess->some_callback)) {
                return call_user_func([$this, $this->controllerAccess->some_callback]);
            } else if ($this->controllerAccess->code == 401) {
```

I will investigate how to make this better, and it would be good if you will advise me some known solution.

yurabakhtin:
> @luke- Also I don't like my this solution https://github.com/humhub/humhub/pull/4602/files#diff-58224309d127ec31027d9beae8cefa081ae86c16bc92f5426d3b8f309ceab9b5R151-R152:
>
> ...
>
> I will investigate how to make this better, and it would be good if you will advise me some known solution.

I have implemented a solution here https://github.com/humhub/humhub/pull/4602/commits/015c87d374746702d76607368e2d1b71f2ffdccc, let me know if you don't agree something there, otherwise this PR is completed and can be merged.

luke-:
Great!

Can we also add a option here for existing users?
image

yurabakhtin:
@luke- yes we can.

yurabakhtin:
@luke- 

> Can we also add a option here for existing users?

Done, please note you named the option as "Must change password at next login" but really it works immediately even when user is currently logged in, I mean if user is logged in and admin change the option then it will take effect right on next any page load because it works as global access restriction, i.e. admin enabled the option then user will be blocked by single page `/user/must-change-password` immediately and otherwise.
If we don't need such global blocking then we should improve this, otherwise the option label must be renamed to "Must change password" without "at next login".

protected/humhub/modules/user/components/User.php
```diff
@@ -36,6 +36,11 @@ class User extends \yii\web\User
      */
     protected $permissionManager = null;
 
+    /**
+     * @var string URL to force user to change password
+     */
+    public $mustChangePasswordUrl = '/user/must-change-password';
```
buddh4, line 42: I'd rename it to `$mustChangePasswordRoute` since its not an url

protected/humhub/modules/user/views/must-change-password/index.php
```diff
@@ -0,0 +1,51 @@
...
+        <div id="must-change-password-form" class="panel panel-default animated bounceIn"
+             style="max-width: 300px; margin: 0 auto 20px; text-align: left;">
+            <div class="panel-heading"><?= Yii::t('UserModule.auth', '<strong>Change</strong> password'); ?></div>
+            <div class="panel-body">
+
+                <?php $form = ActiveForm::begin(); ?>
+
+                <p><?= Yii::t('UserModule.auth', 'Please change your password.'); ?></p>
+
+                <?php echo $form->field($model, 'currentPassword')->passwordInput(['maxlength' => 45]); ?>
```
buddh4, line 28: Please rather use `<?=`

buddh4 (changes requested):
I'am not a fan of the solution to manually block  `PollController` and that the `mustChangePassword` setting is transfered to the js `user` module. Maybe we find another solution for this. We probably need a general way of blocking or disabling the live provider in such cases.

E.g:

- https://github.com/humhub/humhub/pull/4602/files#diff-76a9bb6021fd7c3a95159c19587aa1a479067caa90f4b863bc4007e588bf9deaR11
- https://github.com/humhub/humhub/pull/4602/files#diff-317f0fa4325a73e2ab8ce7aae8f64d1fbc6819d1a4cc62a246d60e680189a28fR36
- https://github.com/humhub/humhub/pull/4602/files#diff-2daff17675bbb32d8376390f00578ebc6b6631aadf0d0bfb7ddae2cb9808fc0bR7

Suggestion:

- Check if user must change password in controller access
- If yes and its a pajx or full page load
- Redirect to must change password page and disable live module by js configuration (e.g. add `isActive` js config)
- If yes and its a normal ajax call, just return 403

We could also think about using a custom http code so clients as the live module could detect the issue and stop polling.

protected/humhub/components/behaviors/AccessControl.php
```diff
@@ -146,7 +146,10 @@ public function beforeAction($action)
         $this->controllerAccess = $this->getControllerAccess($this->rules);
 
         if (!$this->controllerAccess->run()) {
-            if ($this->controllerAccess->code == 401) {
+            if (isset($this->controllerAccess->codeCallback) &&
```
buddh4, line 149: Could you add some further docs for usage and use-cases of `codeCallback`

protected/humhub/components/behaviors/AccessControl.php
```diff
@@ -220,4 +223,14 @@ protected function loginRequired()
 
         return false;
     }
+
+    /**
+     * @return Response Redirect user to force to change password
+     */
+    protected function mustChangePassword()
```
buddh4, line 230: Maybe rename to `forceChangePassword` for me there are a bit too many new `mustChangePassword` functions :)

protected/humhub/components/behaviors/AccessControl.php
```diff
@@ -220,4 +223,14 @@ protected function loginRequired()
...
+    {
+        if (!Yii::$app->user->isMustChangePasswordUrl()) {
+            return Yii::$app->getResponse()->redirect(Yii::$app->user->mustChangePasswordUrl);
```
buddh4, line 233: Maybe check if its a pjax or full page load. In case of simple ajax call it should just return a forbidden code. Maybe this would also partly solve the PollController issue?

yurabakhtin:
@buddh4 

> I'am not a fan of the solution to manually block `PollController` and that the `mustChangePassword` setting is transfered to the js `user` module. Maybe we find another solution for this. We probably need a general way of blocking or disabling the live provider in such cases.

I created the new js method `user.mustChangePassword()` and config `module.config.mustChangePassword` similarly already existing `user.isGuest()` and `module.config.isGuest`.

As I understood you want to add new config property `isActive` instead of `mustChangePassword`.
Do you think to combine the `isGuest` and `mustChangePassword` into new `isActive`?
We can do this on js module side in the file `humhub.user.js` like code:

```
var isActive = function() {
    return !module.config.isGuest && !module.config.mustChangePassword;
};
```

either we can do this from php side in the file `CoreJsConfig.php` like code:

```
$userConfig['isActive'] = !Yii::$app->user->isGuest && !Yii::$app->user->mustChangePassword();
```

---

Now I have found the `user.isGuest()` is used also in the file `protected/humhub/modules/notification/resources/js/humhub.notification.js` before init the module, probably we need to use the same restriction there too, but I don't know this code and I didn't get any error there when I was developing the new feature "must change password". I did the restriction in live module only because there I have a page refreshing and xhr requestes each 3-5 seconds so I limited that as it was limited for guest users.

protected/humhub/modules/user/components/User.php
yurabakhtin: Ok, I renamed that to `$mustChangePasswordRoute`, but please note I named this as `...Url' because here https://github.com/humhub/humhub/blob/enh/3996-force-password-change/protected/humhub/config/web.php#L36 we have `'loginUrl' => ['/user/auth/login'],` as well.

protected/humhub/modules/user/views/must-change-password/index.php
yurabakhtin: ah yes, I just copied this from another file, maybe this https://github.com/humhub/humhub/blob/enh/3996-force-password-change/protected/humhub/modules/user/views/account/changePassword.php#L13.
Fixed in commit https://github.com/humhub/humhub/commit/fc918208d1605da4b39bc739200502ae3f48c43f.

protected/humhub/components/behaviors/AccessControl.php
yurabakhtin: I added a doc right in the place, commit https://github.com/humhub/humhub/pull/4602/commits/003cc56f842f7fb0471ee1f812324476f89af444,
but doc is also in declaration https://github.com/humhub/humhub/pull/4602/files#diff-230df3417b1e6af913d2567248216004090ea2735ad49c67502cc3a2262ad5acR205.

protected/humhub/components/behaviors/AccessControl.php
yurabakhtin: ok, commit https://github.com/humhub/humhub/pull/4602/commits/6444ceb70ce37b9a2b8251407d4c491edb0ccbf8

protected/humhub/components/behaviors/AccessControl.php
yurabakhtin: Please see my investigation with code:

```
if (Yii::$app->getRequest()->getIsAjax()) {
    $this->forbidden();
}
```
investigate_ajax_restriction

---

I.e. with current code if user becomes blocked with option "must change password" then auto redirected to `/user/must-change-password` on ajax request, but if we will add the forbidden exception code then user will stay on the same page without redirect, do you think it is better? If yes, I will add the code.

> Maybe this would also partly solve the PollController issue?

I didn't see the poll requests are locked by access controller, they continue work with 200 code.

luke-:
We will also need something similar for the upcoming 2FA module. Ideally, the solution could be controllered by a module.

buddh4:
What I'am concerned of is, that now all modules which may implement a similiar feature as the live poll now must know that there is a feature `mustChangePassword` and directly use functionality of this feature. For example the mail module used to implement a custom polling mechanism, which we would probably have to stop in this situation.

Also there may be other feautres in the future which will require stopping such polling or other background work etc. E.g. a "Under Maintainance" view.

I think, what we need is a mechnaism to stop such background work/request indepdendently of the `mustChangePassword` feature. My first suggestion was that the `mustChangePassword` view can just stop the polling manually by js configuration. So the mustChangePassword knows about the polling client but not the other way around.
This way the live module would not need to know anything about the `mustChangePassword` and we won't need any js code related to this.

If we want a general way of stopping the poll we could introduce a such a blocking config-flag/function to the `view` or perhaps `client` js module independently of `mustChangePassword`.

buddh4:
What about just adding a `isActive` config setting to the live js module. And setting something like this in the `CoreJsConfig`

```
'live' => [
   'isActive' => !Yii::$app->user->isGuest && !Yii::$app->user->mustChangePassword();
]
```

Then we can think about how to handle other background jobs in another PR.

yurabakhtin:
@buddh4 Ok, as I understand you prefer to set new `isActive` from server side:

1. `Yii::$app->getModule('live')->isActive`: do we need to add a new `public $isActive = true;` in the class `\humhub\modules\live\Module`? Or we should init the `$this->isActive = !Yii::$app->user->isGuest && !Yii::$app->user->mustChangePassword()` inside the method `\humhub\modules\live\Module->init()`?
2. Currently we have `CoreJsConfig` options `['user']['isGuest']` and `['user']['mustChangePassword']`, do you want to remove them from `['user']` and create a new `['live']['isActive']` instead?
3. The JS methods `user.isGuest()` and `user.mustChangePassword()` will be removed and new `live.isActive()` will be created instead in order to use new config `module.config.isActive`, right?
4. In all places where the js `if (user.isGuest() || user.mustChangePassword())` is used we will replace with new `if (live.isActive())`, right?

buddh4:
@yurabakhtin 

1. I'd set `public $isActive = true;` in the module class and for now just add the `!Yii::$app->user->isGuest && !Yii::$app->user->mustChangePassword()` in the CoreJsConfig. The idea is that modules can just switch the `isActive` if required and the live module does not know about details such as `mustChangePassword`. So we could also consider setting `Yii::$app->getModule('live')->isActive = false` within the mustChangePassword controller, but this is up to you.

2. Leave the ` $userConfig['isGuest']` setting and remove the `['user']['mustChangePassword']` since we do not require it elsewhere and add the `isActive` live config to https://github.com/humhub/humhub/blob/master/protected/humhub/widgets/CoreJsConfig.php#L42

3. Yes, remove those two checks in the live module, you can remove the `mustChangePassword` function from the user js module since we do not currently use it, but do not remove the `isGuest` method from the user module, since its used in other places.

4. Yes, this should only be used in the live module itself anyways as far as I see in the PR?

buddh4:
Just as clarification why I consider the `isGuest` check to be fine but the `mustChangePassword` not. Guest support is a fundamental concept of HumHub and all module developers should be familiar with the concept and almost all modules contain logic to enable or disable guest support. The `mustChangePassword` on the other hand is a rather specific detail of the user module and instead of requiring other modules to know about this feature and react to the behavior the feature itself should be responsible to set the behavior as far as possible in order to keep the footprint of this feature as small as possible.

yurabakhtin:
@buddh4 please check commit https://github.com/humhub/humhub/pull/4602/commits/0e1c51a78f40282579ec2a74c9bd1f1c9f86b8e5

buddh4:
@yurabakhtin looks good! Haven had the time to test yet.

luke-:
@buddh4 Please take a look at it when you have some time.

protected/humhub/modules/user/components/User.php
buddh4: You are right its passed to ` Yii::$app->getResponse()->redirect` which also supports real urls same with `loginUrl`.

protected/humhub/modules/live/controllers/PollController.php
```diff
@@ -53,6 +53,10 @@ public function beforeAction($action)
             return false;
         }
 
+        if (Yii::$app->user->mustChangePassword()) {
```
luke-, line 56: Just for my understanding, this is actually not necessary and also not needed for other similar modules (like TwoFactor Auth), because by the Poll Module variable "isActive" the polling is deactivated anyway?

buddh4: https://github.com/humhub/humhub/pull/4602/files#diff-0dee4426f877787aaeae9be039601f75e47410fba822e0c39ecf8c1cea0e4a7aR35 will deactivate the polling client in the current tab. But the live provider in other browser tabs will still be active. Not sure if this is a problem.

buddh4: Maybe we should add a user setting independent of mustchangepassword if this user setting flag is set, the poll controller returns a 503 https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503.

luke-: @buddh4 I would leave it at the current solution for now.

A setting is probably not dynamic enough, e.g. with TwoFactor Auth. If it is necessary, however, I would tend to a new event that checks whether the background tasks should be disabled.

Apart from that, had you been able to have a look at the PR? I would like to complete the task otherwise. It looks good to me.

buddh4: Not sure what you mean with "not dynamic enough"? The setting is more or less an abstraction of the current behavior and allows features as "must change password" and "maintainance mode" to inform background services to not deliver content and inform clients as the "poll client" to handle such a state accordingly. Currently the poll client in other tabs will just keep requesting and logging errors. This state configuration won't affect any controller unless a controller like the poll controller manually checks such a state.

luke-: Because of the "not dynamic" I meant that settings are mostly stored in the database.

How about a simple value in the session? That would be super performant (especially with polling, we should avoid any queries there as much as possible).

Then MustChangePassword,2FA,e.g. could set this session variable (activate limited mode) and then remove it again.
In the case of an error, there would also be no persistent settings in the database.

buddh4: This will only work on a single browser/session basis and not between different devices etc. Currently We need a proper way of informing the frontend about such state changes in the backend and broatcast this state to all tabs/browses/devices which are currently active. Its the same problem with session timeout, user logout, session renew etc. I think we should not discuss this here, but rather open an additional issue/pr for this.

luke-: @buddh4 Ok I have created an issue here for this. However, it is more of a placeholder, we still need to define more precisely what we need here.
https://github.com/humhub/humhub/issues/4760

luke- (approved):

luke- (approved):

beankylla:
Hello,
Sorry to bug but is this enabled by default? It does not seem to be and i can find nowhere in the doc where it is mentionned.
Unless it somehow does not apply to massimport created users?
BR,

Issue #4670, opened 2020-11-27
Editable ContentContainer Default Permissions


luke-:
Make ContentContainer (e.g. Space / User) Permissions editable by administators.

Space Admin:
![image](https://user-images.githubusercontent.com/4736168/100457738-e218d900-30c2-11eb-87d6-b11727b73a24.png)

User Admin:
![image](https://user-images.githubusercontent.com/4736168/100457903-27d5a180-30c3-11eb-829e-857fa2ecda1c.png)

### Is your feature request related to a problem? Please describe.

Configuration of default Space permissions is rather complex.
e.g. https://docs.humhub.org/docs/admin/permissions

### Describe the solution you'd like

Permission editor grid on Space settings menu

yurabakhtin:
@luke- It seems you attached the same image for "User Admin:" as for "Space Admin, please confirm you meant this:
<img width="1153" alt="user_admin" src="https://user-images.githubusercontent.com/6995524/100569828-45338700-32e0-11eb-8e5f-05c736b869df.png">
"

luke-:
@yurabakhtin Oh, yes! This would of course be the best place for the profile default settings.

yurabakhtin:
@luke- https://github.com/humhub/humhub/pull/4678/commits/62f14bb3cc70a0a4c5d8512a0ee3af1c30748875 Please see how I implemented default permissions for Spaces:
<img width="1148" alt="prview_space_default_perms" src="https://user-images.githubusercontent.com/6995524/100609830-b9d9e600-331f-11eb-8870-ec94d5b9a803.png">

As you can see many permissions are disabled. I specially did this because I think almost all modules has a not completed initialization for permissions.
Only single core module was implemented enough, please see here what I mean https://github.com/humhub/humhub/blob/enh/4670-editable_contentcontainer_default_permissions/protected/humhub/modules/post/Module.php#L32.
`new permissions\CreatePost(['contentContainer' => $contentContainer])` - the Content Container is passed there, so we can know what Model(Space or User) where have there.
Also you can see on my screenshot a Space permission is active only, it is why I did this here https://github.com/humhub/humhub/commit/62f14bb3cc70a0a4c5d8512a0ee3af1c30748875#diff-2672f0ad55d9b222ba975092e5439006c0058e175b02c4a40cecc70eef4d601bR70.
But all other modules don't initialize the Content Container, I suggest to modify ALL modules, firstly only core and then all external modules to do this fix. After that we can edit all default permissions.

---
[...]

PR #4678, opened 2020-11-30
Default Permissions for ContentContainer(Space/User)

yurabakhtin:

- [x] Default permissions for Spaces
- [x] Default permissions for Users
- [x] Pass Content Container on initialize permissions in `Module->getPermissions()` for **core** modules
- [x] Unit tests
- [x] Caching: Since the default state probably has to be checked often, we should pay special attention to caching and try to minimize DB queries.
- [x] **?** Investigate probably unnecessary columns `class` from the tables `contentcontainer_default_permission` and `contentcontainer_permission`, see https://github.com/humhub/humhub/issues/4670#issuecomment-737007280. **Answered** here https://github.com/humhub/humhub/issues/4670#issuecomment-737266493
- [ ] **?** Decide to complete converting `$space->getUserGroups()` into `Space::getUserGroups()`
- [ ] **?** Decide to revert the passing Content Container from list below:
- [x] Pass Content Container on initialize permissions in `Module->getPermissions()` for **external** modules:
  - [x] Free:
    - [x] cfiles - https://github.com/humhub/humhub-modules-cfiles/pull/101
    - [x] mail - https://github.com/humhub/humhub-modules-mail/pull/206
    - [x] calendar - Doesn't have permissions
    - [x] wiki - Doesn't have permissions
    - [x] updater - Doesn't have permissions
    - [x] translation - Doesn't have permissions
    - [x] tasks - Doesn't have permissions
    - [x] polls - https://github.com/humhub/humhub-modules-polls/pull/76
    - [x] custom-pages - Doesn't have permissions
    - [x] 2fa - Doesn't have permissions
    - [x] rest - Doesn't have permissions
    - [x] example-basic - Doesn't have permissions
    - [x] advanced-ldap - Not found/implemented or not free module
    - [x] fcm-push - Doesn't have permissions
    - [x] theme-builder - Not found/implemented or not free module
    - [x] translation-manager - Not found/implemented or not free module
    - [x] solr - Not found/implemented or not free module
    - [x] linkpreview - Not found/implemented or not free module
    - [x] enterprise-theme - Not found/implemented or not free module
    - [x] meetings - Not found/implemented or not free module
    - [x] massuserimport - Not found/implemented or not free module
  - [x] Contrib:
    - [x] gallery - Doesn't have permissions
    - [x] youracclaim - Doesn't have permissions
    - [x] termsbox - Doesn't have permissions
    - [x] sms - Doesn't have permissions
    - [x] reportcontent - Doesn't have permissions
    - [x] popover-vcard - Doesn't have permissions
    - [x] notes - Doesn't have permissions
    - [x] newmembers - Doesn't have permissions
    - [x] mostactiveusers - Doesn't have permissions
    - [x] linklist - Doesn't have permissions
    - [x] jitsi-meet - Doesn't have permissions
    - [x] drawio - Doesn't have permissions
    - [x] breakingnews - Doesn't have permissions
    - [x] birthday - Doesn't have permissions
    - [x] auto-follow - Doesn't have permissions
    - [x] legal - Doesn't have permissions
    - [x] devtools - Doesn't have permissions
    - [x] profiler - Doesn't have permissions
    - [x] external-calendar - https://github.com/humhub-contrib/external-calendar/pull/21
    - [x] auth-google - Doesn't have permissions
    - [x] fcm-push - Doesn't have permissions
    - [x] katex - Doesn't have permissions
    - [x] blog - Doesn't have permissions
    - [x] examples - Not module
    - [x] image-compressor - Empty repository
    - [x] sharebetween - Doesn't have permissions
    - [x] social-share - Doesn't have permissions
    - [x] onlydocuments - Doesn't have permissions
    - [x] dropbox - Doesn't have permissions / Deprecated / Archived
    - [x] webshell - Doesn't have permissions
    - [x] block-profile-changes - Doesn't have permissions

- [x] Changelog was modified

**Other information:**

Issue: https://github.com/humhub/humhub/issues/4670


protected/humhub/modules/content/components/ContentContainerDefaultPermissionManager.php
```diff
@@ -0,0 +1,86 @@
...
+     */
+    public $contentcontainer_class = null;
+
+    /**
+     * @inheritdoc
+     */
+    protected function getModulePermissions(\yii\base\Module $module)
+    {
+        if ($module instanceof \humhub\components\Module) {
+            return $module->getPermissions(new $this->contentcontainer_class);
```
buddh4, line 33: Can't we just set the contentContainer here? Also note that `getPermissions` may return string classes instead of permission instances. In order to stay compatible current module behavior in which contentcontainer is optional?

buddh4 (changes requested):

protected/humhub/libs/BasePermission.php
```diff
@@ -142,15 +143,65 @@ public function getModuleId()
...
+
+    /**
+     * Returns the default state stored in DB per container type.
+     * This method returns null in case the default state for this permission or group is not stored in DB yet.
+     *
+     * @param int $groupId
+     * @return int|null
+     * @since 1.8
+     */
+    protected function getDefaultStoredState($groupId)
```
buddh4, line 171: Currently the BasePermission does not have any query logic and the permissionmanager is responsible for querying, this seperation is broken by this PR. I think the permissionManager should add this default state check instead.

buddh4 (changes requested):

protected/humhub/libs/BasePermission.php
buddh4: Also note that `BasePermssion` is currently used for container as well as group permission and `BasePermission::getDefaultStoredState()` implements concrete logic for container permission default state. But group level permissions also have a default state. So this logic should rather be part of `ContentContainerPermissionManager` which is responsbile for content container permission specific logic.

protected/humhub/modules/space/models/Space.php
```diff
@@ -605,7 +605,7 @@ public function getOwnerUser()
      *
      * @return array user groups
      */
-    public function getUserGroups()
+    public static function getUserGroups()
```
buddh4, line 608: We need to be careful with such changes:

grafik

protected/humhub/modules/content/components/ContentContainerDefaultPermissionManager.php
yurabakhtin: Ah, yes you are absolutely right, if we will change here the code `return $module->getPermissions(new $this->contentcontainer_class);` to:

```
$contentContainer = new $this->contentcontainer_class;
$permissions = $module->getPermissions($contentContainer);
foreach ($permissions as $permission) {
    if ($permission->contentContainer === null) {
        $permission->contentContainer = $contentContainer;
    }
}
return $permissions;
```

then we don't my migrations fixes in external modules.

Thank you for advice.
Sorry for the bad solution and what I did in all external modules.
Should I apply this new code and then revert all related changes in another modules?

buddh4: No problem, wasn't neccesarrily a bad solution. The purpose of `BasePermission::$contentContainer` is not clear on first sight and this is a reason we should get rid of it as mentioned in https://github.com/humhub/humhub/issues/4670#issuecomment-737865080

buddh4: If its not too much effort we should rather remove the migration, since the field may get deprecated soon.

buddh4:
@yurabakhtin reading the code I do not fully understand how the container is set when caling `$space->permissionManager->can(MyPermission::class)` could you show me the line where this is handled in the actual permission verification process.

yurabakhtin:
@buddh4 Do you mean when we call `$space->permissionManager->can(MyPermission::class)` there we create the permission object without params because there is a code like `Yii::createObject($permission)` without params like `Yii::createObject($permission, ['contentContainer' => $this])`?

protected/humhub/libs/BasePermission.php
yurabakhtin: ok I will review the code to move the query logic from `BasePermission` to `ContentContainerDefaultPermissionManager`.

protected/humhub/modules/space/models/Space.php
yurabakhtin: Do you mean we should replace all code like `$space->getUserGroups()` with `Space::getUserGroups()`?

luke-: Probably it makes sense not to make the method static. That way we can support dynamic user groups later. 

What do you think? @buddh4 @yurabakhtin

luke-: In this case we need an additional static getDefaultUserGroups for the default permissions anyway.

buddh4:
@yurabakhtin I assume when calling `$space->permissionManager->can(MyPermission::class)` the permissionManager somehow needs to set the `BasePermission::$contentContainer` in order to fetch the right default permission for this container `ContentContainerPermissionManager::$contentContainer`. I did not find where this happens in code or did, maybe I did oversee something.

protected/humhub/modules/space/models/Space.php
buddh4: I think its ok to make it static but we should align the other calls. Its possible that non static calls to static functions will be deprecated in a futre PHP version also its just not the right syntax and IDE's and code analysis tools probably will complain.

One minor concern is that a static function can not be part of a possible interface unifying ContentContainerActiveRecords (Space/user) in the future.

yurabakhtin:
@buddh4 I retested again and yes you are right the new default space/user permission values are not used in real verification, i.e. currently they are used only in the selector on edit permissions of space/user containers.
So this PR is not completely.
We need somehow to pass `ContentContainer` when call `$space->permissionManager->can(MyPermission::class)`, i.e. correct code must be like `$space->permissionManager->can(new MyPermission($space))`, but of course we don't need to change this in each such place, it must be implemented through `PermissionManager`.
I am working to implement your suggested solution "moving the logic to ContentContainerPermissionManager anyways"...

protected/humhub/libs/BasePermission.php
yurabakhtin: done in commit https://github.com/humhub/humhub/pull/4678/commits/0d6ad4c2f3241d020ff017650e8d802c5ca74def

yurabakhtin:
@buddh4 I have implemented the moving query logic from `BasePermission` into `ContentContainerPermissionManager` in commit https://github.com/humhub/humhub/pull/4678/commits/0d6ad4c2f3241d020ff017650e8d802c5ca74def, and this solution solved the wrong verifying by new default permissions.
Also after this fix/moving we don't need anymore the passing `['contentContainer' => $contentContainer]` in each permission, i.e. we can revert all those changes in all modules where I added the `['contentContainer' => $contentContainer]` to each permission in `Module->getPermissions($contentContainer = null)`.

protected/humhub/modules/content/components/ContentContainerDefaultPermissionManager.php
yurabakhtin: after implementation described in https://github.com/humhub/humhub/pull/4678#issuecomment-738632155 we may revert the initialization of `contentContainer` on each permission, because currently all query logic was moved into Permission Manager (`ContentContainerPermissionManager`) that already has an initialized Content Container.

protected/humhub/modules/space/models/Space.php
yurabakhtin: I was not sure how we can convert methods to different type, because I don't know how often this method may be used in external modules, so I was asking here https://github.com/humhub/humhub/issues/4670#issuecomment-735766589 and @luke- answered "GetUserGroups: We can mark them as static.".
Of course it was my not enough clean converting in all places, so I would agree with @buddh4 because we really cannot be sure how it may be deprecated in future PHP versions.

But i left the codes without modifications because currently PHP works well with such code, and my IDE PhpStorm 2020.3 doesn't show any warnings(suggestions to some replacing) even with "PHP language level = 8.0".

It would be good if someone of both you tell me exactly I should convert all rest or keep it as is now, thanks.

protected/humhub/modules/comment/Module.php
```diff
@@ -37,7 +37,7 @@ public function getPermissions($contentContainer = null)
     {
         if ($contentContainer instanceof Space) {
```
ArchBlood, line 38: In light of #4738 it would be better to remove `instanceof Space` and open the permissions up to also profile comments as well as Space, your thoughts @yurabakhtin?

yurabakhtin: Yes, it is strange why the permission was developed only for Space, because I see we can create comments in User container as well, so no reason to keep this only for Space. But I didn't do this so I cannot change this right now, maybe @luke- or @buddh4 know this code better.

luke-: @ArchBlood thank you for your PR here!

protected/humhub/modules/content/models/ContentContainerDefaultPermission.php
```diff
@@ -0,0 +1,61 @@
...
+            [['permission_id'], 'string', 'max' => 150],
+            [['contentcontainer_class', 'class'], 'string', 'max' => 255],
+            [['group_id', 'module_id'], 'string', 'max' => 50],
+        ];
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function attributeLabels()
```
luke-, line 49: It should be ok to just delete this method?

protected/humhub/modules/space/models/Space.php
luke-: Since @buddh4 also has no problem with us marking the method as `static`, I would leave it as it is.

Also, in the examples in the PHP documentation there are also static calls to a variable: 
https://www.php.net/manual/en/language.oop5.static.php

But let's just change the few calls to Space::getUserGroups, especially where it is possible.

luke-:
@yurabakhtin Sorry for the review delay. I was able to test it now and from my side everything works as expected. 

However, it would be good to revert the permission changes (with the ContentContainer). Can you please create PRs or even direct commits to the modules (if possible).

protected/humhub/modules/content/models/ContentContainerDefaultPermission.php
yurabakhtin: yes, I don't find where it is used, there is no separate page with edit a single permission, so the labels are not used. I have removed the method and same for the class `ContentContainerPermission`.
Commit https://github.com/humhub/humhub/pull/4678/commits/ff112847b8b9cc735389b8da5759be619598fcd7.

yurabakhtin:
@luke- 

> it would be good to revert the permission changes (with the ContentContainer).

Commit https://github.com/humhub/humhub/pull/4678/commits/a2d52b6e8727f63f51533a25baf4f71731ff3dd4 for core.

yurabakhtin:
@luke- 

> Can you please create PRs or even direct commits to the modules (if possible).

- cfiles: direct commit https://github.com/humhub/humhub-modules-cfiles/commit/b3de0d11817609b2f2622215dc8d44ea76f7e96d
- mail: direct commit https://github.com/humhub/humhub-modules-mail/commit/e1824d2e66a4ec489a3551965b7d6a722cfacacc
- polls: direct commit https://github.com/humhub/humhub-modules-polls/commit/bc96f5762817d8a05f53f3d8fca3431922a35d80
- external-calendar: the PR was not merged
- custom-pages: direct commit https://github.com/humhub/humhub-modules-custom-pages/commit/f9a0609071852d03825f45351f866938b9f240e4
- news: direct commit https://github.com/humhub/humhub-modules-news/commit/af7bb04db13632a26cd48f7a3da4367035936aac

PR #4708, opened 2020-12-11
Doc for AdminMenu->addItem()

yurabakhtin:
**What kind of change does this PR introduce?** (check at least one)

- [x] Doc for a deprecated method

**Does this PR introduce a breaking change?** (check one)

- [x] No

If yes, please describe the impact and migration path for existing applications:

**The PR fulfills these requirements:**

- [x] It's submitted to the `develop` branch, _not_ the `master` branch if no hotfix

**Other information:**

Issue: https://github.com/humhub/humhub/issues/4695#issuecomment-743097525

PR #4709, opened 2020-12-14
Removed deprecated addItem from AdminMenu

luke-:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [X] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [X] Yes
- [ ] No

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**

### DIFF
--- a/protected/humhub/modules/ui/menu/MenuEntry.php
+++ b/protected/humhub/modules/ui/menu/MenuEntry.php
@@ -39,9 +39,9 @@ abstract class MenuEntry extends BaseObject
     protected $htmlOptions = [];
 
     /**
-     * @var bool
+     * @var bool|null
      */
-    protected $isVisible = true;
+    protected $isVisible = null;
 
     /**
      * @var bool mark this entry as active
@@ -69,8 +69,9 @@ abstract class MenuEntry extends BaseObject
      * @param array $extraHtmlOptions
      * @return string
      */
-    public function render($extraHtmlOptions = []) {
-        if(!$this->isVisible) {
+    public function render($extraHtmlOptions = [])
+    {
+        if (!$this->isVisible()) {
             return '';
         }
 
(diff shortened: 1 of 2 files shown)


luke-:
Migration Notes: https://github.com/humhub/documentation/blob/master/docs/develop/modules-migrate.md#migrate-from-17-to-18

buddh4:
In https://github.com/humhub/humhub/pull/4708?notification_referrer_id=MDE4Ok5vdGlmaWNhdGlvblRocmVhZDEzODkzMDU3NDI6OTgwMDg1NA%3D%3D&notifications_query=reason%3Amention you mentioned overwriting addEntry/addItem in order to set a admin check in case no manual check was added to an entry. I think this is a good idea.

luke-:
@buddh4 This is unfortunately not possible. In the MenuEntry class "isVisible" is not null by default. So we can't detect this state.

buddh4:
Maybe we could set it to null by default and set it to true somewhere else prior to render it?

luke-:
@buddh4 Should work well. See here the commit: https://github.com/humhub/humhub/pull/4709/commits/c074fdce266a2cd9a08e849e118bd5b6b4a49d26

protected/humhub/modules/ui/menu/MenuEntry.php
```diff
@@ -230,6 +231,18 @@ public function setIsVisible($isVisible)
...
     }
 
+    /**
+     * Returns the value of the isVisible property. If the value is not set, null is returned.
+     * To query visibility, use the method `isVisible()`.
+     *
+     * @return bool|null
+     * @since 1.8
+     */
+    public function getIsVisible()
```
buddh4, line 241: I'd prefere something like `isVisibilitySet()` or `hasVisibilityState()` and document that this may be used in menues to be able to set a default value. I think there is no other use case for directly determining the value of `isVisible` since we only have three states `uninitialized`, `true`, `false`. But its your call in case you prefere the current solution.

luke-: Agreed, see commit: 06bacae

Issue #4713, opened 2020-12-15
Content canView should check membership status and container visibility for owner of content


fae7em:
### What steps will reproduce the problem?
A member creates private content in space.
Member cancels his/her membership in that space.

### What is the expected result?
Since the content is private and this user is not a member anymore, the Content `canView` should return `false`.

### What do you get instead?
The Content `canView` just check ownership of content and return `true`.

### Additional info

| Q                | A
| ---------------- | ---
| HumHub version   | 1.7.1

fae7em:
Actually, I'm not sure about the right scenario. I think, if a content is in a private container, it should be just displayed to the container members, and container is the owner of that content, not the user who create that content.

Currently, if a user create a content in a private space, even if this user cancel his/her membership in that space, he/she can access to that content in her profile, with access to delete the content, like post and its comments, add/edit/delete comments, see any new comment on that post. This user is not a member now, but can have activity in the space through her posts.

buddh4:
Private space content should only be visible for space members, therefore there should not be public content in private spaces. We should force private content visibility for private spaces. Regarding my own content of a private space I'am not a member anymore, I think this is a very specific case and the user should still be able to see the content, otherwise I assume the visibility checks (which kind of are spread in different places unfortunately) will get too complex. Currently the rule is that users are always allowed to see his/her own content. In the future maybe we could add a way of change the ownership of such content. What do you think @luke-

fae7em:
>  We should force private content visibility for private spaces in `beforeSave` (if its not the case already). 

No, it's not already, if space visibility change to private, its content visibility stay public, and as I mentioned in #4712, in the content `canView` function, it just checks if the content is public (and not its container) and return true.

buddh4:
So we'd need to switch the visibility of all contents when changing the space visibility. Also we should render a warning that all content visibilities will be changed.

luke-:
Yes, public content should not exist in a private Space. So we should switch existing public content (including its activities) to private when the Space setting is changed. (incl. a Warning of course). This process should ideally be executed as an ActiveJob. 
@fae7em  Would you like to start a new issue & PR for this?   

Because of the second issue, I also think that content should belong to the ContentContainer.  As an example, from a Wiki page that grows, the original creator should no longer have access if they are no longer part of the Space.  I think making the change everywhere consistently should not be a big problem.  However, we should then consider all places:Streams, Search and ContentActiveQuery here.

buddh4:
[...]

Issue #4632, opened 2020-11-13
mentioned non-member users get notification in private content


fae7em:
### What steps will reproduce the problem?
In private content of public space, if we mention a non-member user in a post or comment, that user doesn't receive any notification of it.
But after that, the user will receive notification for every comment on that post (which the user is mentioned in it or in its comments)

### What is the expected result?
If the non-member user is mentioned in a private space, he/she should receive notification about it, if the scenario is to ignore notification for non-members, then the user should not receive other notifications of that posts.

### Additional info

| Q                | A
| ---------------- | ---
| HumHub version   | 1.7
| PHP version      | 7.4

buddh4:
I think there is an issue related to this. Currently users only get mentioned if the user can access the content. I think we should  discuss this further prior to change this behavior. In which case it makes sense that a users is informed about a mentioning in a content he can not access?

buddh4:
Related: 

- https://github.com/humhub/humhub/issues/3767 I did not find the other discussion
- https://github.com/humhub/humhub/issues/3701 (in which the wished behavior was marked as bug)
- https://github.com/humhub/humhub/issues/2791

fae7em:
Actually, it's not about the decision of informing a non-member mentioned user or not, it's about that it doesn't have the same behavior in the current scenario.
I just tried some cases and in each case, there is a different behavior: 

Imagine there is a public space and we have users: `x`, `y`,`z`, `w` (which user `w` is not a space member)
In each case, user `x` create a private post in the public space.

-x create a post, x mention w in its comment (no one gets notification), x reply that comment (**w get notification of new comment now while w is not a member, and w didn't get notif for mentioning in parent comment**)
-x create a post, x mention w in its comment (no one gets notification), y reply that comment (x get notification)
-x create a post, y mention w in its comment (x get notification), x reply that comment (**w get a notification while it's not a member, y didn't get notification although is the owner of the parent comment**)
-x create a post, y mention w in its comment (x get notification), y reply that comment (x get a notification)
-x create a post, y mention w in its comment (x get notification), z reply that comment (**just x get a notification, not also y that is the owner of the parent comment**)
-x create a post, y mention w in its comment (x get notification), x create **a new** comment (**w get a notification while it's a totally new comment, not reply to that comment, and also w is not a member**)

buddh4:
[...]

PR #4719, opened 2020-12-16
Fix #4712 Force content to be private in private space

### LINKED ISSUE #4712: Content canView should check contentContainer visibility

Create public content in a public space and mention a non-member in it or suppose a non-member comment on that post. (In both, non-member now follow this content)
Change space visibility to private. (Content visibility doesn't change.)

Since space is private now, the Content `canView` should return false for non-members for all kinds of content visibility.

### What do you get instead?
`canView` just check the content visibility and not its container visibility and return true for non-members.

### Additional info

| Q                | A
| ---------------- | ---
| HumHub version   | 1.7.1

fae7em:
Related to issue #4712 

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [] Yes
- [x] No

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
Need to run migration

### DIFF
--- a/protected/humhub/modules/content/migrations/m201216_173119_fix_public_content_private_space.php
+++ b/protected/humhub/modules/content/migrations/m201216_173119_fix_public_content_private_space.php
@@ -1,0 +1,50 @@
+<?php
+
+use humhub\modules\content\models\Content;
+use humhub\modules\space\models\Space;
+use yii\db\Migration;
+
+/**
+ * Class m201216_173118_fix_public_content_private_space
+ */
+class m201216_173119_fix_public_content_private_space extends Migration
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function safeUp()
+    {
+        $query = Content::find();
+        $query->innerJoin('space', 'space.contentcontainer_id = content.contentcontainer_id');
+        $query->andWhere('space.visibility = :visibility', [':visibility' => Space::VISIBILITY_NONE]);
+
+        foreach ($query->all() as $contact) {
+            $contact->updateAttributes(['visibility' => Content::VISIBILITY_PRIVATE]);
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function safeDown()
+    {
+        echo "m201216_173118_fix_public_content_private_space cannot be reverted.\n";
+
+        return true;
+    }
+
+    /*
+    // Use up()/down() to run migration code without a transaction.
+    public function up()
+    {
+
+    }
+
+    public function down()
+    {
+        echo "m201216_173118_fix_public_content_private_space cannot be reverted.\n";
+
+        return false;
+    }
+    */
+}
--- a/protected/humhub/modules/space/models/Space.php
+++ b/protected/humhub/modules/space/models/Space.php
@@ -261,6 +261,12 @@ class Space extends ContentContainerActiveRecord implements Searchable
             $activity->create();
         }
 
+        if ($this->visibility != $changedAttributes['visibility']) {
+            if ($this->visibility == Space::VISIBILITY_NONE) {
+                Content::updateAll(['visibility' => Content::VISIBILITY_PRIVATE], ['contentcontainer_id' => $this->contentcontainer_id]);
+            }
+        }
+
         Yii::$app->cache->delete('userSpaces_' . $user->id);
     }
 
--- a/protected/humhub/modules/space/modules/manage/views/security/index.php
+++ b/protected/humhub/modules/space/modules/manage/views/security/index.php
@@ -24,7 +24,7 @@ use humhub\libs\Html;
 
         <?php
         $visibilities = [
-            Space::VISIBILITY_NONE => Yii::t('SpaceModule.base', 'Private (Invisible)'),
+            Space::VISIBILITY_NONE => Yii::t('SpaceModule.base', 'Private (Invisible, All Contents would be private)'),
             Space::VISIBILITY_REGISTERED_ONLY => Yii::t('SpaceModule.base', 'Public (Registered users only)')
         ];
         if (AuthHelper::isGuestAccessEnabled()) {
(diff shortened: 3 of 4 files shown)


luke-:
I think we should definitely show a confirm dialogue here, especially if we also remove all followers. 

I would prefer to trigger the ConfirmDialog (humhub.ui.modal.confirm) when the drop-down value is changed to "Private".
If that is too complex, we could also offer a button to change the visibility.

fae7em:
I agree about the confirm dialog, but we're not going to remove the followers. In #4632, we're adding a new flag to `user_follow` that activates/deactivates the content following. In these cases, we're going to use it.

protected/humhub/modules/space/models/Space.php
```diff
@@ -261,6 +261,12 @@ public function afterSave($insert, $changedAttributes)
             $activity->create();
         }
 
+        if ($this->visibility != $changedAttributes['visibility']) {
+            if ($this->visibility == Space::VISIBILITY_NONE) {
+                Content::updateAll(['visibility' => Content::VISIBILITY_PRIVATE], ['contentcontainer_id' => $this->contentcontainer_id]);
```
buddh4, line 266: In https://github.com/humhub/humhub/issues/4713 we've discussed about maybe using an ActiveJob for this in order to be able to handle content followers etc, which may not be able to see the content anymore. I think we did not come to a final decision yet or did you think about another solution for this?

protected/humhub/modules/space/modules/manage/views/security/index.php
```diff
@@ -24,7 +24,7 @@
 
         <?php
         $visibilities = [
-            Space::VISIBILITY_NONE => Yii::t('SpaceModule.base', 'Private (Invisible)'),
+            Space::VISIBILITY_NONE => Yii::t('SpaceModule.base', 'Private (Invisible, All Contents would be private)'),
```
buddh4, line 27: I think `All content becomes private` is better.

protected/humhub/modules/content/migrations/m201216_173119_fix_public_content_private_space.php
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function safeUp()
+    {
+        $query = Content::find();
+        $query->innerJoin('space', 'space.contentcontainer_id = content.contentcontainer_id');
+        $query->andWhere('space.visibility = :visibility', [':visibility' => Space::VISIBILITY_NONE]);
+
+        foreach ($query->all() as $contact) {
```
buddh4, line 21: Not a big deal but why `$contact` ? :)

buddh4 (changes requested):

buddh4:
Are there maybe other use cases in which we need to know about content visibility changes? Maybe e.g. the bookmark module needs to clean content relations if a bookmarked content is not visible anymore.

protected/humhub/modules/space/modules/manage/views/security/index.php
luke-: We should remove that and put the notice in the Confirm dialogue.

protected/humhub/modules/content/migrations/m201216_173119_fix_public_content_private_space.php
fae7em: Ahh, my mistake :D

fae7em:
> I would prefer to trigger the ConfirmDialog (humhub.ui.modal.confirm) 

Could you please guide me on how to use `humhub.ui.modal.confirm`?

buddh4:
You can either just add a `data-action-change="ui.modal.confirm"` to your select  with following data properties to change texts:

```
  'body': $trigger.data('action-confirm'),
  'header': $trigger.data('action-confirm-header'),
  'confirmText': $trigger.data('action-confirm-text'),
  'cancelText': $trigger.data('action-cancel-text')
```

The only issue I see here is that "Cancel" and "Confirm" will have the same effect. 

You can implement a custom action (did not test):

View:
```
<select data-action-change="space.settings.confirmVisibilityChange">
```

Script:
```
humhub.module('space.settings', function(module, require, $) {
let modal = require('ui.modal'); 
module.export({
    confirmVisibilityChange: function(evt) {
        modal.confirm().then(function(confirmed) {
            if(!confirmed) {
                 // reset select
            }
       });
    }
 });
});
```

protected/humhub/assets/CoreBundleAsset.php
```diff
@@ -65,7 +66,8 @@ class CoreBundleAsset extends WebStaticAssetBundle
         LikeAsset::class,
         StreamAsset::class,
         ActivityAsset::class,
-        SpaceChooserAsset::class
+        SpaceChooserAsset::class,
+        SpaceSettingAsset::class,
```
buddh4, line 70: We should not add this to the CoreBundleAsset, the CoreBundleAsset should only contain bundles which are commonly and initially required at first page load (e.g. dashboard) or containing modules which are requried very often.

protected/humhub/modules/space/resources/js/humhub.space.setting.js
```diff
@@ -0,0 +1,22 @@
+
+humhub.module('space.settings', function(module, require, $) {
+    var modal = require('ui.modal');
+
+    var confirmVisibilityChange = function (evt) {
+        if (evt.$trigger.val() === '0') {
+            modal.confirm().then(function(confirmed) {
```
buddh4, line 7: We should use some specfic texts here: https://docs.humhub.org/docs/develop/javascript-modals/#modal-confirm

buddh4: You can define such texts in your AssetBundle as follows:

```
 public static function register($view)
    {
        $view->registerJsConfig('calendar.Calendar', [
            'confirmText' => [
                'header' => Yii::t(''SpaceModule.base'', 'Confirm space visibility change'),
                 'body' => Yii::t(''SpaceModule.base'', 'All content of this space will be changed to private group visibility!'),
            ]
        ]);
        return parent::register($view);
    }
```

In your module
```
modal.confirm(module.config.confirmText).then(...
```

buddh4 (changes requested):
Sorry for the delay, I've found some minor issues, otherwise looks good!

PR #4725, opened 2020-12-17
Fix #4715: Call to `Content::canArchive()` throws error on global content

### LINKED ISSUE #4715: Content::canArchive() broken on global content
Calling `$content->canArchive()` on a global content fails, duo to:

```
public function canArchive()
{
   return $this->getContainer()->permissionManager->can(new ManageContent());
}
```

This for example happens when generating archive links for global custom page wall entries.

buddh4:
Fixes: https://github.com/humhub/humhub/issues/4715

- [x] Bugfix

- [x] No


protected/humhub/modules/content/models/Content.php
```diff
@@ -409,6 +409,16 @@ public function isArchived()
      */
     public function canArchive()
     {
+        // Currently global content can not be archived
+        if(!$this->contentcontainer_id) {
+            return false;
+        }
+
+        // No need to archive content on an archived container, content is marked as archived already
+        if($this->content->content->getContainer()->isArchived()) {
```
luke-, line 418: @buddh4 Can you please check your codestyle settings in the IDE? Actually, there should be a whitespace after the `if`.

luke- (changes requested):

protected/humhub/modules/content/models/Content.php
buddh4: Not an issue with the ide, just did not auto format. Not really used to the space after such blocks.

luke- (approved):

Issue #4586, opened 2020-10-29
Rebuild group default Space Mapping 


luke-:
### Is your feature request related to a problem? Please describe.

Currently only the default spaces for new group members are considered.
Therefore an option (checkbox or button) should be introduced to reassign the group members to all default spaces.

PR #4732, opened 2020-12-18
Hot fix for -- Rebuild group default Space Mapping #4586

MasterHans:
Changed m201025_095247_spaces_of_users_group.php migration
Deleted foreign key fk_group-space_id.
migrate-succesfully

### DIFF
--- a/protected/humhub/modules/admin/controllers/GroupController.php
+++ b/protected/humhub/modules/admin/controllers/GroupController.php
@@ -101,7 +101,7 @@ class GroupController extends Controller
                     }
 
                     $this->view->info(Yii::t('AdminModule.user',
-                        'The Space memberships of all group members will be updated. This may take a several minutes.'));
+                        'The Space memberships of all group members will be updated. This may take up to several minutes.'));
                 }
             }
         }
--- a/protected/humhub/modules/space/models/Space.php
+++ b/protected/humhub/modules/space/models/Space.php
@@ -323,12 +323,7 @@ class Space extends ContentContainerActiveRecord implements Searchable
         }
 
         Invite::deleteAll(['space_invite_id' => $this->id]);
-
-        // Delete the corresponding entries of the table group_space
-        if (!empty($this->id)){
-            $groupSpace = GroupSpace::findOne(['space_id'=>$this->id]);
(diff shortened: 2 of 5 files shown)


buddh4:
Now we get:
grafik

buddh4:
@luke- do you think we need to align other parts as for example LDAP or MassUserImport.

luke-:
@MasterHans Please see my update/commit. I think a TryCatch is easier here. Also, the FK should be automatically deleted by the databaseg when the field is deleted.

@buddh4 Good point, but should not affect the modules here. The modules do not change a group,  but only add members.

MasterHans:
@buddh4  Something strange happened here can you please help to understand


test-error



error message in test

1) GroupTest: Single default spaces
Test ../../user/tests/codeception/unit/models/GroupTest.php:testSingleDefaultSpaces
Failed asserting that actual size 0 matches expected size 1.
#1 /home/runner/work/humhub/humhub/protected/humhub/modules/user/tests/codeception/unit/models/GroupTest.php:63

System try to run this test method
../../user/tests/codeception/unit/models/GroupTest.php:testSingleDefaultSpaces

But my branch ref_4586 does not contain this method - testSingleDefaultSpaces.

Looks like that in some how tests starting to run from the develop branch. And this error occurs.
I can not debug this error ...
What should I do ?
Add ../../user/tests/codeception/unit/models/GroupTest.php:testSingleDefaultSpaces to ref_4586 from develop branch?

MasterHans:
> do you think we need to align other parts as for example LDAP or MassUserImport.

I think needed to find all mentioning of space_id and fix them....

MasterHans:
> @buddh4 Good point, but should not affect the modules here. The modules do not change a group, but only add members.

Should I correct this module as well?

MasterHans:
> @MasterHans Please see my update/commit. I think a TryCatch is easier here. Also, the FK should be automatically deleted by the databaseg when the field is deleted.

Ok I'll do it thanks

MasterHans:
> Please see my update/commit. I think a TryCatch is easier here. Also, the FK should be automatically deleted by the databaseg when the field is deleted.

Done!

PR #4738, opened 2020-12-25
Enh: canComment permissions

ArchBlood:

- [x] Feature
- [x] Refactor

- [x] Yes
- [ ] No

With this p/r it allows the User to also allow/deny commenting on their post(s)

**Other information:**
https://community.humhub.com/content/perma?id=232539
Screenshot_1

### DIFF
--- a/protected/humhub/modules/comment/Module.php
+++ b/protected/humhub/modules/comment/Module.php
@@ -7,6 +7,7 @@ use humhub\modules\comment\permissions\CreateComment;
 use humhub\modules\comment\notifications\NewComment;
 use humhub\modules\content\components\ContentActiveRecord;
 use humhub\modules\space\models\Space;
+use humhub\modules\user\models\User;
 use Yii;
 
 /**
@@ -35,7 +36,7 @@ class Module extends \humhub\components\Module
      */
     public function getPermissions($contentContainer = null)
     {
-        if ($contentContainer instanceof Space) {
+        if ($contentContainer) {
             return [
                 new permissions\CreateComment()
             ];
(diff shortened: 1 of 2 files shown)


protected/humhub/modules/comment/Module.php
```diff
@@ -83,7 +84,7 @@ public function canComment($object)
 
         $content = $object->content;
 
-        if ($content->container instanceof Space) {
+        if ($content->container) {
```
ArchBlood, line 87: @luke- would it be smarter to change this to the following or is this enough?

```php
        if($content->container) {
            /** @var User $user */
            $user = $content->container;
            if (!$user->permissionManager->can(CreateComment::class)) {
                return false;
            }
        }

        if ($content->container) {
            /** @var Space $space */
            $space = $content->container;
            if (!$space->permissionManager->can(CreateComment::class)) {
                return false;
            }
        }
```

protected/humhub/modules/comment/Module.php
```diff
@@ -83,7 +84,7 @@ public function canComment($object)
...
             $space = $content->container;
             if (!$space->permissionManager->can(CreateComment::class)) {
```
luke-, line 89: I would just change the line here:
```php
 if (!$content->container->permissionManager->can(CreateComment::class)) {
 ```

ArchBlood: Note that if `intsanceof Space` isn't removed then the permissions don't set right for User permissions on their posts for profile stream, this is why it's still removed in this p/r

luke- (approved):

luke- (approved):

Issue #4743, opened 2020-12-28
LikeLink also when no CanLike permission  


luke-:
### What steps will reproduce the problem?

Set `CanLike` permission to `Deny` on Space or Profile level. 

### What is the expected result?

No Like link or some deny indicator.

### What do you get instead?

Not clickable like link.

### Additional info

| Q                | A
| ---------------- | ---
| HumHub version   | master/develop

Should be fixed in `master` branch.

yurabakhtin:
@luke- I have fixed this in PR https://github.com/humhub/humhub/pull/4745 for `develop` branch. Yes, you wrote "Should be fixed in `master` branch." but is it normal to do fix only in `master` branch?
I am a bit confused and need more clear instructions, how to do fixes right into `master` because I think if we do such fixes only in `master` then they may be overridden on release of new version from `develop` branch, or am I wrong?

luke-:
@yurabakhtin We regularly merge all commits from the `master` branch into the `develop` branch. This means that the `develop` branch also receives this PR, even it's not the base branch.

yurabakhtin:
@luke- Ah, then I get it, because before I thought you do otherwise, ok I will fix the PR to `master` base.

luke-:
Thanks!

PR #4747, opened 2020-12-29
Hide "Like" link when no permission "Can like"

yurabakhtin:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
Issue: https://github.com/humhub/humhub/issues/4743

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ HumHub Changelog
 - Fix #4649: Success message rendered although password validation failed
 - Fix #4717: Repsonsive layout alignment issue on small screens
 - Fix #4715: Call to `Content::canArchive()` throws error on global content
+- Fix #4743: Hide "Like" link when no permission "Can like"
 
 
 1.7.1 (November 27, 2020)
--- a/protected/humhub/modules/like/Events.php
+++ b/protected/humhub/modules/like/Events.php
@@ -10,7 +10,6 @@ namespace humhub\modules\like;
 
 use humhub\components\ActiveRecord;
 use humhub\components\Event;
-use humhub\modules\content\components\ContentContainerActiveRecord;
 use humhub\modules\like\models\Like;
 use Yii;
 
@@ -93,16 +92,6 @@ class Events extends \yii\base\BaseObject
      */
     public static function onWallEntryLinksInit($event)
     {
-        if (!static::getModule()->isEnabled) {
-            return;
-        }
-
-        if (!isset($event->sender->object->content->container) ||
-            !($event->sender->object->content->container instanceof ContentContainerActiveRecord) ||
-            !$event->sender->object->content->container->can(new permissions\CanLike())) {
-            return;
-        }
-
         $event->sender->addWidget(widgets\LikeLink::class, ['object' => $event->sender->object], ['sortOrder' => 10]);
     }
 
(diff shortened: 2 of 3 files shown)


luke- (changes requested):
A Like button still appears in my comments and sub-comments. Can you please check this again?

yurabakhtin:
@luke- You wrote here https://github.com/humhub/humhub/pull/4747:

> A Like button still appears in my comments and sub-comments. Can you please check this again?

I have fixed this in commit https://github.com/humhub/humhub/pull/4747/commits/09b32c9270e315092fc1e5f77ba74de55ae159cc

The button was not hidden there because it is printed by separate code like `<?= LikeLink::widget(...) ?>` instead of event call, and this code is used in many other external modules. But please note in such cases the `LikeLink` often is used together with other links like `CommentLink`, e.g.:

```php
<?= LikeLink::widget(['object' => $gallery]) ?>
|
<?= CommentLink::widget(['object' => $gallery]) ?>
```

so if some link is hidden because of no permission then the separator `|` or `&middot;` is visible, so it looks not good I think.
If we need to display the separator  properly then it would be good to implement some new function like:

```php
<?= Widget::widgets([LikeLink::class, CommentLink::class], ['object' => $gallery, 'separator' => '|'])?>
```

Maybe a similar function already exists and if you know please tell me and let me know if we should fix the separator visibility in all modules.

luke-:
We have something like this:

```php 
<?= \humhub\widgets\BaseStack::widget([
    'widgets' => [
        [
            CommentLink::class, ['object' => $comment], ['sortOrder' => 100],
            LikeLink::class, ['object' => $comment], ['sortOrder' => 200]
        ]
    ]
]);
```

But instead of using the BaseStack class directly, it would be good to create a new one like CommentEntryControls.

Hmm, since the PR is already a bit bigger as expected and the bug is not that critical, I would almost suggest to put it in the `Develop` branch. What do you think?

If we go the BaseStack widget way, we also have to move the canComment to the CommentLink widget here.

```php
//comment.php#90
        <div class="wall-entry-controls">
            <?php if ($module->canComment($comment)): ?>
                <?= CommentLink::widget(['object' => $comment]); ?>&nbsp;&nbsp;&middot;&nbsp;
            <?php endif; ?>
            <?= LikeLink::widget(['object' => $comment]); ?>
        </div>
```

yurabakhtin:
@luke- 

> But instead of using the BaseStack class directly, it would be good to create a new one like CommentEntryControls.

I have reviewed all places and yes the `LikeLink::widget(...)` is used together with `CommentLink::widget()`, so yes we can new widget with name like `CommentEntryControls`, but please note the link widgets are used for different modules "wiki", "cfiles" and etc.

> Hmm, since the PR is already a bit bigger as expected and the bug is not that critical, I would almost suggest to put it in the Develop branch. What do you think?

I think the done fixes can be merged into `master` because they solve the visibility issue, but next the not urgent fix we can do in `develop` because it will just hide a separator between `LikeLink` and `CommentLink`.

> If we go the BaseStack widget way, we also have to move the canComment to the CommentLink widget here.

I see the `CommentLink` already has the restriction `canComment` inside.

luke-:
Ok great! 

So you're going to do a separate PR for the CommentEntryControls?

yurabakhtin:
@luke- Yes, we can do this in a separate PR for `develop`, I created the issue here: https://github.com/humhub/humhub/issues/4751.

luke- (approved):
OK, then I'll merge this PR into the Master Branch.

Issue #4752, opened 2020-12-30
Download yields outdated file due to browser caching


dreua:
I reported this against the onlyoffice plugin and was redirected: https://github.com/ONLYOFFICE/onlyoffice-humhub/issues/28

If a file is modified in place (i.e. the download URL does not change), you will not get the updated file on subsequent downloads due to browser caching.

### What steps will reproduce the problem?
See https://github.com/ONLYOFFICE/onlyoffice-humhub/issues/28

### Additional information
Humhub serves the file download with this header: `Cache-Control: public, max-age=3600`

### Workaround / Fix?
```diff
--- a/DownloadAction.php_bkup
+++ b/DownloadAction.php
@@ -73,7 +73,8 @@ class DownloadAction extends Action
             return true;
         }
 
-        $httpCache = new HttpCache();
+       $httpCache = new HttpCache();
+       $httpCache->cacheControlHeader = 'max-age=0, must-revalidate'; // can't rely on cache when in place editing of files can occur
         $httpCache->lastModified = function() {
             return Yii::$app->formatter->asTimestamp($this->file->updated_at);
         };

```

| Q                | A
| ---------------- | ---
| HumHub version   | 1.7.1
| PHP version      |  7.2.24
| Operating system | CentOS 8

ArchBlood:
Not sure but something to look into when it comes to this issue

> ### Requiring revalidation
> `no-cache` and `max-age=0, must-revalidate` indicates same meaning.
Clients can cache a resource but must revalidate each time before using it. This means HTTP request occurs each time though, it can skip downloading HTTP body if the content is valid.

dreua:
> Clients can cache a resource but must revalidate each time before using it.

That is exactly the behavior I'd like to see, therefore I chose this value, what exactly are you not sure about?

The only other approach I can think of would be changing the download URL each time the file changes, e.g. by appending the last changed timestamp as a parameter. (This would not be parsed on the serving side, since it only needs to indicate to the browser and possible proxy caches that this is a different file.)

(Third option: Declare file contents immutable and toss the ball back to onlyoffice plugin and possibly others.)

dreua:
History of download caching:
- introduced in https://github.com/humhub/humhub/commit/8d02f902b056196c185a604e4e2ece6d21acf0eb
[...]

PR #4769, opened 2021-01-11
File hash for download url

yurabakhtin:

- [x] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
Issue: https://github.com/humhub/humhub/issues/4752

### DIFF
--- a/CHANGELOG_DEV.md
+++ b/CHANGELOG_DEV.md
@@ -17,3 +17,4 @@
 - Fix #4718: Fix space or profile banner size after crop
 - Enh #4738: Allow CanComment permission also on profile level
 - Enh #4749: Add Migration helper to determine installation state
+- Enh #4752: Implement file column `hash` and use it in download URL
--- a/CHANGELOG_DEV.md
+++ b/CHANGELOG_DEV.md
@@ -17,4 +17,4 @@
 - Fix #4718: Fix space or profile banner size after crop
 - Enh #4738: Allow CanComment permission also on profile level
 - Enh #4749: Add Migration helper to determine installation state
-- Enh #4752: Implement file column `hash` and use it in download URL
+- Enh #4752: Added file column `hash_sha1` to avoid caching problems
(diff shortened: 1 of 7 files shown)


luke- (approved):

luke-:
@yurabakhtin FYI, i have renamed the attribute `hash` to `hash_sha1`.

buddh4:
In case a file somehow is missing in the store, the `File::getUrl()` will throw an error, since it can not create a hash. Should we maybe consider a more failsave approach and just leave out the hash in case there is no matching file in the store due to data inconsistency? I saw this issue in a test case in which I just created an empty file model, not sure how often this may occure on a production system.

yurabakhtin:
@buddh4 Could you tell us how to reproduce the error from `File->getUrl()` step by step?

> here we need to create a file manually and then add the file to the store.

Please tell me where I can create a file manually because I cannot find such feature in back-office, thanks.

---

Probably we have the error here https://github.com/humhub/humhub/pull/4769/commits/ffcdcc6d54b37a69f5719fad804c55a11c80fd71#diff-2f3a5cf58d46a6971793c97d7c4defb317661fce7c7cb141d28467442049dbf9R82

@luke- Do you think we really need it there? I think at least it should be changed to `$file->saveHash();` because the `$file` is provided as a function argument. But I think we don't need this because the same function is called in the parent `File->afterSave()`.

The second case https://github.com/humhub/humhub/pull/4769/commits/ffcdcc6d54b37a69f5719fad804c55a11c80fd71#diff-2f3a5cf58d46a6971793c97d7c4defb317661fce7c7cb141d28467442049dbf9R93 I don't find where it may be used.
Here we can add a checking like `if (isset($this->file) && ($this->file instanceof File))`.

luke-:
I think it's primarily about calling an "is_file" before the sha function.

To reproduce I would:

- Upload some files
- Set sha columns to null 
- Delete files in filesystem (uploads/files) 
- Try access files

yurabakhtin:
@luke- Thanks for the steps, I could reproduce the error and fixed in the PR https://github.com/humhub/humhub/pull/4813

buddh4:
@yurabakhtin Here are the test cases which failed: https://github.com/humhub/humhub/pull/4686/files#diff-d0c6bdb00b33daa3f05874e2c6428e3e61e88635795e2d59b13ecf194b4f1ff1R196 but should be ok with your fix. I'll run the test with your fix later.

PR #4785, opened 2021-01-20
Left menu with list of topic tags on space view

raphaeljolivet:
I have added a left menu with list of topic tags for my instance.
Here is what it looks like (in french here, but the menu label is internationalized)

image

I leave it there if you want to integrate it.
Thanks for your work BTW !

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,8 @@ HumHub Changelog
 ================
 
 
-1.7.1 (Unreleased)
-------------------
+1.7.1 (November 27, 2020)
+-------------------------
 - Fix #4612: Dashboard wall stream entry displays user name in sub title in post style layout
 - Fix #4614: Profile Image crop broken
 - Fix #4607: Changed grid size of image preview on fluid themes
@@ -17,6 +17,7 @@ HumHub Changelog
 - Fix #4660: Topic stream filter leads to stream entry duplication in combination with stream suppression
 - Fix #4638: Profile settings do not accept birthdate in russian format
 - Fix #4596: Set `autocomplete="off"` on date picker fields
+
 
 1.7.0 (November 4, 2020)
 ------------------------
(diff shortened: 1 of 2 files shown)


ArchBlood:
To be honest, I think it would be better suited as a module idea but not a bad idea, makes Topics more visible

luke-:
Thanks for your contribution.

This is really a good idea to make the topics more prominent this way.
However, as @ArchBlood  says, this would probably be something for a module or at least an additional option. 

It would probably also work well as a Space module, since only the menu needs to be added.

I mark the PR as a draft that it is not lost!

raphaeljolivet:
Yeah sure,
I just wanted to leave it somewhere.

ArchBlood:
> Thanks for your contribution.
> 
> This is really a good idea to make the topics more prominent this way.
> However, as @ArchBlood says, this would probably be something for a module or at least an additional option.
> 
> It would probably also work well as a Space module, since only the menu needs to be added.
> 
> I mark the PR as a draft that it is not lost!

Here's some ideas that I've had when reviewing this p/r

### Solution 1
- Add to`/protected/humhub/modules/space/modules/manage/widgets`

I suggested this method because if it were added it should be manageable, meaning completely configurable per space with options for space owners/admins/moderators to be able to remove Topics or at least hide unwanted Topics from view

### Solution 2
- Create a third-party module with permissions so that this is optional per user/space

Solution 2 seems to be more realistic as then it could also be extended into other areas where Topics are used such as user profiles, other modules, & etc.

felixhahnweilheim:
Hi @raphaeljolivet, I've added your topic list to my child theme (https://github.com/felixhahnweilheim/humhub-themes-orange).
Using my theme module was the easiest way for me to add the widget without manipulating core code and without writing a whole module for it.

For my usecase it doesn't have to be optional per space because for those spaces that use topics it always makes sense to show the list. Other spaces don't use topics at all and won't be affected.

ArchBlood:
This is being planned within the Trending module;

https://github.com/GreenMeteor/trending

Issue #2950, opened 2018-01-18
Modal stays open on pjax page load


buddh4:
### What steps will reproduce the problem?
1. Open a modal with pjax link
2. Click the link

### What is the expected result?
The new page should load and Modal should be closed

### What do you get instead?
Modal stays open and the page loads in the background

### Additional info

| Q                | A
| ---------------- | ---
| HumHub version   | 1.2

yurabakhtin:
@buddh4 @luke- I tested this with a link "Forgot your password?" on the modal login window:

1) Currently the link has the following code https://github.com/humhub/humhub/blob/master/protected/humhub/modules/user/views/auth/login_modal.php#L60:

```html
<a id="recoverPasswordBtn" href="#" data-action-click="ui.modal.load" data-action-url="<?= Url::to(['/user/password-recovery']) ?>">
```

On click, the page `/user/password-recovery` is loaded in the same modal window as expected.

2) If I change that to `<a href="<?= Url::to(['/user/password-recovery']) ?>">` then on click the page is loaded in background and the modal window is not hidden. I guess we should fix this problem.

3) Here https://github.com/humhub/humhub/blob/master/static/js/jquery.pjax.modified.js#L104-L107 I find the code:

```js
// Ignore links with data-pjax="0"
if ($link.data('pjax') == 0) {
    return;
}
```

3a) If add the attribute `data-pjax="0"` to the `<a href=...>` then it works as required, i.e. modal is closed and new page is opened with the URL from `href`. It may be as solution \#1.

4) solution \#2 is to modify the js code to `if ($link.data('pjax') == 0 || $link.attr('href') !== '#') {...`.
But I am not sure I can modify the file, what if some links will be broken in some modal windows. Or all correct links in modal window should have the attributes like `data-action-url` and if a link has only `href="some-ulr"` then such link should be opened as normal page then we can apply the fix.

What do you think?
Is it enough solution \#1?
Can we modify the file for solution \#2?

buddh4:
Maybe we can just close all open modals within the `unload` function of the modal js module?

yurabakhtin:
@buddh4 I don't find the `unload` function in the modal js module. But please see again here https://github.com/humhub/humhub/blob/master/static/js/jquery.pjax.modified.js#L104-L127 what ignore cases we have for the pjax click:

```js
// Ignore links with data-pjax="0"
if ($link.data('pjax') == 0) {
    return;
}

if (link.tagName.toUpperCase() !== 'A')
[...]

PR #4812, opened 2021-02-01
Close modal window on click link with url in href

yurabakhtin:

- [x] Bugfix

- [x] Changelog was modified

**Other information:**
Issue: https://github.com/humhub/humhub/issues/2950


static/js/jquery.pjax.modified.js
```diff
@@ -130,6 +126,15 @@
         if (event.isDefaultPrevented())
             return
 
+        // Close a modal window if the link with provided URL is opened from the modal window
+        // in order to display the loaded page content in background under the modal
+        if (link.href.length && link.href !== '#') {
```
buddh4, line 131: I still prefer a solution similar to the one I mentioned in https://github.com/humhub/humhub/issues/2950#issuecomment-770648113 since:

- This solution does only work on click events but pjax can also be triggered directly by javascript e.g. pjax.redirect() or pjax.reload()
- There are potentially multiple modals open at once
- The jquery.pjax.modified is a patched version of the original I'd rather prevent custom code here if possible and also don't mix in some modal specific code

buddh4 (changes requested):

static/js/jquery.pjax.modified.js
yurabakhtin: Ok, please check commit https://github.com/humhub/humhub/pull/4812/commits/091704f1b637fee09fb4ed16554103ef93435df4

yurabakhtin: But I see many tests are failed here https://github.com/humhub/humhub/runs/1807645931?check_suite_focus=true, I have to fix them.

yurabakhtin: The failed tests work without errors on my local server:

test_activity
test_space
test_stream
test_user

luke-:
@buddh4  Do you agree with the solution approach? Then we can check why the tests fail. 

I think the approach is so good. There is not the case that a modal content is loaded via PJAX.  
The modal init for each PJAX load should be ok.

static/js/humhub/humhub.ui.modal.js
```diff
@@ -542,6 +543,17 @@ humhub.module('ui.modal', function (module, require, $) {
 
     var init = function () {
         module.global = Modal.instance('#globalModal');
+
+        if (pjax.isActive()) {
```
buddh4, line 547: Use this instead as mentioned  in https://github.com/humhub/humhub/issues/2950#issuecomment-770648113:

```
module.initOnPjaxLoad = true;
var init(pjax) {
    if(pjax) {
    ...
    }
```

This means that the init function will also be called after a pjax page load. An by the pjax argument of the init function you can check if the init call was triggered after pjax page load. You don't have to check if pjax is active.

buddh4 (changes requested):

static/js/jquery.pjax.modified.js
buddh4: Maybe the acceptance tests run a bit faster in the git test environment and this maybe leads to race condition. Are those errors related to modal?

static/js/humhub/humhub.ui.modal.js
yurabakhtin: ok please review the commit https://github.com/humhub/humhub/pull/4812/commits/12d03a1522b489f340054c575345abc52b4bfc69

buddh4: Looks good, think this should be ok, pjax has a custom loader animation anyways also this case is not that common.

buddh4 (approved):

Issue #3733, opened 2019-11-12
Open profile counter in new tab


buddh4:
When opening a profile counter e.g. user spaces in a new tab, the space list modal opens in a new tab without any other layout.

luke-:
We should find a way to ensure that Ajax Actions are not called directly. There must be a header or request type that we can use to identify this.

An action filter / middleware or similar would be good here. 

If such a method is called directly, a definable URL should be called (if not specified dashboard) and a status message should appear (The specified URL cannot be called directly).

buddh4:
Also we should use buttons or links without href attribute in such cases in order to disable the option in the browser.

yurabakhtin:
@luke- PR https://github.com/humhub/humhub/pull/4814

buddh4:
Can we add some documentation for this when and how to use it (at least in the migration guide). I assume no one will think of using this access rule and the `open in new tab` issue when implementing a new feature otherwise.

PR #4814, opened 2021-02-01
Forbid to open AJAX actions as separate page

yurabakhtin:

- [x] Bugfix
- [x] Feature

- [x] Changelog was modified

**Other information:**
Issue: https://github.com/humhub/humhub/issues/3733


luke- (approved):
Merge in 1.9

buddh4:
Just a suggestion, should we name the access rule just `ajax`, we have another rule `json`. When implementing (and not using the const) its easier to remember `ajax` instead of `onAjaxOnly`. I know its a more descriptive, but I think both names are clear enough.

luke-:
Do you mean only the value or also the name constant?

For me it's ok when we change the value of the constant to "ajax" only, otherwise I would prefer the name RULE_AJAX_ONLY  because it is very descriptive.

buddh4:
Yeah, we can keep the constant name but change the value. Ideally the constant is used anyways.

luke-:
Added to Migratin Docu: https://github.com/humhub/documentation/blob/develop/docs/develop/modules-migrate.md#migrate-from-18-to-19

Issue #4828, opened 2021-02-03
1.8: Default Permissions some Modules missing


luke-:
### What steps will reproduce the problem?
Install and activate Wiki Module
Go to space / user default permission settings,

### What is the expected result?

See Wiki permissions

### What do you get instead?

No Wiki Permissions

![image](https://user-images.githubusercontent.com/4736168/106762552-3b0e6b80-6636-11eb-9acb-b4b93ad0a813.png)

### Additional info

Some other module permissions appear

| Q                | A
| ---------------- | ---
| HumHub version   | 1.8
| PHP version      | 
| Operating system |

> Note: Also provide the full error stacks from `Administration --> Information --> Logging` and errors from your browser console if possible.
> Note: Please use the [community](https://community.humhub.com) site for any questions on customizing HumHub.

yurabakhtin:
@luke- Fixed in PR https://github.com/humhub/humhub/pull/4833.

I detected the default permissions were hidden only for modules that are deactivated by default:
<img width="854" alt="check_perms" src="https://user-images.githubusercontent.com/6995524/106870274-f7713b80-66e1-11eb-8eb2-53d9d7cfcfee.png">

Also after investigation I find some modules (extended from `ContentContainerModule`) define permissions in the method `getContainerPermissions()` and those modules had the problem like this https://github.com/humhub/humhub-modules-wiki/commit/6f2e9db1, but some modules define permission in the method `getPermissions()` like this https://github.com/humhub/humhub-modules-cfiles/blob/master/Module.php#L61-L71.
Here https://github.com/humhub/humhub/blob/master/protected/humhub/modules/content/components/ContentContainerModule.php#L184-L195 the method `ContentContainerModule->getContainerPermissions()` is called only for Activated modules.

It seems the modification like this https://github.com/humhub/humhub-modules-wiki/commit/6f2e9db1 should be done for all modules extended from `ContentContainerModule` instead of current `getPermissions()` with conditions `if ($contentContainer instanceof Space)` or `if ($contentContainer !== null)` inside. But I didn't investigate/test this deeply, let me know if I should, thanks.

yurabakhtin:
In additional to above the method `ContentContainerModule->getContainerPermissions()` https://github.com/humhub/humhub/blob/master/protected/humhub/modules/content/components/ContentContainerModule.php#L205 should be `public` and not `protected` I think.

luke-:
[...]

PR #4833, opened 2021-02-04
Display default permissions of all modules even if they are deactivat…

yurabakhtin:
…ed by default

- [x] Bugfix

- [x] Changelog was modified

**Other information:**
Issue: https://github.com/humhub/humhub/issues/4828


luke- (changes requested):
I get following error, since the method is not set to `public`.
image

What do you think about this approach? here we would not need to change any modules:
https://github.com/humhub/humhub/issues/4828#issuecomment-773345719

yurabakhtin:
@luke- I have fixed the error of unknown method `getContainerPermissions()` in the commit https://github.com/humhub/humhub/pull/4833/commits/fe5838273e9b9e0dc58070c5eff953bdece29d92.

- I changed the method `getContainerPermissions()` of the parent/general class `ContentContainerModule` to public, because as I see all modules have the method as `public` and not `protected`, and after my first commit of this PR we have to keep this as `public`.
- Then as I understand all modules extended from `ContentContainerModule` must define permissions in the method `getContainerPermissions()` and not in the `getPermissions()`. It seems not all container modules were updated to use the method. It why I did the fix https://github.com/humhub/humhub/pull/4833/commits/fe5838273e9b9e0dc58070c5eff953bdece29d92#diff-3ebc21df74e6cf05d89de34d4e59c1492ae1182da2cc03fe232e6235775c46ceR33-R37

---

I think we should fix all `class Module extends ContentContainerModule` to use `getContainerPermissions()` instead of current:

```php
public function getPermissions($contentContainer = null)
{
    if (!$contentContainer) {
        return [
            new permissions\SomePermission(),
        ];
    }

    return [];
}
```

but some modules have the permissions only for Space or User container by condition like `if ($contentContainer instanceof Space) {`, so we cannot modify the modules easy, probably we should firstly modify core in order to display the permission depending on `BasePermission::defaultAllowedGroups`:

```php
protected $defaultAllowedGroups = [
    Space::USERGROUP_OWNER,
    Space::USERGROUP_ADMIN,
    User::USERGROUP_SELF,
];
```

As I understand the permission should be used only for defined `defaultAllowedGroups`, so it is strange why there is additional conditions like `if ($contentContainer instanceof Space)`.

luke- (approved):

luke-:
@yurabakhtin `defaultAllowedGroups`  indicates whether the permission is enabled by default for a user group. The permission can also be used for other user groups. What is the problem with the used condition `if ($contentContainer instanceof Space) {`?

The next step is to make sure that all modules really use the getContentContainerPermission.

yurabakhtin:
@luke- 

> What is the problem with the used condition `if ($contentContainer instanceof Space) {`?

I think the condition should be kept in the `getContainerPermissions()` if we will modify all `ContentContainerModule` to use it instead of `getPermissions()`.
I mean the condition `if (!$contentContainer) {` can be easy removed during the modification like it was done e.g. here https://github.com/humhub/humhub-modules-wiki/commit/6f2e9db1#diff-074f64cd19bc13bc527290951a58a65cb54f2f35a43ce5bf95e5dbb20b4d297dL73, but we should be more careful for the conditions like `if ($contentContainer instanceof Space)` because if we will remove them then the permission will become visible for other Container (User) that is wrong.

protected/humhub/modules/content/components/ContentContainerDefaultPermissionManager.php
```diff
@@ -28,9 +28,21 @@ class ContentContainerDefaultPermissionManager extends PermissionManager
      */
     protected function getModulePermissions(\yii\base\Module $module)
     {
+        if ($module instanceof ContentContainerModule && method_exists($module, 'getContainerPermissions')) {
```
buddh4, line 31: Think we can remove the `method_exists` here since isntanceof should make sure this method exists, hiding it in sub classes should not be allowed.

protected/humhub/modules/content/components/ContentContainerDefaultPermissionManager.php
```diff
@@ -28,9 +28,21 @@ class ContentContainerDefaultPermissionManager extends PermissionManager
...
+            $containerPermissions = $module->getContainerPermissions(new $this->contentcontainer_class);
+            if (!empty($containerPermissions)) {
+                // Don't try to find container permissions in the ContentContainerModule::getPermissions() below
+                // since they are already defined in more proper method ContentContainerModule::getContainerPermissions()
+                return $containerPermissions;
```
buddh4, line 36: A ContentContainerModule can also export non container related permissions in `getPermissions()` think those won't be included in this logic?

buddh4: Also note that some container modules simply overwritten the `getPermissions()` instead of using the `getContainerPermissions()`

protected/humhub/modules/content/components/ContentContainerDefaultPermissionManager.php
luke-: https://github.com/humhub/humhub/commit/2059b55bc78fe127acfb0806b1e1c9df3a5e996e

protected/humhub/modules/content/components/ContentContainerDefaultPermissionManager.php
luke-: Look good for me. Modules can either export permissions via `getContainerPermissions()` (preferred) or `getPremissions()`.

buddh4: Ok, this permission manager is solely used to determine default permissions in the admin settings.

Issue #4831, opened 2021-02-03
1.8: Username validation breaks language change


luke-:
### What steps will reproduce the problem?

- Create a username with whitespace.
- Update to 1.8 with stricter username valdiation: https://github.com/humhub/humhub/commit/2bdc435486bd2b855fe170a29de6b2c8f675647e
- Try to change language: Account settings -> Settings -> language

### What is the expected result?

Changed language

### What do you get instead?

Stay in old language.

### Additional info

Ideally, the regexp username should apply only to new users or when the username is changed.

yurabakhtin:
@luke- Fixed in PR https://github.com/humhub/humhub/pull/4834

PR #4834, opened 2021-02-04
Validate username by pattern only on change or for new users

yurabakhtin:

- [x] Bugfix

- [x] Changelog was modified

**Other information:**
Issue: https://github.com/humhub/humhub/issues/4831


luke-:
@yurabakhtin this condition should solve the problem. Can we maybe check this in the "when"? then we dont need a new public method in the user model. https://www.yiiframework.com/doc/api/2.0/yii-validators-validator#$when-detail

luke- (approved):

yurabakhtin:
@luke- Yes, it works, thank you for the advice.

PR #4854, opened 2021-02-06
Update AuthController.php

funkycram:
I'm not sure it is correct and I haven't much time for testing, but `$authClient` doesn't always extends `BaseFormAuth` class. So this `onSuccessLogin` doesn't always exists.

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

- [X] It's submitted to the `develop` branch, _not_ the `master` branch if no hotfix
- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**

### DIFF
--- a/protected/humhub/modules/user/controllers/AuthController.php
+++ b/protected/humhub/modules/user/controllers/AuthController.php
@@ -226,7 +226,9 @@ class AuthController extends Controller
 
         if ($success) {
             $this->trigger(static::EVENT_AFTER_LOGIN, new UserEvent(['user' => Yii::$app->user->identity]));
-            $authClient->onSuccessLogin();
+            if (method_exists($authClient, 'onSuccessLogin')) {
+                $authClient->onSuccessLogin();
+            };
         }
 
         return $result;


protected/humhub/modules/user/controllers/AuthController.php
```diff
@@ -226,7 +226,9 @@ protected function login($user, $authClient)
 
         if ($success) {
             $this->trigger(static::EVENT_AFTER_LOGIN, new UserEvent(['user' => Yii::$app->user->identity]));
-            $authClient->onSuccessLogin();
+            if (method_exists($authClient, 'onSuccessLogin')) {
```
yurabakhtin, line 229: It would be better to use `if ($authClient instanceof \humhub\modules\user\authclient\BaseFormAuth) {` instead.

funkycram: Will work, yes, but prevents the possibility to trigger this method in a custom auth client.

yurabakhtin: ah yes, sorry, we think you solution is better because it may works as you wrote on custom auth clients and also on old core versions where the method `onSuccessLogin()` doesn't exist yet.

luke- (approved):

protected/humhub/modules/user/controllers/AuthController.php
ArchBlood: I'd also like to note that it's already possible to remove the old core OAuth clients in the future do to the new modules that were pushed to the marketplace, so in the future this might be the correct solution as of right now but I also agree that @yurabakhtin's solution is the right one to use in the future

https://www.humhub.com/en/marketplace?keyword=Sign-In

Issue #4868, opened 2021-02-11
Reset of existing email summaries / notifications settings


luke-:
### Is your feature request related to a problem? Please describe.

![image](https://user-images.githubusercontent.com/4736168/107624701-8db1de00-6c5b-11eb-9b93-17037e067819.png)

yurabakhtin:
@luke- I have stared this in PR https://github.com/humhub/humhub/pull/4873, but I think it is not completed yet, because it seems not all notification settings are reset. I will do this later because no time now.

luke-:
@yurabakhtin Looks good so far! Have a nice weekend! :-)

yurabakhtin:
@luke- I have completed the PR, but I think one setting should be reset as well, currently it is not reset even per current User, please see on the screenshot:

<img width="1148" alt="user_settings" src="https://user-images.githubusercontent.com/6995524/107914342-42f1d800-6f73-11eb-8dd0-7177df44cdbb.png">

On my opinion the checkbox setting should be reset on click the grey button "Reset to defaults" and on click new red button "Reset for all users", do you agree?

<img width="1147" alt="general_settings" src="https://user-images.githubusercontent.com/6995524/107914440-7d5b7500-6f73-11eb-98e2-ad7c5e3f9eea.png">

luke-:
@yurabakhtin Yes, I agree, the "Desktop Noticiations" should be also reset.

yurabakhtin:
@luke- 

> Yes, I agree, the "Desktop Noticiations" should be also reset.

Commit https://github.com/humhub/humhub/pull/4873/commits/cea8a91a65b57d6bc93f7ef8496fe06a16c97cac.

yurabakhtin:
@luke- I think this the PR is completed and can be merged.

PR #4873, opened 2021-02-12
Reset email summaries / notifications settings for all users

yurabakhtin:

- [x] Feature

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
Issue: https://github.com/humhub/humhub/issues/4868


protected/humhub/modules/notification/components/NotificationManager.php
```diff
@@ -338,6 +338,18 @@ public function setSpaces($spaceGuids, User $user = null)
...
+    /**
+     * Reset the notification space settings for all users
+     */
+    public function resetSpaces()
+    {
+        // Reset notifications for all selected membership spaces
+        Membership::updateAll(['send_notifications' => 0]);
+
+        // Delete all selected following spaces
+        Follow::deleteAll(['object_model' => Space::class]);
```
luke-, line 350: Would that delete all existing Space Followings? That would be more than expected. Is that necessary?

luke- (changes requested):

protected/humhub/modules/notification/components/NotificationManager.php
yurabakhtin: > Would that delete all existing Space Followings?

Yes, because the table contains the single flag column `send_notifications`:

space_following

But if you think we should not delete the records and can only update the flag to `0`, for that case if in future some new flag/columns will be added there, then we should replace the `deleteAll()` to:

```php
Follow::updateAll(['send_notifications' => 0], ['object_model' => Space::class]);
```

Do you prefer this solution?

luke-: @yurabakhtin Yes, the solution is better. Then it really only affects the notifications.

yurabakhtin: @luke- Ok, I have fixed that in commit https://github.com/humhub/humhub/pull/4873/commits/7bcbd87149bce7be422f9f4651154a4cea051d39

luke- (approved):

Issue #4841, opened 2021-02-05
1.8: Notification Filter misplaced


luke-:
![image](https://user-images.githubusercontent.com/4736168/107040485-90c24f80-67bf-11eb-90c5-fb6487e1eefc.png)

buddh4:
The `margin-bottom: 0;` seems to be the problem, not sure when and why it was introduced:
![grafik](https://user-images.githubusercontent.com/9800854/107042893-88b7df00-67c2-11eb-96d4-b0c7c77adfe3.png)

yurabakhtin:
@luke- PR https://github.com/humhub/humhub/pull/4846
<img width="1152" alt="fixed" src="https://user-images.githubusercontent.com/6995524/107057820-e5c48c80-67e4-11eb-8cbf-53644bf9dd4d.png">

buddh4:
@yurabakhtin did you investigate when this was introduced? Just to make sure this is not an issue anywhere else.

luke-:
@MasterHans  Seems it was introduced with https://github.com/humhub/humhub/commit/2cddf6b8babe257ec9228f53ed1db6c77d110e48 and related to the `/admin/pending-registrations` feature. Please make sure that such CSS/LESS changes are always as precise as possible to avoid such effects. 

Can you fix that?

buddh4:
Same with https://github.com/humhub/humhub/commit/2cddf6b8babe257ec9228f53ed1db6c77d110e48#diff-cee91cfe4da6d567cb8e4101eb29e760ed346def1c9b1de8e2d9472e4b1a2fffR122 is there a reason this was added?

yurabakhtin:
@buddh4 

> did you investigate when this was introduced?

I didn't find when it was introduced but I found where the `checkboxList()` works well and compared the differences there:

<img width="1487" alt="differences" src="https://user-images.githubusercontent.com/6995524/107176146-9e562000-69df-11eb-9aeb-34e757203ac6.png">

What differences between `yii\widgets\ActiveForm` and `yii\bootstrap\ActiveForm`:

- `yii\widgets\ActiveForm` uses same template for all fields:
  - `public $template = "{label}\n{input}\n{list}\n{error}";` - `protected/vendor/yiisoft/yii2/widgets/ActiveField.php`
 
- `yii\bootstrap\ActiveForm` uses custom template for checkbox:
[...]

PR #4874, opened 2021-02-12
Delete pending registrations automatically - 2 bugfix #4225

MasterHans:
BugFix for
#4841 

Made style of label selector for checkboxes in the pending registration grid.

### DIFF
--- a/static/less/form.less
+++ b/static/less/form.less
@@ -70,6 +70,10 @@ select.form-control:not([multiple]) {
 
 label {
     font-weight: normal;
+}
+
+//.select-on-check-all:first-child
+thead > tr > th > label, tbody > tr > td > label {
     margin-bottom: 0;
     height: 17px;
 }
--- a/static/less/form.less
+++ b/static/less/form.less
@@ -72,7 +72,6 @@ label {
     font-weight: normal;
 }
 
-//.select-on-check-all:first-child
 thead > tr > th > label, tbody > tr > td > label {
     margin-bottom: 0;
     height: 17px;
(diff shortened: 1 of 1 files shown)


static/less/form.less
```diff
@@ -70,6 +70,10 @@ select.form-control:not([multiple]) {
 
 label {
     font-weight: normal;
+}
+
+//.select-on-check-all:first-child
```
luke-, line 75: @MasterHans Why is this commented out? Shouldn't that be restricting the change?

luke- (changes requested):

static/less/form.less
MasterHans: This is debugging comment. Unfortunately its could not work on here

luke-:
@MasterHans 

Is the Label CSS change only for Pending Registrations Table or are there issues here in other locations as well? 
If it is just the Pending Registration table, it might be good to give the table a class or find another unique selector.

e.g.
image

Untested:

```less
div.PendingRegistrations {
    thead > tr > th > label, tbody > tr > td > label {
        margin-bottom: 0
        height: 17px;
    }
}
```

MasterHans:
> @MasterHans
> 
> Is the Label CSS change only for Pending Registrations Table or are there issues here in other locations as well?
> If it is just the Pending Registration table, it might be good to give the table a class or find another unique selector.
> 
> e.g.
> image
> 
> Untested:
> 
> ```css-less
> div.PendingRegistrations {
>     thead > tr > th > label, tbody > tr > td > label {
>         margin-bottom: 0
>         height: 17px;
>     }
> }
> ```

Done thank you!

luke- (approved):

Issue #871, opened 2015-03-31
RTL Version


sajaddp:
Hi,
I today see your script it's very good! :)
if you want i can help you to complete set script for RTL language (ex: Persian, Pashto, Arabic, ...).
Since three years ago i worked on SHARETRONIX and create many plugin for Persian site's.

2012mjm:
I can help you to complete set script for RTL language ;)

buddh4:
Hi, think this should be done within an own theme? http://docs.humhub.org/theme-README.html

Felli:
I agree it would be just as easy to just create a rtl theme design.

rezaervani:
What about RTL button on prosemirror ? What should we do to get it ?

We don't want to change all theme into RTL version, just add both LTR and RTL Text function to editor.

Thank you,

luke-:
@rezaervani Can you please create an issue here: https://github.com/humhub/humhub-prosemirror/issues

ArchBlood:
@luke- or @buddh4 I'd like to bring a slightly better solution to the table when it comes to RTL, is that alright? The following code should trigger RTL in the case that Arabic language is used, of course if there is another language that also uses RTL then it could be added as well, any thoughts? Reason behind this, I don't think we should force custom theme option because the core code is always changing and will force the theme developer to recompile over each major change

Here's the code that I provide be used
```php
<html <?php if(Yii::$app->language == 'ar')  echo "dir='rtl'"; ?>>
```

PR #4875, opened 2021-02-12
Enh: RTL support by default on language change

ArchBlood:

- [x] Feature

- [x] Yes

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**
I believe this will make all parties happy if it's implemented by default on language change and it is a better solution to #871 overall than view overwriting by themes

### English
Screenshot_1

### Arabic
Screenshot_2

### DIFF
--- a/protected/humhub/views/layouts/main.php
+++ b/protected/humhub/views/layouts/main.php
@@ -6,7 +6,7 @@
 ?>
 <?php $this->beginPage() ?>
 <!DOCTYPE html>
-<html lang="<?= Yii::$app->language ?>">
+<html lang="<?= Yii::$app->language ?>" <?php if(Yii::$app->language == 'ar') echo "dir='rtl'"; ?>>
     <head>
         <title><?= strip_tags($this->pageTitle); ?></title>
         <meta charset="<?= Yii::$app->charset ?>">


ArchBlood:
Hmm... Just noticed that collapsible widgets don't support RTL so that could be a problem here as well :thinking:

luke- (changes requested):
Couldn't you set such a property via Javascript?

Otherwise, we need to find a solution here that also supports other languages without such a template change.

ArchBlood:
> Couldn't you set such a property via Javascript?
> 
> Otherwise, we need to find a solution here that also supports other languages without such a template change.

This method should support all languages by default, on change to `AR` (i.e Arabic) by the user/admin settings it will only show the RTL affect, setting this through JavaScript would probably add more load time I'd think, and you'd still have to specify all of the RTL languages, also you'd need to add CSS/LESS regardless of the method used :thinking:

luke-:
Seems you could also specify the direction as an attribute via CSS. We could then set this dynamically if the active language requires it.

However, Bootstrap only seems to really support this from version 5 onwards. https://getbootstrap.com/docs/5.0/getting-started/rtl/

Without knowing the exact consequences, I would be careful with such a change.

ArchBlood:
Very true, `dir="RTL"` is still the best option to be honest and requires less work but CSS/LESS changes will still need done in some areas such as panels and many dropdown elements :thinking:

luke-:
I'm not sure if it's good to enable the RTL option when elements (Menus, DropDowns and generally Bootstrap) don't fully support it yet? Presumably it's then more usable  (also in languages like AR) in LTR than with a buggy RTL view. 

Instead of the PR, it would be good to open a RTL Feature Branch and Draft PR and collect RTL specific improvements & tests there.

What do you think about that?

ArchBlood:
> I'm not sure if it's good to enable the RTL option when elements (Menus, DropDowns and generally Bootstrap) don't fully support it yet? Presumably it's then more usable  (also in languages like AR) in LTR than with a buggy RTL view. 
> 
> Instead of the PR, it would be good to open a RTL Feature Branch and Draft PR and collect RTL specific improvements & tests there.
> 
> What do you think about that?

That would be completed fine, I'm sure that there would be a lot of people that would be happy with such a feature

ArchBlood:
After searching throughout all available Bootstrap communities to me, here's three options that are currently available for this very thing;

- https://github.com/morteza/bootstrap-rtl (Bootstrap 3.x)
- https://bootstrap.rtlcss.com/ (Bootstrap 4.x)
- https://getbootstrap.com/docs/5.0/getting-started/rtl/ (Migrate to 5.x for official RTL support as you've pointed out)

ArchBlood:
So I've gotten some of the RTL finished and placed into a single CSS file, now the issue is implementing it into the core of the project as a single file, this is where I'll have to leave it to the core team :thinking:

### theme.css in rtl
[rtl.zip](https://github.com/humhub/humhub/files/6065345/rtl.zip)

PR #4879, opened 2021-02-15
External users module

yurabakhtin:

- [x] **Extend core for new module https://github.com/humhub/humhub-modules-external-users**
- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
Task/Issue: https://github.com/humhub/tasks/issues/64


luke- (approved):

buddh4:
Is there a (public) ticket available describing the planned API and steps for this? As far as I understood this will be a non trivial feature and will affect beside others the following parts of the application:

- `ContentActiveRecord::find()->readable()`
- `Streams: Content::find()` (Especially Dashboard streams)
- `Content::canView()`,  `Content::canEdit()` etc
- Space memberships / invites / default spaces
- Search system
- Global and Container Permission Manager?
- UserPicker/SpacePicker/TopicPicker/...
- Overview pages as calendar/tasks/directory
- All modules not using `readable()` or `visible()`
- Not related to content visibility, but should external users be allowed to install modules etc?

Furthermore unfortunately the content visibility query logic is duplicated e.g. in streams and `readable()` etc.
Also are there other use-cases we can think of beside the external user module which is planed?

luke-:
@buddh4 No, the main development takes place here: https://github.com/humhub/humhub-modules-external-users

buddh4:
@luke- Ok, do you think I should open a ticket for the mentioned list, or do you have those issues already in mind? Not sure if all of the issues will be a problem, but at least we should test them.

luke-:
@buddh4 Thanks. Just keep it here. At a later stage in the dev, I would create separate issues in the repo for relevant issues.

buddh4:
Here some possible use-cases:

- external staff
- multi-client system (crm)
- spaces only acessible by condition (e.g. payment)
- ban users from space
- block users

protected/humhub/modules/user/behaviors/Followable.php
```diff
@@ -143,37 +143,47 @@ public function isFollowedByUser($userId = null, $withNotifications = false)
         return false;
     }
 
+    /**
+     * Get a query of users which are followers of this object.
+     *
+     * @param ActiveQueryUser $query e.g. for limit the result
+     * @return ActiveQueryUser
+     */
+    public function getFollowersQuery($query = null)
```
buddh4, line 152: Do we need or use the `$query` param anywhere? Limit etc. can be also set on the returned query. I can't find a use case in we'd need to provide a query beforehand. Same with `getFollowingQuery()`.

yurabakhtin: @buddh4 I created the method `getFollowersQuery($query = null)` just to factorize the code from the already existed method `getFollowers($query = null, ...)` which already had the param `$query = null` and used like `$user->getFollowers(User::find()->limit(16))` here https://github.com/humhub/humhub/blob/enh/64-external-users-module/protected/humhub/modules/user/widgets/views/userFollower.php#L6

buddh4: Ok, In this case I'd leave the query param out. The `getFollowers()` does not return a query and therefore its not possible to set a limit otherwise. But in the `getFollowersQuery()` and `getFollowingQuery()` we receive a query and can set such a limit etc. afterwards.

I'd also prefer a similar `public function getFollowers($withNotification = false)` function, but think we can't change this without introducing a breaking change.

buddh4: Can we maybe add the `$withNotifications` in the `getFollowersQuery()` and reuse this in the `getFollowers()` maybe we could even deprecate the old `getFollowers()` in favor of  `getFollowersQuery()`

yurabakhtin: @buddh4 Please check how I implemented this here https://github.com/humhub/humhub/pull/4879/commits/f3019536ea07952abbd4e57a6284cc88f327a6ea.

protected/humhub/modules/admin/controllers/GroupController.php
```diff
@@ -301,4 +297,27 @@ public function checkGroupAccess($group)
...
         }
     }
+
+    /**
+     * Get a restriction message if the Group cannot be deleted
+     *
+     * @param Group $group
+     * @return string|bool Error message if the Group cannot be deleted, false - if no restrictions to delete the Group
+     */
+    protected function getGroupDeleteRestriction($group)
```
buddh4, line 307: We should make this private, I think there is no need for a protected visibility. Private functions can be refactored/deleted without introducing breaking changes in the future. If we need it as protected we can change it in the future.

protected/humhub/modules/user/models/Group.php
```diff
@@ -472,4 +473,21 @@ public function getGroupSpaces()
...
+
+    /**
+     * Check if this Group can be deleted by current User
+     *
+     * @return bool
+     * @since 1.9
+     */
+    public function canDelete()
+    {
+        return Yii::$app->user->isAdmin() && !(
```
buddh4, line 486: We also have a `ManageGroups` permission, we may rather should check for this.

protected/humhub/modules/admin/views/group/index.php
```diff
@@ -36,7 +36,8 @@
                 'format' => 'html',
                 'value' => function (Group $group) {
                     return $group->name .
-                        ($group->is_default_group ? ' <span class="badge">' . Yii::t('AdminModule.user', 'Default') . '</span>' : '');
+                        ($group->is_default_group ? ' <span class="badge">' . Yii::t('AdminModule.user', 'Default') . '</span>' : '') .
```
buddh4, line 39: Maybe use the `humhub\widgets\Label` here.

protected/humhub/modules/user/behaviors/Followable.php
buddh4: We still should remove the `$query` param here. 

Also we probably should move all the details of this query logic to the `humhub\modules\user\models\Follow` record. The `Follow` class also contains follow related query logic and we should keep all follow related query details in one place. `Followable` should just delegate the calls to the `Follow` model class. Something like `Follow::findFollowersOf($container)` and `Follow::findFollowing($container)`.

protected/humhub/modules/user/behaviors/Followable.php
```diff
@@ -143,37 +143,59 @@ public function isFollowedByUser($userId = null, $withNotifications = false)
...
     /**
      * Returns an array of users which are followers of this object.
      *
-     * @param CDbCriteria $eCriteria e.g. for limit the result
+     * @deprecated Use $this->getFollowersQuery()->all() instead
+     * @param ActiveQueryUser $query e.g. for limit the result
      * @param boolean $withNotifications only return followers with enabled notifications
      * @param boolean $returnQuery only return the query instead of User objects
-     * @return Users[]|ActiveQueryUser the user objects or the active query
+     * @return User[]|ActiveQueryUser the user objects or the active query
```
buddh4, line 194: We should deprecate this one.

protected/humhub/modules/user/behaviors/Followable.php
```diff
@@ -208,16 +243,14 @@ public function getFollowingCount($objectModel)
      *
      * E.g. Get list of spaces which are the user follows.
      *
-     * @param CDbCriteria $eCriteria e.g. for limit the result
+     * @deprecated Use $this->getFollowingQuery()->all() instead
+     * @param ActiveQueryUser $query e.g. for limit the result
      * @param string $objectModel HActiveRecord Classname to restrict Object Classes to (e.g. User)
-     * @return Array
+     * @return User[]
```
buddh4, line 249: Also deprecate this one, I think its only used in `userFollower` view.

protected/humhub/modules/user/behaviors/Followable.php
```diff
@@ -186,20 +208,33 @@ public function getFollowers($query = null, $withNotification = false, $returnQu
         return $query->all();
     }
 
+    /**
+     * Get a query of objects which the owner object follows
+     *
+     * @param ActiveQueryUser $query e.g. for limit the result
+     * @return ActiveQueryUser
+     */
+    public function getFollowingQuery($query = null)
```
buddh4, line 217: I think the only reason we have to keep the `$query` parameter here is because of the old deprecated functions. I don't like the idea of adding a new function with a parameter we do not really need/want. Maybe we can remove the $query parameter and find a better solution. We should maybe just ignore the `$query` param in the deprecated functions and refactor all the usages. I assume the $query param is hardly used anyways.

protected/humhub/modules/user/controllers/ProfileController.php
```diff
@@ -129,7 +129,7 @@ public function actionFollowerList()
         $query->leftJoin('user_follow', 'user.id=user_follow.user_id AND object_model=:userClass AND user_follow.object_id=:userId', [':userClass' => User::class, ':userId' => $this->getUser()->id]);
         $query->orderBy(['user_follow.id' => SORT_DESC]);
         $query->andWhere(['IS NOT', 'user_follow.id', new Expression('NULL')]);
-        $query->active();
+        $query->visible();
```
buddh4, line 132: Move both of those queries to the `Follower` class. Controllers should not contain db query logic anyways.

buddh4:
Here a summary of the current changes, I think its easier to get an overview of the main api changes:

# Group changes

- Added `Group::is_protected` db flag to groups in order to prevent certain groups from being deleted. 
Custom modules can use this flag for special group types which are not manually deletable.

- Added `Group::canDelete()` to determine if the current user is allowed to delete a group instance.

# ContentContainer changes

- Added `ContentContainer::findVisibleRecord($guid)` in order to fetch a `ContentCotnainerActiveRecord` instance by guid and validate visibility checks

# Space changes

- Added `ActiveQuerySpace::EVENT_CHECK_VISIBILITY` in order for custom modules to extend visibility checks of `ActiveQuerySpace::visible()`

# Permission changes

Added 

- `BasePermission::EVENT_INIT`
- `BasePermission::addFixedGroups($groupIds)`

in order to overwrite fixed groups and other permission settings by event.

buddh4 (commented):
Since the min PHP version is 7.2 we now should use a proper method signature with return type and property types where possible.

protected/humhub/modules/user/behaviors/Followable.php
buddh4: When further thinking about it I suggest deprecating all old query related functions:

- `Followable::getFollowers()`
- `Followable::getFollowingObjects()`

and not adding any new function to  `Followable`, but instead move (or merge with existing) query function to `Follow` class.

buddh4: Can you also have a look at, not sure if this is a bug:

grafik

yurabakhtin: > We still should remove the $query param here.

Commit https://github.com/humhub/humhub/pull/4879/commits/5394e00c89fce92a6220d916d5a555a5f127fb8e.

protected/humhub/modules/user/behaviors/Followable.php
yurabakhtin: > We still should remove the $query param here.

Commit https://github.com/humhub/humhub/pull/4879/commits/5394e00c89fce92a6220d916d5a555a5f127fb8e.

protected/humhub/modules/user/behaviors/Followable.php
yurabakhtin: Commit https://github.com/humhub/humhub/pull/4879/commits/5394e00c89fce92a6220d916d5a555a5f127fb8e

protected/humhub/modules/user/controllers/ProfileController.php
yurabakhtin: Commit https://github.com/humhub/humhub/pull/4879/commits/54b9428194fdc706860bdeceebf2773c6cbeeee0

protected/humhub/modules/user/behaviors/Followable.php
yurabakhtin: I already wrote above this `* @deprecated Use $this->getFollowingQuery()->all() instead`, is it not enough for deprecation? If no, please let me know what I should do, thanks.

protected/humhub/modules/user/models/Group.php
yurabakhtin: Thanks, fixed in commit https://github.com/humhub/humhub/pull/4879/commits/774fab0ff7e84b69ba3fcef180d88232041ea862

protected/humhub/modules/admin/controllers/GroupController.php
yurabakhtin: Ok, commit https://github.com/humhub/humhub/pull/4879/commits/eac110a6822adbfabbf0ff8e1dc85dd9d70abe3c

protected/humhub/modules/admin/views/group/index.php
yurabakhtin: Ok, I have applied this in commit https://github.com/humhub/humhub/pull/4879/commits/87b0b4163c7863a41bcb27acca934d97deedf6cb but please note the `.label` is used there instead of `.badge`, for me it is almost same style, if we really need badge style I can revert the commit:

label_vs_badge

buddh4: I think it looks better. We probably should get rid of `badges` in favor of `label`.

protected/humhub/modules/user/behaviors/Followable.php
buddh4: Sorry, probably missed the phpdoc. But we should always add the version like `@deprecated since 1.9 some replacement infos`

protected/humhub/modules/user/behaviors/Followable.php
yurabakhtin: @buddh4 

> but instead move or merge with existing query functions in Follow class and use those instead of the deprecated ones.

As I understood you all query methods in the `humhub\modules\user\behaviors\Followable` should be deprecated at all, so instead of current `$this->user->getFollowersQuery()->limit(16)->all()` we will use new `humhub\modules\user\models\Follow::getFollowersQuery($this->user)->limit(16)->all()`.

Note the `Follow::getFollowersQuery(ActiveRecord $target, $withNotifications = null)` already exists with the params, but I see there different query with subquery instead of `LEFT JOIN` as we have in the `Followable->getFollowersQuery()`, so probably we should change this to `LEFT JOIN` as well, or do you think the subquery solution is better?

And probably we should create new `Follow::getFollowingQuery(ActiveRecord $target)` instead of `Followable->getFollowingQuery()` or I should to investigate other methods inside the `Follow`, maybe there some method like `getFollowedContainerIdQuery()` already does same.

buddh4: > As I understood you all query methods in the humhub\modules\user\behaviors\Followable should be deprecated at all, so instead of current $this->user->getFollowersQuery()->limit(16)->all() we will use new humhub\modules\user\models\Follow::getFollowersQuery($this->user)->limit(16)->all().

Yes I'd deprecate the two function and implement the actual logic in `Follow` in order to unify and remove redundance:

- Followable::getFollowers()
- Followable::getFollowingObjects()

Also logic like `getFollowerCount()` should only delegate `Follow` without needing to know about query details.

```
public function getFollowerCount()
    {
        return Follow::find()->where(['object_id' => $this->owner->getPrimaryKey(), 'object_model' => $this->owner->className()])->count();
    }
```

> Followable->getFollowersQuery(), so probably we should change this to LEFT JOIN as well, or do you think the subquery solution is better?

The join is probably faster.

> And probably we should create new Follow::getFollowingQuery(ActiveRecord $target) instead of Followable->getFollowingQuery() or I should to investigate other methods inside the Follow, maybe there some method like getFollowedContainerIdQuery() already does same.

Feel free to make a suggestion what you think is the best way of migrating/moving the logic from `Followable` to  `Follow` if possible without breaking changes and redundence. But if you feel like your prefered solution would require a breaking change, we can also think about it.

yurabakhtin:
@buddh4 

> Since the min PHP version is 7.2 we now should use a proper method signature with return type and property types where possible.

I try to use same code style what was written before.
Do you mean for all new functions I should use new code style for php 7.2 like `public function funcName(string $param1, bool $param2): int`?

protected/humhub/modules/user/behaviors/Followable.php
yurabakhtin: > Can you also have a look at the following, pretty sure this is a bug, but is probably never called:

I don't find where the method `Followable->beforeDelete()` is used, so I deleted it completely in commit https://github.com/humhub/humhub/pull/4879/commits/b3f1240adabdbb8c3d7325e93f49b9a89bea84a8

luke-:
@yurabakhtin FYI, the PR looks good so far and I've merged it for v1.10. If more changes are needed, we will have to open a new PR. Thanks for the good work!

Issue #4869, opened 2021-02-11
Comment Notification E-Mail 


luke-:
### What steps will reproduce the problem?

Receive a notification about a comment

### What is the expected result?

See the comment and the original text. 

### What do you get instead?

See comment twice.
![image](https://user-images.githubusercontent.com/4736168/107632558-edfa4d00-6c66-11eb-8987-21eba50b6368.png)

May introduced which RichText parser enhancements.

### Additional info

| Q                | A
| ---------------- | ---
| HumHub version   | 1.8

yurabakhtin:
@luke- PR https://github.com/humhub/humhub/pull/4870:

<img width="1245" alt="test_email_content" src="https://user-images.githubusercontent.com/6995524/107736195-8bf72180-6d12-11eb-8c03-1b9755e0095c.png">

The problem was in the cached content with same key for Comment and for Post because there(https://github.com/humhub/humhub/blob/fix/4869-cached-comment-notification-content/protected/humhub/modules/content/widgets/richtext/converter/BaseRichTextConverter.php#L170-L174) same value id is used, probably there we should use a code like `$content->id` instead of current `$content->content->id`, but currently I fixed only the single case to make the key unique, you can see the debug for more details:

<img width="1703" alt="debug_cached_mail_content" src="https://user-images.githubusercontent.com/6995524/107736446-1b9cd000-6d13-11eb-80f4-0b5c87cbf20d.png">

luke-:
@yurabakhtin Thanks, for the fix. 
@buddh4 As you are currently working on it, you might want to move the fix a little deeper into the API (if necessary).

Issue #4857, opened 2021-02-06
Reply Notifications share same message as first reply


ArchBlood:
### What steps will reproduce the problem?
Comment on any post then reply to the comment twice and have someone like both replies to your comment and you'll get 2 notifications with the same text from the first reply

### What is the expected result?
When someone likes a reply or multiple replies it should be the exact text not a copy of the first reply

### What do you get instead?
![Screenshot_1](https://user-images.githubusercontent.com/35392110/107129588-6db59f80-6894-11eb-8db7-2e3443cc7736.png)

### Additional info
This is very confusing and could cause users to think that the notifications are being spammed by someone

| Q                | A
| ---------------- | ---
| HumHub version   | 1.8-beta.1
| PHP version      | 7.3
| Operating system | CloudLinux

> Note: Also provide the full error stacks from `Administration --> Information --> Logging` and errors from your browser console if possible.
> Note: Please use the [community](https://community.humhub.com) site for any questions on customizing HumHub.

luke-:
@yurabakhtin Can you please take a look? Sounds similar to this: https://github.com/humhub/humhub/commit/17e514355ce34572fd36b5c2fd885421c0f18951

yurabakhtin:
@luke- Yes, this but is related to the commit from issue https://github.com/humhub/humhub/issues/4869, as I wrote there it would be better to fix main function `BaseRichTextConverter::buildCacheKeyForContent()`. I have done this in new PR https://github.com/humhub/humhub/pull/4882 because as we see it was not enough to fix the single case in the previous PR.

---

Not related to this issue but please see what error IDE shows me:
<img width="1281" alt="abstract_func_call" src="https://user-images.githubusercontent.com/6995524/108039291-f5e23480-704c-11eb-92b4-471d9cbfbf96.png">
Probably the code `return parent::renderAbsy($blocks);` never calls but I think it should replaced with `return ''`, do you agree?

luke-:
@yurabakhtin Please see my PR comments. I also get the warning about `renderAbsy`. Can you please fix that too? I think either we remove the `parent::renderAbsy` or implement an empty function in the parent.

yurabakhtin:
@luke- 

> I also get the warning about renderAbsy. Can you please fix that too?

Commit https://github.com/humhub/humhub/pull/4882/commits/ccad96646c51cd4f3db7aa873faed352a5afeafd

buddh4:
[...]

PR #4882, opened 2021-02-16
Fix wrong content id used in building cache key

yurabakhtin:

- [x] Bugfix

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
Issue: https://github.com/humhub/humhub/issues/4857
Related issue: https://github.com/humhub/humhub/issues/4869

### DIFF
--- a/protected/humhub/modules/content/widgets/richtext/converter/BaseRichTextConverter.php
+++ b/protected/humhub/modules/content/widgets/richtext/converter/BaseRichTextConverter.php
@@ -12,6 +12,7 @@ use cebe\markdown\GithubMarkdown;
 use humhub\components\ActiveRecord;
 use humhub\components\Event;
 use humhub\libs\Html;
+use humhub\modules\content\components\ContentAddonActiveRecord;
 use humhub\modules\content\interfaces\ContentOwner;
 use humhub\modules\content\widgets\richtext\extensions\link\LinkParserBlock;
 use humhub\modules\content\widgets\richtext\extensions\link\RichTextLinkExtension;
@@ -169,10 +170,12 @@ abstract class BaseRichTextConverter extends GithubMarkdown
      */
     public static function buildCacheKeyForContent(ContentOwner $content, $prefix = null)
     {
-        $result = 'content_' . $content->content->id;
-        if (isset($content->id)) {
-            $result .= '_' . $content->id;
+        if ($content instanceof ContentAddonActiveRecord) {
+            // prevent cache key for comments to use the same cache key of original content
+            return static::buildCacheKeyForRecord($content);
         }
+
+        $result = 'content_' . $content->content->id;
         return $prefix ? $prefix.'_'.$result : $result;
     }
 
--- a/protected/humhub/modules/content/widgets/richtext/converter/BaseRichTextConverter.php
+++ b/protected/humhub/modules/content/widgets/richtext/converter/BaseRichTextConverter.php
@@ -287,7 +287,7 @@ abstract class BaseRichTextConverter extends GithubMarkdown
             });
         }
 
-        return '';
+        return parent::renderAbsy($blocks);
     }
 
 
--- a/protected/humhub/modules/content/widgets/richtext/converter/BaseRichTextConverter.php
+++ b/protected/humhub/modules/content/widgets/richtext/converter/BaseRichTextConverter.php
@@ -169,7 +169,10 @@ abstract class BaseRichTextConverter extends GithubMarkdown
      */
     public static function buildCacheKeyForContent(ContentOwner $content, $prefix = null)
     {
-        $result = 'content_'.$content->id;
+        $result = 'content_' . $content->content->id;
+        if (isset($content->id)) {
+            $result .= '_' . $content->id;
+        }
         return $prefix ? $prefix.'_'.$result : $result;
     }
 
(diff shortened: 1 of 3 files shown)


luke- (changes requested):
Unfortunately the ContentOwner interface doesn't provide an ID attribute, what do you think about something like this:

```php
        $result = 'content_' . $content->content->id;
        if (property_exists($content, 'id')) {
            $cacheKey .= '_' . $content->id;
        }
``` 

A better approach would of course be:
- Pass `id` somehow as prefix to the method
- Add getId to `ContentOwner` (Probably not possible yet in Beta due to dependencies & modules etc.)

yurabakhtin:
@luke- Yes, I thought about the solution like `isset($content->id)` or `!empty($content->id)` too, so I agree your solution.

yurabakhtin:
@luke- https://github.com/humhub/humhub/pull/4882/commits/a29e95d8d3866ccfad4af38241f7645ada0f3e90 - I did to use `isset($content->id)` because the `property_exists($content, 'id')` returns `false`, but I didn't use `!empty($content->id)` in order to append the suffix for `0` value even if it may be not happen.

protected/humhub/modules/content/widgets/richtext/converter/BaseRichTextConverter.php
```diff
@@ -169,7 +169,10 @@ public static function process($text, $options = [])
      */
     public static function buildCacheKeyForContent(ContentOwner $content, $prefix = null)
     {
-        $result = 'content_'.$content->content->id;
+        $result = 'content_' . $content->content->id;
+        if (isset($content->id)) {
+            $result .= '_' . $content->id;
```
buddh4, line 174: We always need the id of the content and not the record here. Otherwise a wiki page with id 1 will return the same result as post with id 1. The `ContentOwner` interface should always return a `Content` by `$content->content` unfortunately the interface is broken with content active records since it returns an activequery instead of the content itself when calling `getContent()`

protected/humhub/modules/content/widgets/richtext/converter/BaseRichTextConverter.php
```diff
@@ -284,7 +287,7 @@ protected function renderAbsy($blocks)
             });
         }
 
-        return parent::renderAbsy($blocks);
+        return '';
```
buddh4, line 290: Why this change this will should break the whole parser process or am I missing something?

buddh4 (changes requested):
So this seems to be an issue with comments? Maybe just add a `instanceof ContentAddon` or something?

protected/humhub/modules/content/widgets/richtext/converter/BaseRichTextConverter.php
buddh4: Note there is also a `buildCacheKeyForRecord` which will include the record class.

protected/humhub/modules/content/widgets/richtext/converter/BaseRichTextConverter.php
yurabakhtin: I changed this because the `parent::renderAbsy($blocks)` is abstract method and IDE displays error, please review my second screenshot from this comment https://github.com/humhub/humhub/issues/4857#issuecomment-779682809.

buddh4: See https://github.com/humhub/humhub/issues/4857#issuecomment-779821434 this seems to be a problem with phpstorm. The method is overwritten in a parent class already.

buddh4:
Also related to https://github.com/humhub/humhub/pull/4870/files I assume that the issue is not the `buildCacheForContent`, but rather that the `ContentOwner` interface is violated in such cases.

protected/humhub/modules/content/widgets/richtext/converter/BaseRichTextConverter.php
buddh4: @yurabakhtin I think the following would solve the issue:

```
if($content instanceof ContentAddonActiveRecord) {
    return static::buildCacheKeyForRecord($content);
}

// revert the rest back to prior this PR
```

protected/humhub/modules/content/widgets/richtext/converter/BaseRichTextConverter.php
yurabakhtin: ok, I reverted that here https://github.com/humhub/humhub/pull/4882/commits/27e2022aebcf6c84b8098b715a9566dc47360075

protected/humhub/modules/content/widgets/richtext/converter/BaseRichTextConverter.php
yurabakhtin: @buddh4 Yes, your solution works well, so I applied this here https://github.com/humhub/humhub/pull/4882/commits/7c31388d80703f3643b4e11fc81fd32862162532 and also reverted there the fix from PR https://github.com/humhub/humhub/pull/4870/files

buddh4 (approved):

protected/humhub/modules/content/widgets/richtext/converter/BaseRichTextConverter.php
ArchBlood: Just an idea and could be ignored here if a better method could be implemented

```php
return parent::renderAbsy($blocks['content']);
```
or
```php
return rtrim(parent::renderAbsy($blocks['content']));
```

yurabakhtin: @buddh4 Do you agree to apply `rtrim()` there?

buddh4: No, please leave as is.

PR #4884, opened 2021-02-16
Implemented admin-permission to add users to spaces directly

jwleon:
It would be great to **allow users to add other users to space without an invitation**, but giving them the "ManageUsers" admin permission (where it is part of at the moment) is a huge deal. This would allow them to also edit and delete users, for example. I therefore **added a specific permission "Add Users to Spaces" as a sub-permission to it for just that purpose**. If you have "Manage Users" as permission, this will not introduce any changes. If you don't have "Manage Users" permission, this will neither introduce any changes, since the default for "Add Users to Spaces" is _deny_. So there is no harm done by giving the admin more freedom.

---

- [ ] Bugfix
- [x] Feature
- [x] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] All tests are passing
- [ ] Changelog was modified

- [x] A convincing reason for adding this feature

### DIFF
--- a/protected/humhub/modules/space/Module.php
+++ b/protected/humhub/modules/space/Module.php
@@ -42,10 +42,10 @@ class Module extends \humhub\components\Module
      */
     public $disableFollow = false;
 
-	/**
-	 * @var boolean defines if a space members can add anyone the the space without invitation
-	 * @since 1.8
-	 */
+    /**
+     * @var boolean defines if a space members can add anyone the the space without invitation
+     * @since 1.8
+     */
     public $membersCanAddWithoutInvite = false;
 
     /**
--- a/protected/humhub/modules/space/Module.php
+++ b/protected/humhub/modules/space/Module.php
@@ -42,6 +42,12 @@ class Module extends \humhub\components\Module
      */
     public $disableFollow = false;
 
+	/**
+	 * @var boolean defines if a space members can add anyone the the space without invitation
+	 * @since 1.8
+	 */
+	public $membersCanAddWithoutInvite = false;
+
     /**
      * @var int maximum space url length
      * @since 1.3
--- a/protected/humhub/modules/space/Module.php
+++ b/protected/humhub/modules/space/Module.php
@@ -46,7 +46,7 @@ class Module extends \humhub\components\Module
 	 * @var boolean defines if a space members can add anyone the the space without invitation
 	 * @since 1.8
 	 */
-	public $membersCanAddWithoutInvite = false;
+    public $membersCanAddWithoutInvite = false;
 
     /**
      * @var int maximum space url length
--- a/protected/humhub/modules/space/models/forms/InviteForm.php
+++ b/protected/humhub/modules/space/models/forms/InviteForm.php
@@ -116,7 +116,30 @@ class InviteForm extends Model
      */
     public function isQueuedJob()
     {
-        return ($this->withoutInvite || $this->allRegisteredUsers) && Yii::$app->user->can(ManageUsers::class);
+		// Pre-check if adding without invite / adding all members was requested
+		if (!($this->withoutInvite || $this->allRegisteredUsers)) {
+			return false;
+		}
+
+		// If user has permission to manage users, this action is allowed
+		if (Yii::$app->user->can(ManageUsers::class)) {
+			return true;
+		}
+
+		// Allow users to perform this action if this is allowed by config file
+		// Pre-check if user is member of the space in question
+		if (Yii::$app->getModule('space')->membersCanAddWithoutInvite === true) {
+			$membership = Membership::findOne([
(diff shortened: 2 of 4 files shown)


jwleon:
As a side-effect of generating the translation files, the generator changed the coding style of all translations for this module from `array()` to `[]`, this creating a huge amout of files in this commit. Sorry for that!

luke-:
@jwleon Thanks for your contribution. I'll discuss this with the team. Actually, we try to keep the number of permissions as clear as possible, but I can understand your use case well.

The translation  files are controlled by our translation community/system. I.e. they are not allowed to be part of the PR in order to avoid conflicts.

luke-:
@jwleon Before we add such a permission, we should generally start splitting the grouped admin permissions into more smaller ones. Adding only a single sub-permission in the current state probably makes it a bit unclear.

We are also thinking about making the option "Add people to a space without invitation" generally available to all members regardless of a permission (optional).

The potential for abuse is limited here. What do you think?

spoorun:
> @jwleon Before we add such a permission, we should generally start splitting the grouped admin permissions into more smaller ones. Adding only a single sub-permission in the current state probably makes it a bit unclear.
> 
> We are also thinking about making the option "Add people to a space without invitation" generally available to all members regardless of a permission (optional).
> 
> The potential for abuse is limited here. What do you think?

One can see the benefit, though we have Spaces with confidential information held and protected by GDPR, so in our case we really wouldn't like it, other groups/organisations using it may also have such concerns. So we'd suggest a more granular approach with separating the ability to add from the ability to add/remove, but being able to restrict it.

jwleon:
@luke- For my use-case, I would be totally fine if it is allowed globally, but I also understand that this might be a breaking change for many system implementations. One option would be to allow it by Module config (Module.php / config files) and have it disabled by default, so that there is no undesired change for existing systems.

In more general terms, I understand the approach to keep permissions easy to understand but think that reducing their number is not really a vital approach to go there. In my opinion it rises and falls with a clear system, presentation and explanation texts for each permission. I wrote another platform in PHP some time ago (not comparable in complexity by any means, it was way smaller) and spent a lot of time thinking about a permission system that is both versatile and logical. I think I have come to a pretty good solution, which I would like to sketch out here in case giving the permission system in HumHub and overhaul should come to the table at one point:

---

For me, the biggest problem was that fine-graining permissions automatically generates a type of tree, where the fine-grained permission would be a part of a larger permission, which would by itself be part of an even larger permission again, and so on. This would be confusing, intransparent and dangerous in implementation. So my solution was to note each and every child permission, no matter how far down the tree, in a "children list" of each permission. Example: If you were to obtain the permission "edit users", this permission would have "view users" and "view users basic infos" in the children list. "view users" would only have "view users basic infos" in it's children list.

If you obtain a permission, you also obtain all of its child permissions (because they are logically included in it). It is explecitely desired to obtain child permissions (like "view users basic infos" from many different other permissions. I then spent a lot of time on presentation so that it is clear which permissions were directly assigned and which permissions were obtained because they are part of another permission (see attached screenshot).

image

For explanation, because the idea differs a bit from HumHub; "roles" are like HumHubs "groups". The screenshot above is from the users account page and shows to the user what he/she is allowed to do and where permissions come from.  You can assign permissions to a role (=group), or directly to a user. A user permission overrules a role (=group) permission, which is why some are greyed out in the role-permission or child-permission section.

I am extremely happy with this approach, but would not in the slightest be able to develop it for a system like HumHub. Just wanted to give it in for inspiration purposes in case this topic is discussed in the future.

luke-:
This is really a very interesting solution and thanks for the short introduction. 

This would also be a nice solution for the HumHub Permission System as well, but of course we have to plan such changes in a longer term and currently there are few requests for this.

I think a config file option to show this permission would be good from the UI but probably gets lost in maintenance/testing.
For most business use cases, it's probably rather better if anyone can add anyone to a Space without invite.Misuse is rarely the case here.

Do you like to make this PR as an option or should I check our resources?

jwleon:
I totally understand. This was primarily meant as inspiration/input for the near or distant future because the permission system is such a fundamental aspect of a web-app (especially a social network web-app with a lot of people) and will probably grow and evolve with the whole app itself.

I will make the PR within the next week 👍

luke-:
@jwleon Ok great!

yurabakhtin (approved):

yurabakhtin:
@luke- I have reviewed and tested, it works well, but it is not a separate permission that can be changed per Group, it is an  option of the module "Space" that is applied for all users, if you think it is ok then it can be merged.

luke-:
@yurabakhtin Thanks for testing!
@jwleon Thanks for the contribution, works really well. I will adapt the changelog.

PR #5080, opened 2021-05-23
Enh: Show update tick on AdminMenu

ArchBlood:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] All tests are passing
- [ ] Changelog was modified

**Other information:**
Currently, the only way to actually see that there are any updates for modules is if you actually enter the Module tab, this p/r will help admins/those with the permissions to manage modules that there are updates for modules

Screenshot_1

### DIFF
--- a/protected/humhub/modules/admin/widgets/AdminMenu.php
+++ b/protected/humhub/modules/admin/widgets/AdminMenu.php
@@ -106,7 +106,7 @@ class AdminMenu extends LeftNavigation
 
         $this->addEntry(new MenuLink([
             'id' => 'modules',
-            'label' => Yii::t('AdminModule.base', 'Modules') .$updatesBadge,
+            'label' => Yii::t('AdminModule.base', 'Modules') . $updatesBadge,
             'url' => ['/admin/module'],
             'icon' => 'rocket',
             'sortOrder' => 500,
--- a/protected/humhub/modules/admin/widgets/AdminMenu.php
+++ b/protected/humhub/modules/admin/widgets/AdminMenu.php
@@ -89,9 +89,24 @@ class AdminMenu extends LeftNavigation
             ])
         ]));
 
+        /** @var Module $module */
+        $module = Yii::$app->getModule('marketplace');
+
+        if (!$module->enabled) {
+            return;
+        }
+
+        $updatesBadge = '';
+        $updatesCount = count($module->onlineModuleManager->getModuleUpdates());
+        if ($updatesCount > 0) {
+            $updatesBadge = '&nbsp;&nbsp;<span class="label label-danger">' . $updatesCount . '</span>';
+        } else {
+            $updatesBadge = '&nbsp;&nbsp;<span class="label label-default">0</span>';
+        }
+
         $this->addEntry(new MenuLink([
             'id' => 'modules',
-            'label' => Yii::t('AdminModule.base', 'Modules'),
+            'label' => Yii::t('AdminModule.base', 'Modules') .$updatesBadge,
             'url' => ['/admin/module'],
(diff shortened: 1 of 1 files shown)


luke-:
Thank you for this really good contribution.

I'm a bit concerned that in non-production mode with every admin page request the module update count calculation (incl. REST API request)  is executed. This is because the cache is not fully activated in this mode.

Can you do some tests here? Otherwise I would want to move the feature to Humhub 1.10. There are some module-specific things planned here anyway.

ArchBlood:
> Thank you for this really good contribution.
> 
> I'm a bit concerned that in non-production mode with every admin page request the module update count calculation (incl. REST API request) is executed. This is because the cache is not fully activated in this mode.
> 
> Can you do some tests here? Otherwise I would want to move the feature to Humhub 1.10. There are some module-specific things planned here anyway.

Sure thing, although I'm not seeing any performance issues on my end and in the network tab, I've seen all greens (Status 200, 302, 304) so I'm sure that no problems with the update counter calculation, if anything it works much faster in dev mode, I'm not currently able to do REST API requests on larger scales when it comes to this so it would be a good idea to draft this p/r and probably add it into the mix for 1.10 🤔

One issue that I did find is that updating the twofa module in dev mode shouldn't be recommended, no cache breaks the QR code visibility with 1.0.3

ArchBlood:
Overall, my thoughts are that on request it will prioritize what needed information in dev mode over collecting all information as a whole, calculating the number of an update is overall not very stressful on the calls it seems, updating of the module(s) with no cache will of course break some things (i.e external CSS/JS), if a better solution can be made I'm sure that it would have to start in 1.10

luke-:
@ArchBlood The problem are not the browser network accesses, but could be the requests from the HumHub server itself to our Marketplace Server. I'll mark the PR for 1.10. The beta for 1.9 will hopefully be released today or tomorrow.

ArchBlood:
After looking at the performance it doesn't seem to show much difference between either prod or dev
Screenshot_1 (1)
Screenshot_2
Screenshot_3

I will continue to monitor and report back on this if requested

protected/humhub/modules/admin/widgets/AdminMenu.php
```diff
@@ -89,9 +89,24 @@ public function init()
             ])
         ]));
 
+        /** @var Module $module */
+        $module = Yii::$app->getModule('marketplace');
+
+        if (!$module->enabled) {
```
luke-, line 95: Hides all further Admin Menu entries when Marketplace is disabled.

protected/humhub/modules/admin/widgets/AdminMenu.php
```diff
@@ -89,9 +89,24 @@ public function init()
...
+            return;
+        }
+
+        $updatesBadge = '';
+        $updatesCount = count($module->onlineModuleManager->getModuleUpdates());
+        if ($updatesCount > 0) {
+            $updatesBadge = '&nbsp;&nbsp;<span class="label label-danger">' . $updatesCount . '</span>';
+        } else {
+            $updatesBadge = '&nbsp;&nbsp;<span class="label label-default">0</span>';
```
luke-, line 104: I think we don't need a badge here when no updates available.

luke- (changes requested):
I've modified your PR a bit and moved it into a private method.

Here my commit:
https://github.com/humhub/humhub/commit/a5bb9e323661985e1bac11d741e9ceaf26d58b56

Thanks!

PR #5135, opened 2021-06-19
Remove pcntl check

dreua:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**


protected/humhub/libs/SelfTest.php
```diff
@@ -160,9 +160,10 @@ public static function getResults()
         }
 
         // Checks PCNTL Extension
-        $title = 'PHP - PCNTL Extension';
-
-        if (function_exists('pcntl_signal')) {
+	$title = 'PHP - PCNTL Extension';
+	// PCNTL may be disabled in webserver context but available in php-cli
+	// where it is actually needed.
+        if (function_exists('pcntl_signal') || shell_exec('php -r \'echo function_exists("pcntl_signal");\'') == '1') {
```
ArchBlood, line 166: To be honest `shell_exec()` can't be used by all systems in some shared hosts so I'd recommend first doing an `ini_get()` to check if permissions allow for it to be used.

dreua: Yeah, I was wondering whether that would be a problem, not sure how to reproduce and test this though: try/catch / ini_get / function_exists('shell_exec')? Maybe a small function with try/catch could be the most solid solution, what do you think?

ArchBlood: It would be more logical to generate a `try` and then `catch` for unexpected results that could happen

ArchBlood: As to check `ini_get('disable_functions')` should do to display disabled functions but not sure if this is too specific or not 🤔

luke-: Checking whether 'shell_exec' is available is apparently a bit more complicated.
https://stackoverflow.com/questions/21581560/php-how-to-know-if-server-allows-shell-exec

Important would be that no fatal error is triggered if the function is not available.

luke-:
Maybe we should simply remove the whole check. Especially in the web context the PCNTL extension should not be active.

dreua:
That would be fine with me, too. Otherwise I'd try to update this PR with a try/catch solution this weekend, let me know if I should do that.

luke-:
@dreua Please just remove the check. I think that causes too much confusion.

ArchBlood:
I can also agree, as it stands if you only have a check for `shell_exec()` then you'd also have to check all other alternatives to `shell_exec()` 🤔

dreua:
I removed the check, as requested. Let me know if you want a different entry in the changelog, I just tried to fit in with the style.

luke- (approved):

luke-:
@dreua Great, thank you!

Issue #5170, opened 2021-07-16
UserPickerField - Instant display of users


luke-:
### Is your feature request related to a problem? Please describe.

When typing the @ sign in a post or comment, it would be nice to directly see a list of possible and relevant users.

### Describe the solution you'd like

For example, in a post, this could be a list of the last 10 members active in the Space/ContentContainer.

For a comment, this could be e.g. the followers of the content. (The creator, other commenters, Liking Users).

yurabakhtin:
@luke- Do you mean the already existing feature:

<img width="666" alt="sign" src="https://user-images.githubusercontent.com/6995524/126488898-acd10b31-284a-4237-b703-8e609ca07d81.png">

If yes, is what you want, or should it modified somehow?

luke-:
The idea is to get rid of the "Please type at least 2 character" message and directly show options. 
As example see how it works on  Github.

![image](https://user-images.githubusercontent.com/4736168/126518811-5dda6671-c42e-4473-b296-4bebf4a5042e.png)

yurabakhtin:
@luke- We can quickly to decrease the 2 characters to `1` here https://github.com/humhub/humhub/blob/master/protected/humhub/widgets/CoreJsConfig.php#L110, but we cannot configure there to `0`.
But if we want to display a users list right after `@` I think we should modify the external plugin, probably here https://github.com/humhub/humhub-prosemirror/blob/master/src/editor/core/plugins/mention/provider.js, can I modify this library in order to allow `minInput  == 0`?

luke-:
@yurabakhtin The intention is not to reduce the characters for the search but to suggest/show relevant users directly without searching.

Yes, you can also edit & modify the library.

yurabakhtin:
@luke- Ok, I have modified the library in PR https://github.com/humhub/humhub-prosemirror/pull/63, after it will be merged we can merge core PR https://github.com/humhub/humhub/pull/5243.

luke-:
@yurabakhtin Thanks looks good so far.

Can we please reduce the list to max. 6 entries. (Would be great if this value can be overwritten using config file).
![image](https://user-images.githubusercontent.com/4736168/131505748-a9dd531e-36d5-45f7-a191-eeab55e0526b.png)

- For Posts in Spaces: It would be good if we could always display the last 6 active Space members (space_membership.last_visit - db index?).
 - For comments: Initially display max. 6 users who are following the Content (follow table).

yurabakhtin:
[...]

PR #5243, opened 2021-08-13
Allow 0 for min input size

yurabakhtin:

- [x] Feature

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
Issue: https://github.com/humhub/humhub/issues/5170
**Note:** It will works after merging PR https://github.com/humhub/humhub-prosemirror/pull/63


protected/humhub/modules/search/controllers/MentioningController.php
```diff
@@ -33,37 +43,167 @@ public function getAccessRules()
...
+        return $this->asJson($results);
+    }
+
+    /**
+     * Find space members on mentioning request from RichText editor on Post form
+     *
+     * @return \yii\web\Response
+     * @throws HttpException
+     */
+    public function actionPostSpaceMembers()
```
luke-, line 79: Can we please rename to `actionSpace` so `/search/mentioning/space?id=4` .

protected/humhub/modules/search/controllers/MentioningController.php
```diff
@@ -33,37 +43,167 @@ public function getAccessRules()
...
-                'image' => SpaceImage::widget(['space' => $container, 'width' => 20]),
-                'link' => $container->getUrl()
-            ];
+        $space = Space::findOne(['id' => $spaceId]);
+        if (!$space || !$space->can(CreatePost::class)) {
+            throw new HttpException(403, 'Access denied!');
+        }
+
+        // Find space members
+        $users = User::find()
```
luke-, line 90: Can you please try to use `Membership::getSpaceMembersQuery` here.

protected/humhub/modules/search/controllers/MentioningController.php
```diff
@@ -33,37 +43,167 @@ public function getAccessRules()
...
 
+    /**
+     * Find users followed to the Content on mentioning request from RichText editor on Comment form
+     *
+     * @return \yii\web\Response
+     * @throws HttpException
+     * @throws \yii\base\Exception
+     * @throws \yii\base\InvalidConfigException
+     */
+    public function actionCommentContentFollowers()
```
luke-, line 118: Can you please try to rename to `actionContent` and pass `content.id` to it?  e.g. `/search/mentioning/content?id=4&keyword...`

protected/humhub/modules/search/controllers/MentioningController.php
```diff
@@ -33,37 +43,167 @@ public function getAccessRules()
...
+        $module = Yii::$app->getModule('comment');
+
+        if (!class_exists($modelClass) ||
+            !($object = $modelClass::findOne(['id' => $modelId])) ||
+            !$module->canComment($object)) {
+            throw new HttpException(403, 'Access denied!');
+        }
+
+        // Find users followed to the Content
+        $users = User::find()
```
luke-, line 134: Would be great if we can use `Followable::getFollowers` here.

protected/humhub/modules/search/controllers/MentioningController.php
```diff
@@ -33,37 +43,167 @@ public function getAccessRules()
...
+    public function actionCommentContentFollowers()
+    {
+        $modelClass = (string)Yii::$app->request->get('model');
+        $modelId = (int)Yii::$app->request->get('id');
+        $keyword = (string)Yii::$app->request->get('keyword');
+
+        /** @var \humhub\modules\comment\Module $module */
+        $module = Yii::$app->getModule('comment');
+
+        if (!class_exists($modelClass) ||
```
luke-, line 127: If we use `content.id` as param, it should be easier to instance the corresponding `ContentActiveRecord` (followable).

$content = humhub\modules\content\models\Content
e.g. `$post = $content->getPolymorophicRelaltion'

luke- (changes requested):

protected/humhub/modules/search/controllers/MentioningController.php
yurabakhtin: Ok, I changed, but please note I named firstly this as `PostSpaceMembers` in order to inform the action is used only for Post form side, because inside the action I added a check permission `$space->can(CreatePost::class)`, so it is not recommend to use this action from some other place in future.

protected/humhub/modules/search/controllers/MentioningController.php
yurabakhtin: Done, but only one difference is the `Membership::getSpaceMembersQuery` uses `User::active()` but in previous version we used `User::visible()`, the methods do almost same except the `User::visible()` has an additional event `EVENT_CHECK_VISIBILITY`.

protected/humhub/modules/search/controllers/MentioningController.php
yurabakhtin: I have renamed, but I used the long action name for same reason what I described here https://github.com/humhub/humhub/pull/5243#discussion_r701928224.

protected/humhub/modules/search/controllers/MentioningController.php
yurabakhtin: Done with `Follow::getFollowersQuery()`. This query the same differnce as I wrote here https://github.com/humhub/humhub/pull/5243#discussion_r701929951.

protected/humhub/modules/search/controllers/MentioningController.php
yurabakhtin: Done.

protected/humhub/modules/search/controllers/MentioningController.php
luke-: Ok, thank you for the hint. We can change the permission here in the future so that e.g. all those who have access to Space can search here.

protected/humhub/modules/search/controllers/MentioningController.php
luke-: @yurabakhtin Ok, we may need to change it also to `visible()` in the Membership Query.

protected/humhub/modules/search/controllers/MentioningController.php
luke-: Ok thanks. We may allow all users with read access to the content in future -  instead of CanComment permission.

protected/humhub/modules/search/controllers/MentioningController.php
luke-: @yurabakhtin Ok thanks, if we have a wide use here we can also change the "canComment" to a content `canView` here. But currently I would leave it as it is for now.

protected/humhub/modules/search/controllers/MentioningController.php
```diff
@@ -33,37 +46,157 @@ public function getAccessRules()
...
-            $results[] = [
-                'guid' => $container->guid,
-                'type' => 's',
-                'name' => $container->getDisplayName(),
-                'image' => SpaceImage::widget(['space' => $container, 'width' => 20]),
-                'link' => $container->getUrl()
-            ];
+        $space = Space::findOne(['id' => $spaceId]);
+        if (!$space || !$space->can(CreatePost::class)) {
+            throw new HttpException(403, 'Access denied!');
```
ArchBlood, line 89: Use of `Yii::t('SearchModule.error', 'Access denied!')` as it would not be translatable and issues would be created to make it translatable later, that is of course if @luke- doesn't see this to be necessary?

protected/humhub/modules/search/controllers/MentioningController.php
```diff
@@ -33,37 +46,157 @@ public function getAccessRules()
...
+        $contentId = (int)Yii::$app->request->get('id');
+        $keyword = (string)Yii::$app->request->get('keyword');
+
+        /* @var CommentModule $commentModule */
+        $commentModule = Yii::$app->getModule('comment');
+
+        if (!($content = Content::findOne(['id' => $contentId])) ||
+            !($object = $content->getModel()) ||
+            !$commentModule->canComment($object)) {
+            throw new HttpException(403, 'Access denied!');
```
ArchBlood, line 128: Same as https://github.com/humhub/humhub/pull/5243/files#r702449959

luke- (approved):

protected/humhub/modules/search/controllers/MentioningController.php
luke-: @ArchBlood Thank you, but in my opinion it is not necessary. Actually, 'throw HttpException(403)` should also be sufficient. Especially with such internal errors that should properly be caught via the frontend.

PR #5250, opened 2021-08-18
Allow to view and edit text files

yurabakhtin:

- [x] Feature

- [x] All tests are passing
- [x] Changelog was modified

**Other information:**
Issue: https://github.com/humhub/cfiles/issues/122

### DIFF
--- a/protected/humhub/modules/file/models/File.php
+++ b/protected/humhub/modules/file/models/File.php
@@ -189,12 +189,29 @@ class File extends FileCompat
     }
 
     /**
-     * Checks if given file can deleted.
+     * Checks if given file can be deleted.
      *
      * If the file is not an instance of ContentActiveRecord or ContentAddonActiveRecord
      * the file is readable for all unless there is method canEdit or canDelete implemented.
+     *
+     * @param int $userId
+     * @return bool
      */
     public function canDelete($userId = null)
+    {
+        return $this->canEdit($userId);
+    }
+
+    /**
+     * Checks if given file can be edited.
+     *
+     * If the file is not an instance of ContentActiveRecord or ContentAddonActiveRecord
+     * the file is readable for all unless there is method canEdit or canDelete implemented.
+     *
+     * @param int $userId
+     * @return bool
+     */
+    public function canEdit($userId = null)
     {
         $object = $this->getPolymorphicRelation();
 
--- a/CHANGELOG_DEV.md
+++ b/CHANGELOG_DEV.md
@@ -10,3 +10,4 @@
 - Enh #5216: Separate view document button
 - Enh #5229: Use RichTextField for user approval messages
 - Enh #100: Allow additional toggler for elements with context menu
+- Enh #122: Allow to view and edit text files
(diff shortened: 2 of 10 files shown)


protected/humhub/modules/file/models/File.php
```diff
@@ -189,12 +189,29 @@ public function canRead($userId = "")
...
      *
      * If the file is not an instance of ContentActiveRecord or ContentAddonActiveRecord
      * the file is readable for all unless there is method canEdit or canDelete implemented.
+     *
+     * @param int $userId
+     * @return bool
      */
     public function canDelete($userId = null)
+    {
+        return $this->canEdit($userId);
```
ArchBlood, line 202: Why return `canEdit()` if it's still returning `getPolymorphicRelation()` regardless, I'm also curious to why `canEdit()` should be used for `canDelete()` at all? Just slightly confused and would like validation before trying out some tests😄

ArchBlood: I only ask because `canEdit()` and `canDelete()` should do completely different things, `canEdit()` should be a checker if a file can be edited, while `canDelete()` should be a checker if the file can be deleted, with how it's implemented here it just confuses me looking at the code itself 😅

yurabakhtin: @ArchBlood Thanks for the review, please note I just made a new method `File::canEdit()` as copy of the already existing method `File::canDelete()` in order to keep better call like `if ($file->canEdit()) { ... }` from controller code for example.
I.e. yes these 2 methods have same code, but I am not sure we have a separate permissions: one for edit and one for delete, usually if user can edit/update/modify an object then such user can delete the object.

Probably we should use another name like `canManage` or `canModify`, so code will be like:

```
function canManage()
{
    // All generic code here
}

function canEdit()
{
    return $this->canManage();
}

function canDelete()
{
    return $this->canManage();
}
```

luke-:
We should work towards the direction that in future `File` models (core file module, from the `files` table) are not longer modified. Instead we should create new `File` models on updates.

Then it would be easier to cache them, to store them on CDNs or to implement file versioning.

Of course this makes the FileEditHandler a bit more complicated, because it actually has to create new Files.

The following scenarios exist:

- If a `file` (e.g. Word document) which is attached to a `Post` and is edited, the `File`model is simply exchanged and the older `File` model record will deleted. 
- In the `CFile` module there should be a versioning soon, so there are multiple `File` model records for a `CFile file` model. https://github.com/humhub/cfiles/issues/121

yurabakhtin:
@luke- 

> We should work towards the direction that in future `File` models (core file module, from the `files` table) are not longer modified. Instead we should create new `File` models on updates.

Should I modify the PR to create new model like `FileContent` and move all changes what I did inside model `File`?
Even I cannot use the name `FileContent` because it already exists in core, so I should select another name like `FileUpdate` or `PhysicalFile` or `DiskFile`, right?

ArchBlood:
> @luke- 
> 
> > We should work towards the direction that in future `File` models (core file module, from the `files` table) are not longer modified. Instead we should create new `File` models on updates.
> 
> Should I modify the PR to create new model like `FileContent` and move all changes what I did inside model `File`?
> Even I cannot use the name `FileContent` because it already exists in core, so I should select another name like `FileUpdate` or `PhysicalFile` or `DiskFile`, right?

`FileUpdate` seems more logical in this regard so I'd go with that then move the changes from `File` there and extend the `File` model if need be?

luke-:
Moved: https://github.com/humhub/text-editor

PR #5366, opened 2021-10-19
Fix file attaching for objects without support file versioning

yurabakhtin:

- [x] Bugfix
- [x] Feature

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**
Issue: https://github.com/humhub/text-editor/issues/9#issuecomment-946609828

### DIFF
--- a/protected/humhub/modules/file/Events.php
+++ b/protected/humhub/modules/file/Events.php
@@ -16,7 +16,7 @@ use humhub\modules\file\converter\TextConverter;
 
 /**
  * Events provides callbacks to handle events.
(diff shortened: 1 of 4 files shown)


yurabakhtin:
@luke- 

> As far as I know, this is no longer necessary or?

Yes, it was developed for temp version when we created new record in the table `file` on each file update, but currently we have the separate table `file_history`, so this PR should be closed without merging.

luke-:
@yurabakhtin Thanks for the quick reply!

Issue #5398, opened 2021-10-27
installation error: 1.8.latest > 1.9.2


hj-beckers:
### What steps will reproduce the problem?
installation from 1.8.latest to 1.9.2 Please see the full thread in the community forum (https://community.humhub.com/s/installation-and-setup/?contentId=247458).
### What is the expected result?
a working 1.9.2 installation
### What do you get instead?
The errors as stated in the community forum
### Additional info

| Q                | A
| ---------------- | ---
| HumHub version   | 1.8.5?
| PHP version      |  7.4.6
| Operating system | opensuse 15.2

> Note: Also provide the full error stacks from `Administration --> Information --> Logging` and errors from your browser console if possible.
> Note: Please use the [community](https://community.humhub.com) site for any questions on customizing HumHub.

hj-beckers:
No help, no tipps, no answers?

yurabakhtin:
@hj-beckers @luke- Sorry, it was indeed my bug, because here https://github.com/humhub/humhub/commit/ea0757ab747526bf5d8b44d2efa944fc18313133#diff-165768d5264abf499f64dfebc8457418231c40e499429732ec7080f5695eab96R112 I used a static name on my local database.
I have fixed this in PR https://github.com/humhub/humhub/pull/5416.

luke-:
@yurabakhtin Thanks.

@hj-beckers Will be fixed with 1.9.3. Until you can manually replace following files:
https://github.com/humhub/humhub/pull/5416/files

hj-beckers:
I patched the files and executed according to https://docs.humhub.org/docs/admin/updating the steps Flush caches, Database migration and Update installed modules with success. But restarting humhub yields the following error:

Not instantiable – yii\di\NotInstantiableException
Failed to instantiate component or class "humhub\modules\questionanswer\models\Comment".
↵
Caused by: ReflectionException
Class humhub\modules\questionanswer\models\Comment does not exist

in /srv/www/htdocs/humhub/protected/vendor/yiisoft/yii2/di/Container.php at line 508

Background: During the update from 1.8.2 -> 1.8.5 I got an error which IIRC mentioned the above modul. This was solved by restoring 1.8.2 instance and DB, removing some not used modules and restarting the 1.8.5 update, which succeded.

hj-beckers:
Do I have to open a new issue regarding the new problem or do you reopen issue #5398?

Thanks for your help.
Hermann-Josef Beckers

[...]

PR #5416, opened 2021-11-02
Fix checking of foreign table index on install "contentcontainer_tag"

yurabakhtin:

- [x] Bugfix

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
Issue: https://github.com/humhub/humhub/issues/5398

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ HumHub Changelog
 1.9.3 (Unreleased)
 ------------------
 - Fix #5372: Text shortening in Japanese broken - Fix truncating of multi-byte strings
+- Fix #5398: Fix checking of foreign table index
 
 
 1.9.2 (October 15, 2021)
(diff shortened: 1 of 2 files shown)


luke- (approved):

hj-beckers:
I patched the files and executed according to https://docs.humhub.org/docs/admin/updating the steps Flush caches, Database migration and Update installed modules with success. But restarting humhub yields the following error:

Not instantiable – yii\di\NotInstantiableException
Failed to instantiate component or class "humhub\modules\questionanswer\models\Comment".
↵
Caused by: ReflectionException
Class humhub\modules\questionanswer\models\Comment does not exist

in /srv/www/htdocs/humhub/protected/vendor/yiisoft/yii2/di/Container.php at line 508

Background: During the update from 1.8.2 -> 1.8.5 I got an error which IIRC mentioned the above modul. This was solved by restoring 1.8.2 instance and DB, removing some not used modules and restarting the 1.8.5 update, which succeded.

Issue #5534, opened 2022-02-04
Statistic Input Loading Problem


luke-:
![image](https://user-images.githubusercontent.com/4736168/152610894-f155361a-163f-4daa-875e-f9be223b3481.png)

PR #5539, opened 2022-02-06
Statistic input loading problem

s-tyshchenko:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] All tests are passing
- [x] Changelog was modified

**Other information:**
Issue: https://github.com/humhub/humhub/issues/5534


luke- (changes requested):
@s-tyshchenko Unfortunately, this does not solve the problem. We also prefer "::class".

After a page reload the editor always works. The problem only occurs when navigating from another page to the statistics page (PJAX).

protected/humhub/modules/admin/views/setting/statistic.php
```diff
@@ -17,13 +17,15 @@
 <?= $form->errorSummary($model); ?>
 
 <div class="form-group">
-    <?= $form->field($model, 'trackingHtmlCode')->widget(CodeMirrorInputWidget::class); ?>
+    <?= $form->field($model, 'trackingHtmlCode')->widget(CodeMirrorInputWidget::className()); ?>
```
ArchBlood, line 20: This does not fix the issue causing the loading, not to mention is working backward with coding standards when using PHP 7.2+

Looking it over the issue seems to stem from the CodeMirrorInputWidget widget loading required scripts and CSS.

protected/humhub/modules/admin/views/setting/statistic.php
```diff
@@ -17,7 +19,7 @@
 <?= $form->errorSummary($model); ?>
 
 <div class="form-group">
-    <?= $form->field($model, 'trackingHtmlCode')->widget(CodeMirrorInputWidget::className()); ?>
+    <?= $form->field($model, 'trackingHtmlCode')->widget(CodeMirrorInputWidget::class); ?>
```
luke-, line 22: We should include the below javascript in the InputWidget class so that it doesn't have to be added separately for each view.

protected/humhub/modules/admin/views/setting/statistic.php
```diff
@@ -28,4 +30,32 @@
 
 <?php ActiveForm::end(); ?>
 
+<script>
```
luke-, line 33: We need to add 'Html::nonce()` here.Please see other Javascripts....

protected/humhub/modules/admin/views/setting/statistic.php
```diff
@@ -28,4 +30,32 @@
...
+
+    let initCodeMirrorTimeout = setTimeout(initCodeMirror, 200);
```
luke-, line 35: Hm, I would be very uncomfortable working with the timeouts.

Maybe this event do the trick?
https://github.com/humhub/humhub/blob/master/static/js/humhub/humhub.client.pjax.js#L44

luke- (changes requested):

static/js/humhub/humhub.ui.codemirror.js
```diff
@@ -8,21 +8,34 @@ humhub.module('ui.codemirror', function(module, require, $) {
 
     var init = function () {
         event.on('humhub:ready', function (evt) {
```
luke-, line 10: @s-tyshchenko Did you tried this event here?
https://github.com/humhub/humhub/blob/master/static/js/humhub/humhub.client.pjax.js#L44

luke- (changes requested):
@s-tyshchenko I would like to avoid the timeout solution. Can you take a look at my comment?

luke-:
@s-tyshchenko Okay, let's leave it that way for now, even if the solution is not quite optimal.

Can you please check why the tests fail?

luke- (approved):

PR #5555, opened 2022-02-13
Improved support of the emailRequired property

### LINKED ISSUE #5553: Cannot have two or more users without email when emailRequired is disabled
1. Disable emailRequired in common.php for the user module.
2. Create more than one user without email in the admin menu.
3. (Optional) Edit a user and remove the email --> will not save

Also have more than 2 users without email

### Additional info
I will upload the fix in a moment :)

| Q                | A
| ---------------- | ---
| HumHub version   | since 1.2.2?
| PHP version      | 
| Operating system |

KarlKeu00:

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

- [X] It's submitted to the `develop` branch, _not_ the `master` branch if no hotfix
- [X] When resolving a specific issue, it's referenced in the PR's description (e.g. `Fix #xxx[,#xxx]`, where "xxx" is the Github issue number)
- [ ] All tests are passing
- [X] Changelog was modified

**Other information:**
- Fix #5553 
- Since in the database the email row is unique (and better should remain so), a random invalid email is generated for each user who has not stored an email.
- The email is guaranteed to be invalid because the TLD has been reserved (https://en.wikipedia.org/wiki/.invalid).
- Sorry for #5554, I'm still practicing a bit to deal with GitHub :)


KarlKeu00:
Oh, right. I must have overlooked that.
But still there is the error that for an existing user the email can not be removed. I think the problem is that the HForm supports "required" but not in combination with "when". So the changes in User.php would still have to be partially applied.

luke-:
@KarlKeu00  Yes ok, the fix regarding HForm is good!

Maybe it also makes sense to add a "beforeSave" and make sure that empty email values are always null?

e.g.
```php
if ($this->email == '') {
    $this->email = new \yii\db\Expression('NULL');
}
```

KarlKeu00:
My understanding until now was that MySQL normally replaces an empty string with NULL. This makes total sense from a technical point of view: An empty string consists only of one \0 "character" = NULL. But I don't know how this looks like with other or older databases. But this additional protection certainly helps to avoid errors.

KarlKeu00:
Just implemented the changes.
In the BaseFormAuth.php I have kept parts of the change anyway (more readable)
In User.php the line 12 was not used and could be removed.

KarlKeu00:
Just implemented the changes.
Sorry for the many commits - still have to practice with GitHub and the collaboration feature...

luke- (approved):

luke-:
@KarlKeu00 Thank you and no problem. I did a squash commit. 
https://github.com/humhub/humhub/commit/cb000a1f72e95aeb80969b1444c9544c95cc5861

PR #5573, opened 2022-02-21
Allow replying for sub comments

yurabakhtin:

- [x] Bugfix

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
Issue: https://github.com/humhub/team_tasks/issues/200

### DIFF
--- a/CHANGELOG_DEV.md
+++ b/CHANGELOG_DEV.md
@@ -23,3 +23,4 @@
 - Fix #5564: Close modal button doesn't work after form validation
 - Fix #5533: Users can't live in UTC
 - Fix #5460: Untranslatable string New Updates in Stream
+- Fix #5573: Allow replying for sub comments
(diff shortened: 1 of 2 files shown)


luke- (approved):

Gilbertdelyon:
I just tested this fix in HH1.11 and it's confusing if you are not familiar with HH**.

1/ When you click the "reply" link of a sub-comment you think you will add a new sub-sub-comment. Instead of this you only add a new sub-comment at the end of the list. This is confusing if (like me) you come from a CMS where you can set as many sub levels as you want.

2/ If there are many sub-comments, the input area is rejected at the bottom and can be out of the screen  (to low). So, if you click the "reply" link of the first sub-comment it opens the input area and fill it with author name but you dont see it and you think you clicked a dead link. 

** I just come from a Drupal developpement. In this CMS you can set as many comment sub levels  as you want. When you click the reply link of any comment or sub-sub...-comment  it opens a new page with a display of the comment you want to answer and the input area for your new sub-comment. My idea is that it's much more convenient.

Issue #5522, opened 2022-02-01
File Table - Add Content Id


luke-:
It would be good if a `content_id` field is available in the file table. 

Even if the file is only indirectly related to a content (e.g. it belongs to a comment) it would be good if the `content_id` is set here anyway.

Use cases for this are statistics, e.g. how much memory does a space consume.

yurabakhtin:
@luke- PR https://github.com/humhub/humhub/pull/5583

PR #5583, opened 2022-03-02
Add content id in the file table

yurabakhtin:

- [x] Feature

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
Issue: https://github.com/humhub/humhub/issues/5522


luke- (changes requested):
@yurabakhtin Thanks, can you please check if we get this set somehow in files of comments?

yurabakhtin:
@luke- I found a quick solution for the bug above: if I revert the commit https://github.com/humhub/humhub/commit/18ec9a3 , do you remember why you did this?

Currently the problem in that if we call there `$this->save()` then the code `File->beforeNewStoredFile()` is called twice where the code line `$this->store->delete(null, [FileHistory::VARIANT_PREFIX . '*']);` deletes the file what is uploaded on first call of the `File->save()`.

It seems you wanted to update some info in the table `file` that is updated only on call the `File->save()`.

luke-:
@yurabakhtin - I made the change to update columns like "updated_at" and "updated_by". Do you think I should find another solution here?

- About comments: When a file is added to a comment, the "content_id" should be set here as well. A comment is always assigned to a content. (The migration should also take this into account)

yurabakhtin:
@luke- 

> I made the change to update columns like "updated_at" and "updated_by". Do you think I should find another solution here?

It seems you didn't see my updated text in the comment above:

> I fixed the bug in PR https://github.com/humhub/humhub/pull/5593 without touching your code, so all columns are updated as before.

So you shouldn't find a solution, it was already done, i.e. the fields `updated_at` and `updated_by` are updated as expected.

luke-:
@yurabakhtin Thanks! Can you please check the migration as well? That existing comment files are also assigned to a content ID?

yurabakhtin:
@luke- 

> Can you please check the migration as well? That existing comment files are also assigned to a content ID?

Done in commit https://github.com/humhub/humhub/pull/5583/commits/e12c30373b7f08dc1ba50826518d56c1cb8950d4.

luke- (approved):

Issue #5716, opened 2022-05-21
Saving oEmbed Data


ArchBlood:
### What steps will reproduce the problem?
On any of the current pre-installed or custom oEmbeds click the save button and you'll see that the provider endpoint `%url%` will decode all `%` to `%25` on save.

Example: `https://www.youtube.com/oembed?url=%url%&format=json` > `https://www.youtube.com/oembed?url=%25url%25&format=json`

### What is the expected result?
The transferred data should not be decoding the `%url%` in such a way.

### What do you get instead?
![Screenshot_1](https://user-images.githubusercontent.com/35392110/169667611-e380bf2a-7a0d-48de-8fa8-6b1e342f0be7.png)
![Screenshot_2](https://user-images.githubusercontent.com/35392110/169667612-c9fa91e7-7aac-4a47-8316-c510f2453c51.png)

### Additional info

| Q                | A
| ---------------- | ---
| HumHub version   | 1.11+
| PHP version      | 7.4
| Operating system | CloudLinux & [HumHub Hosting Plans](https://willywonka.humhub.com/dashboard)

> Note: Also provide the full error stacks from `Administration --> Information --> Logging` and errors from your browser console if possible.
> Note: Please use the [community](https://community.humhub.com) site for any questions on customizing HumHub.

spunkram:
I hope the problem is solved immediately.

ArchBlood:
> this problem should be solved with an update , it is not healthy for us to produce manual solutions !

The solution is not a manual implementation, this will be in the next release.

yurabakhtin:
Yes, it is not manual solution, however if someone need it, then the real fix is in the single line https://github.com/humhub/humhub/pull/5720/files#diff-550160e418aa56e463e5eae9d6955ed1faefb2d1fa9755782f9690d7327586fcR52 just wrap by `decodeURI()`.

PR #5720, opened 2022-05-24
Fix saving of OEmbed provider endpoint URL

yurabakhtin:

- [x] Bugfix

- [x] All tests are passing
- [x] Changelog was modified

**Other information:**
Issue: https://github.com/humhub/humhub/issues/5716

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ HumHub Changelog
 - Fix #5693: Scroll to comment on single view
 - Fix #5709: Fix input size for placeholder length in picker field
 - Fix #5704: Don't send notification for non-member of private space
-- 
+- Fix #5716: Fix saving of OEmbed provider endpoint URL
 
 1.11.1 (April 22, 2022)
 -----------------------
--- a/protected/humhub/modules/admin/controllers/SettingController.php
+++ b/protected/humhub/modules/admin/controllers/SettingController.php
@@ -378,14 +378,10 @@ class SettingController extends Controller
             ];
             UrlOembed::setProviders($providers);
 
-            return $this->redirect(
-                [
-                    '/admin/setting/oembed'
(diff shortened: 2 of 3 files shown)


ArchBlood (approved):
Looks good for me so far, although I'm not sure why I'm still getting the following error?

Screenshot_1

yurabakhtin:
@ArchBlood It seems issue from your php editor, because I don't see this in PhpStorm from my side:

phpstorm

You can try to delete the end of line and create it again, probably there is some hidden char in the end of line.

ArchBlood:
> @ArchBlood It seems issue from your php editor, because I don't see this in PhpStorm from my side:
> 
> phpstorm
> 
> You can try to delete the end of line and create it again, probably there is some hidden char in the end of line.

My issue seems to be any entry of `$query[]` and any variation of the array and variable combo, of course, the PHP editor is set to PHP 7.4 and nothing else. :thinking:

yurabakhtin:
@ArchBlood Yes, strange, just try to remove the line and one below completely and retype them both char by char manually, because your editor finds a char `$end-of-file` somewhere there, it is hidden for our eyes, otherwise I don't have any idea more.

ArchBlood:
> @ArchBlood Hmm, for me there is also no error in the IDE.

I'm now understanding why this is the case after reading more into some of the explanations found here;

https://stackoverflow.com/questions/47540423/how-does-isset-not-throw-an-error-when-evaluating-array-properties

So simply put we're using a non-existing variable or something like this, which is probably why most IDE don't read this but built-in server editors do? I'm not completely sure myself.

https://stackoverflow.com/questions/1960509/how-to-avoid-isset-and-empty

After testing with Atom.io I also don't see any errors but still leaves me with questions that I have yet to find any answers to. 🤔

Screenshot_2

PR #5779, opened 2022-06-28
Fix user settings updating on disabled visibility "Guest"

yurabakhtin:

- [x] Bugfix

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**
Fix a not saving user settings when general setting "Allow visitors limited access to content without an account (Adds visibility: "Guest")" is disabled.


luke-:
@yurabakhtin What about the visibility "Hidden"? This should not be able to be changed by a user. Unless he is admin.
But it would be also important that "Hidden" is otherwise not allowed in the rules.

yurabakhtin:
@luke- 

> What about the visibility "Hidden"? This should not be able to be changed by a user. Unless he is admin.
But it would be also important that "Hidden" is otherwise not allowed in the rules.

I think it works as you wrote, or do you get any issue there?

The form `humhub\modules\user\models\forms\AccountSettings` doesn't allow to set the "Hidden" option by the rule:

```php
['visibility', 'in', 'range' => array_keys(User::getVisibilityOptions(false)),
```

see on the param value `false`:

```
public static function getVisibilityOptions($allowHidden = true): array
{
...
        if ($allowHidden) {
            $options[self::VISIBILITY_HIDDEN] = Yii::t('AdminModule.user', 'Invisible');
        }
...
```

but from the administration side we allow to set the "Hidden" option because there is a rule with `true` value for the param.

---

Please compare and let me know if something is wrong:

User Profile edit page:

user-profile-settings

Administration user edit form:

admin_user_edit_form

luke-:
@yurabakhtin Following steps:
- Login as Admin, make sure Guest Access is enabled
- Mark some user as Invisible
- If this user is going to change Profile Settings e.g. Language or Tags - The user is not longer Invisible.


For users which are Invisible we should show this option

yurabakhtin:
@luke- Yes, I can reproduce this, because we allow to edit "Visibility" for users when Guest Access is enabled, but for users we don't display the option "Hidden" so it is update to first selected.
So what should we do in such case when currently user's visibility is hidden?
Maybe we should deny to edit field at all from user side? I.e. the user will can edit his visibility only after admin changes it to not hidden option, right?

luke-:
@yurabakhtin Yes, I like that, deny editing would be a good option to solve the problem if Visibility is on Hidden!

Alternatively, you could also allow "Hidden" as an option if the user is currently on Hidden. But I like the first option better.

yurabakhtin:
@luke- https://github.com/humhub/humhub/pull/5779/commits/1883b86b5e220c6e006abf9286f19b1cce314774 - I have implemented to display the disabled visibility option for simple users (non admin), and after save it is not resetted:

user-noadmin-visibility-hidden

For admins the field is allowed without restriction because they can edit it from administration:

admin-visibility-hidden

luke- (approved):

PR #5799, opened 2022-07-06
Fix filter for select2 fields on "People" page

yurabakhtin:

- [x] Bugfix

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**
Bug: on "People" page, if change some filter with picker feature then all other picker filters are changed to wrong value, e.g. if we have the selected filter:

- First name: Sara
- City: [empty]

and change the filter City to "Friedland", then after page reloading we see that filter "First name" is changed to `firstname` instead of previous and expected value `Sara`, but the "City" looks correctly as was selected.


luke-:
@yurabakhtin Thanks good catch!

PR #5827, opened 2022-08-03
Layout for wall create content form menu

yurabakhtin:

- [x] Feature

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
Issue: https://github.com/humhub/team_tasks/issues/239


protected/humhub/modules/content/components/ContentActiveRecord.php
```diff
@@ -93,6 +94,19 @@ class ContentActiveRecord extends ActiveRecord implements ContentOwner, Movable
...
+     *
+     *  - BasePermission class string
+     *  - Array of type ['class' => '...', 'callback' => '...']
+     *  - Anonymous function
+     *  - BasePermission instance
+     *
+     * @var string permission instance
+     * @since 1.13.0
+     */
+    protected $createPermission = CreatePublicContent::class;
```
luke-, line 108: Maybe `ManageContent` permission as default?

protected/humhub/modules/content/components/ContentContainerModule.php
```diff
@@ -219,4 +219,15 @@ protected function getGlobalPermissions()
...
 
+    /**
+     * This method is called to determine classes of Content models which can be posted on wall.
+     *
+     * @param ContentContainerActiveRecord|null $contentContainer
+     * @return array
+     */
+    public function getContentClasses(?ContentContainerActiveRecord $contentContainer = null): array
+    {
+        return [];
```
luke-, line 230: Consideration: Would it perhaps be better to return an empty e.g. `Post` object here instead of just the class name as string. 

This way one could also cover cases with "dynamic" content objects.
 
Besides, we then have to instantiate anyway to get the name/canEdit/icon/etc.

protected/humhub/modules/content/components/ContentContainerModule.php
```diff
@@ -219,4 +219,15 @@ protected function getGlobalPermissions()
         return [];
     }
 
+    /**
+     * This method is called to determine classes of Content models which can be posted on wall.
+     *
+     * @param ContentContainerActiveRecord|null $contentContainer
+     * @return array
+     */
+    public function getContentClasses(?ContentContainerActiveRecord $contentContainer = null): array
```
luke-, line 228: I am a bit unsure if we should pass the `ContentContainer` as a parameter here.

Would each module then have to check itself whether it is activated in the respective ContentContainer?

For (see next comment) the CC would be handy of course.

protected/humhub/modules/content/widgets/stream/WallStreamEntryWidget.php
```diff
@@ -139,6 +145,16 @@ abstract class WallStreamEntryWidget extends StreamEntryWidget
      */
     protected $renderOptionClass = WallStreamEntryOptions::class;
 
+    /**
+     * @var string
+     */
+    public $menuAction;
```
luke-, line 151: Add Docs. When action for creation, please add word `createForm`

protected/humhub/modules/post/controllers/PostController.php
```diff
@@ -101,4 +103,16 @@ public function actionEdit($id)
         ]);
     }
 
+    public function actionForm()
+    {
+        if (!$this->contentContainer->getPermissionManager()->can(CreatePost::class)) {
```
luke-, line 108: Why not use new style?  `if ((new Post($this->contentContainer))->canEdit())`

luke- (changes requested):
I like it very much! Good job!

- Old legacy modules that are not yet  using  "ContentActiveRecord::$createPermission" and or "WallStreamModuleEntryWidget::createRoute" and or "Module::getContentClasses" should still working (no fatal error) 
- Refactor: We should everywhere use `canEdit` instead of old style: `https://github.com/humhub/humhub/blob/master/protected/humhub/modules/post/controllers/PostController.php#L62`
- What if a module don't want to show up in the form? Leave `createRoute` empty?

- Documentation
	- Migration of existing Modules
	- Create Content documentation
	- PHP Doc Inline in `ContentActiveRecord` 
- Tests

protected/humhub/modules/content/widgets/stream/WallStreamEntryWidget.php
```diff
@@ -139,6 +145,21 @@ abstract class WallStreamEntryWidget extends StreamEntryWidget
...
+
+    /**
+     * @var string
+     */
+    public $menuIcon;
```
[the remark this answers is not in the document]
yurabakhtin: Yes, it was implemented here https://github.com/humhub/humhub/pull/5827/commits/aed77f74e887d35b1662396b21aac89b58dedcaa

yurabakhtin:
@luke- 

> - What if a module don't want to show up in the form? Leave `createRoute` empty?

I have implemented this now in commit https://github.com/humhub/humhub/pull/5827/commits/18c8c21ed5b0842c987dbfa1a26366d6eb946b15.

protected/humhub/modules/content/widgets/stream/WallStreamEntryWidget.php
```diff
@@ -139,6 +145,21 @@ abstract class WallStreamEntryWidget extends StreamEntryWidget
...
+
+    /**
+     * @var int
+     */
+    public $menuSortOrder;
```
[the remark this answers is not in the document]
yurabakhtin: Done in commit https://github.com/humhub/humhub/pull/5827/commits/33cca19cc872d6bdde229d426b2d47a60b82fb74.

protected/humhub/modules/content/components/ContentActiveRecord.php
yurabakhtin: Ok, I have changed it to `ManageContent` permission here https://github.com/humhub/humhub/pull/5827/commits/33cca19cc872d6bdde229d426b2d47a60b82fb74#diff-5fea8ec3ab977d4e8e1db5cf83b2509d7cdccff1e612ca16efb88c2a37e18793R108.
I just thought if we have the separate permission with word "Create" then it is more clear to use it, but I also find the permission `CreatePrivateContent`, so correctly to use `CreatePublicContent || CreatePrivateContent`, but if you think `ManageContent` is ok, then we keep it.

protected/humhub/modules/content/widgets/stream/WallStreamEntryWidget.php
yurabakhtin: Commit https://github.com/humhub/humhub/pull/5827/commits/33cca19cc872d6bdde229d426b2d47a60b82fb74

protected/humhub/modules/post/controllers/PostController.php
yurabakhtin: Ok, done in commit https://github.com/humhub/humhub/pull/5827/commits/33cca19cc872d6bdde229d426b2d47a60b82fb74

yurabakhtin:
@luke- 

> Old legacy modules that are not yet using "ContentActiveRecord::$createPermission" and or "WallStreamModuleEntryWidget::createRoute" and or "Module::getContentClasses" should still working (no fatal error)

Yes, it works without errors.

yurabakhtin:
@luke- 

> Refactor: We should everywhere use canEdit instead of old style: https://github.com/humhub/humhub/blob/master/protected/humhub/modules/post/controllers/PostController.php#L62

Commit https://github.com/humhub/humhub/pull/5827/commits/6b50fe8826b6b878642c8f143559ba0aa463f348

protected/humhub/modules/content/components/ContentContainerModule.php
yurabakhtin: @luke- I thought if a model should to be implemented only for Space and not for User.
If you think we don't need this, so all models will be implemented and for Space and for User then we can remove the param, let me know to remove.

protected/humhub/modules/content/components/ContentContainerModule.php
yurabakhtin: @luke- Do you mean the code like? :

```php
return [new Post($contentContainer)]
```

protected/humhub/modules/content/widgets/WallCreateContentForm.php
```diff
@@ -50,6 +50,11 @@ class WallCreateContentForm extends Widget
      */
     protected $form = "";
 
+    /**
+     * @var bool Display menu above this form in order to select a content type like Post, Poll, Task and etc.
+     */
+    public $displayMenu = false;
+
     /**
```
luke-, line 58: Would be great to have an additional property here, to completely disable the new feature.

e.g. 
```
public $enableContentTabs = true;
```

We could also rename `displayMenu` to  `displayContentTabs`.

Then it's possible via Configuration, to do something like this:

```
    'container' => [
        'definitions' => [
            'humhub\modules\content\widgets\WallCreateContentForm' => [
                'enableContentTabs' => false
            ],
        ],
    ],
```

Maybe we could also provide a callback possibility here to:
- Disable some content tabs
- Individually Sort Content Tabs/Menu Items

yurabakhtin: @luke- Ok, I have renamed `displayMenu` to `displayContentTabs` in the commit https://github.com/humhub/humhub/pull/5827/commits/28fce66d5415f88e837c2a1d7842405e921e1605 and https://github.com/humhub/polls/pull/101/commits/f267a486620736db0e14d17610e311473465e86e.

But really the properly is used only to hide/show all menu tabs completely above the form, it is used on case when the form is loaded by ajax for Post and Poll, i.e. on ajax call we should not display the menu tabs again, only form should be displayed there.

protected/humhub/modules/content/widgets/stream/WallStreamEntryWidget.php
```diff
@@ -367,4 +367,13 @@ public function getAttributes()
...
         ];
     }
+
+    /**
+     * Register assets that are required for menu item of content creating
+     *
+     * @since 1.13
+     */
+    public function registerCreateContentMenuAssets()
+    {
```
luke-, line 377: In case we really need that, i.e. the assets can't be loaded in the Create modal. Would it be good to add a link to the Tasks/Calendar module here for the documentation, so you can see how it can be used.

yurabakhtin: It was deleted here https://github.com/humhub/humhub/pull/5827/commits/7632d9cc15fbbfeb10a97bd91ac04eee93099996

protected/humhub/modules/content/widgets/stream/WallStreamEntryWidget.php
```diff
@@ -139,6 +146,21 @@ abstract class WallStreamEntryWidget extends StreamEntryWidget
...
+
+    /**
+     * @var int Sort order of create form and tab menu on wall stream
+     */
+    public $createFormSortOrder;
+
+    /**
+     * @var string Class name of the Form to create a Content from wall stream
+     */
+    public $createFormClass;
```
luke-, line 162: Must/Should this class based on `WallCreateContentForm`?

luke-, line 162: Must/Should this class based on `WallCreateContentForm`? If yes we should add this to the comment.

protected/humhub/modules/content/widgets/WallCreateContentMenu.php
```diff
@@ -0,0 +1,150 @@
...
+
+    /**
+     * @inheritdoc
+     */
+    public $template = 'wallCreateContentMenu';
+
+    /**
+     * @var int What first menu entries should be visible as tabs top level, rest entries will be grouped into sub-menu
+     */
+    public $visibleEntriesNum = 2;
```
luke-, line 49: Wouldn't it be better to implement this via Javascript, and decide there how many entries are displayed as tabs and which ones in the menu?

Then you could react better to mobile or different resolutions.

But this is just a note, maybe this can be implemented/changed later.

protected/humhub/modules/content/widgets/WallCreateContentForm.php
luke-: @yurabakhtin Okay, now I got it.

Wouldn't it be better to split it then?

- `humhub\modules\post\widgets\Form` extends `humhub\modules\content\widgets\WallCreateContentForm`.
- `humhub\modules\content\widgets\WallCreateContentForm` becomes abstract
- New `humhub\modules\content\widgets\WallCreateForm` as container class. Includes Tabs etc.

protected/humhub/modules/content/widgets/WallCreateContentMenu.php
```diff
@@ -0,0 +1,150 @@
...
+use humhub\modules\ui\menu\MenuLink;
+use humhub\modules\ui\menu\widgets\Menu;
+use Yii;
+
+/**
+ * WallCreateContentMenu is the widget for Menu above wall create content Form
+ *
+ * @property-read ContentContainerActiveRecord $contentContainer
+ * @author luke
+ * @since 1.13.0
```
luke-, line 22: Small note: Please use `1.13` and not include the patch level.

protected/humhub/modules/post/controllers/PostController.php
```diff
@@ -101,4 +104,16 @@ public function actionEdit($id)
         ]);
     }
 
+    public function actionForm()
```
luke-, line 107: Please rename in `actionCreateForm`.

luke- (changes requested):
@yurabakhtin 

- I'm unsure about this lines https://github.com/humhub/humhub/blob/master/protected/humhub/modules/content/widgets/views/wallCreateContentForm.php#L47-L121 - because modules can not control this part. e.g. if the `News` module would become inline editing (like Post/Poll) but don't want the `Upload files` button, there is no change to realize this. Maybe we move the marked lines into a own widget (e.g. `WallCreateContentFormFooter`) and embedd it here: https://github.com/humhub/humhub/blob/enh/239-stream-create-form-menu/protected/humhub/modules/post/widgets/views/form.php#L22
@yurabakhtin What do you think?
 
- Mobile, would be great if we can implement following style: 
image
  
- Calendar: The stream entry is created correctly after creation. However, when I add a user in the Participant dialog (second step), the stream entry is not updated. 

- Polls: The error reporting seems not to work.

protected/humhub/modules/content/components/ContentActiveRecord.php
```diff
@@ -269,6 +283,31 @@ public function getContentDescription()
...
+     */
+    public function getCreatePermission()
+    {
+        return $this->hasCreatePermission()
+            ? $this->getPermissionValue($this->createPermission)
+            : null;
+    }
+
+    /**
+     * Determines weather or not this records has an additional createPermission set.
```
ArchBlood, line 301: `Determines whether or not the record has an additional createPermission set.`

protected/humhub/modules/content/widgets/stream/WallStreamEntryWidget.php
yurabakhtin: Yes, it must, I added this in commit https://github.com/humhub/humhub/pull/5827/commits/8638d49b492a58fb7135705b49501166c094c8ca.

protected/humhub/modules/content/widgets/WallCreateContentMenu.php
yurabakhtin: Commit https://github.com/humhub/humhub/pull/5827/commits/ee09a612696597d5d6b2fd83c78d188941ce5f2c

protected/humhub/modules/content/widgets/WallCreateContentForm.php
yurabakhtin: @luke- I am not sure we need to split it, please read how it works now after commit https://github.com/humhub/humhub/pull/5827/commits/8d7e70a299761f9ef641a145b7f7d50105e8c693:

- In old version the post create form was rendered on home steam page by code `humhub\modules\post\widgets\Form::widget([...])`
- but in new version we should render top sorted form instead of the Post Form, so now the code was replaced with `humhub\modules\content\widgets\WallCreateContentForm:widget([...])` - https://github.com/humhub/humhub/pull/5827/commits/8d7e70a299761f9ef641a145b7f7d50105e8c693#diff-164af2fab55bf7dc8259870dcd57fb78b900ad9e1a121f5989cd026efddde697R31, there we find what form should be rendered as first form, so it may be Post or Poll form.
- The Post and Poll forms are extended from `humhub\modules\content\widgets\WallCreateContentForm`

---

If you need the split only in order to remove the property `public $displayContentTabs = false;` from the class `WallCreateContentForm` completely, then please confirm and I will try to implement this solution.
So new widget `humhub\modules\content\widgets\WallCreateForm` will be rendered as:

```
WallCreateContentMenu::widget([...]);
WallCreateContentForm::widget([...]);
```

right?

yurabakhtin:
> * I'm unsure about this lines https://github.com/humhub/humhub/blob/master/protected/humhub/modules/content/widgets/views/wallCreateContentForm.php#L47-L121 - because modules can not control this part. e.g. if the `News` module would become inline editing (like Post/Poll) but don't want the `Upload files` button, there is no change to realize this. Maybe we move the marked lines into a own widget (e.g. `WallCreateContentFormFooter`) and embedd it here: https://github.com/humhub/humhub/blob/enh/239-stream-create-form-menu/protected/humhub/modules/post/widgets/views/form.php#L22

Commit https://github.com/humhub/humhub/pull/5827/commits/ddeaa105964d824f0a6663b6baa815451ba38c02, and changes for Poll module - https://github.com/humhub/polls/pull/101/commits/1f8152ae800730abaf312e3cc34339a4b24808da.

protected/humhub/modules/content/components/ContentActiveRecord.php
yurabakhtin: Fixed here https://github.com/humhub/humhub/commit/ddeaa105964d824f0a6663b6baa815451ba38c02#diff-5fea8ec3ab977d4e8e1db5cf83b2509d7cdccff1e612ca16efb88c2a37e18793R301

protected/humhub/modules/space/controllers/SpaceController.php
```diff
@@ -98,7 +99,7 @@ public function actionIndex()
     public function actionHome()
     {
         $space = $this->contentContainer;
-        $canCreatePosts = $space->permissionManager->can(new CreatePost());
+        $canCreatePosts = (new Post($space))->content->canEdit();
```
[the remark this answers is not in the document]
yurabakhtin: Done in commit https://github.com/humhub/humhub/pull/5827/commits/662b9ba8c287214880253e7def85e0da2cd11c41.

yurabakhtin:
@luke- 

> Mobile, would be great if we can implement following style:

Commit https://github.com/humhub/humhub/pull/5827/commits/8ca1de706f5db025d8a04359e48bbde6dfba8299, please note I have done this style for widh <= 480px:

mobile_tabs_create_wall_entry

yurabakhtin:
@luke- 

> - Calendar: The stream entry is created correctly after creation. However, when I add a user in the Participant dialog (second step), the stream entry is not updated.

Done in commit https://github.com/humhub/calendar/pull/331/commits/54f38151e637273792a64fc89bcff0d4095b4be0.

yurabakhtin:
@luke- 

> Polls: The error reporting seems not to work.

Fixed in the commit https://github.com/humhub/humhub/pull/5827/commits/e961589aeea8e29d801a4c36f6ed950284564701.

protected/humhub/modules/content/widgets/WallCreateContentForm.php
luke-: @yurabakhtin  Thanks for the explanation, it is not clear for me why `The Post and Poll forms are extended from humhub\modules\content\widgets\WallCreateContentForm`.

`content\widgets\WallCreateContentForm` is more like a container, it holds the menu and renders inside `post\widgets\Form`. In your current solutions it's both. I think its much clearer if we split them: 

For example, I would find such a solution better separated from the code: 

## `content\widgets\WallCreateContentFormContainer`

formerly `content\widgets\WallCreateContentForm`. Is only the container, contains/integrates the tab menu. Contains the logic to preload the first available `content\widgets\AbstractWallCreateContentForm`. Checks if it should be displayed at all. (e.g. the user does not have permission to display a content).

This widget is placed above a Stream (Dashboard, Space, User Profile) to show the create input + tabs.
  
## e.g. `post\widgets\WallCreateContentForm`

Based on new `content\widgets\AbstractWallCreateContentForm`. Has mainy the code from the old `content\widgets\WallCreateContentForm` class for form submitting / content creation...

This class is not used directly in views. It's loaded by `WallCreateContentFormContainer` or returned by `WallStreamEntryWidget::createRoute`.

--- 

What do you think about the approach?

yurabakhtin: @luke- Yes, I agree with your approach, I will implement this.

yurabakhtin: @luke- Done in commit https://github.com/humhub/humhub/pull/5827/commits/ff9a8ae6a764bcd7b621d5ab1b1445b6f2e93e39.

luke- (approved):

PR #5859, opened 2022-09-12
Update SocialActivity.php

funkycram:
The `getUrl` method can exist in models that are not a content and not an instance or Comment or ContentContainerActiveRecord (e.g. in a custom module).

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**

### DIFF
--- a/protected/humhub/components/SocialActivity.php
+++ b/protected/humhub/components/SocialActivity.php
@@ -251,11 +251,9 @@ abstract class SocialActivity extends BaseObject implements rendering\Viewable
     {
         $url = '#';
 
-        if ($this->source instanceof Comment) {
-            $url = $this->source->getUrl();
-        } else if ($this->hasContent()) {
+        if ($this->hasContent()) {
             $url = $this->getContent()->getUrl();
-        } elseif ($this->source instanceof ContentContainerActiveRecord) {
+        } elseif (method_exists($this->source, 'getUrl')) {
             $url = $this->source->getUrl();
         }
 


funkycram:
About the errors:
E MailTargetRenderTest: Default view
E StreamCest: Test the stream date filters 
S TourCest: Test tour
✖ ChangeEmailCest: Ensure that changing email works
E InviteCest: Ensure that inviting external users by mail form works.
✖ PasswordRecoveryCest: Ensure that password recovery works
S ProfileCest: Ensure that saving the account works 
E NonceCest: Test statistic
E NonceCest: Test invalid statistic

I tested date filtering, changing email, inviting, password recovery and profile saving: I had no error and it worked properly with Humhub 1.12.1.
I could not understand what is the problem. I didn't test the NonceCest error nor the Tour.

luke-:
@funkycram I think the errors have a different cause and do not related to this  PR.
Regarding the  PR: Basically we try not to use functions like "method_exists" and rather work with defined classes or interfaces.   

Can you give an example where a custom module wants to use that?

ArchBlood:
> @funkycram I think the errors have a different cause and do not related to this PR. Regarding the PR: Basically we try not to use functions like "method_exists" and rather work with defined classes or interfaces.
> 
> Can you give an example where a custom module wants to use that?

The only thing that I could think of would be a call to an external source?

funkycram:
OK, I understand.
I my custom module, I have notifications about a model extending an ActiveRecord and having the `getUrl()` method.
So it's not a comment neither a content.
I just thought it would be better to have the `method_exists` to be more generic with various models (e.g. it's a bit strange to have an exception for the Comment class), but in my case it's not a problem as I can overwrite the `getUrl()` method in the notification class.

luke-:
@funkycram Yes, you are right, even the `Comment` case is not ideally solved.
The best solution and future would be to work with interfaces here. But I think we still need some time for the conception. Can you solve your current problem otherwise?

funkycram:
Yes, I can solve the problem, so no rush, thanks!
I close this PR as this suggestion is not so good.

Issue #3506, opened 2019-03-04
Opengraph support in controllers


buddh4:
Controller should be able to set [Open Graph](http://ogp.me/) metadata.

e.g. ContentContainerController should set the space image by default.
Other Controllers should be able to set images e.g. related to a content entry (first image within richtext).

Example:

```php
$this->setOpenGraphImage($imgUrl);
$this->setOpenGraphTitle($title);
...
```

Ace2213:
+1

| | Space URL | Profile URL | Content URL | Fallback | 
| ------------- | ------------- | ------------- | ------------- | ------------- |
| OG Image  | Space image | Profile image | (if) Attachment | Site logo |
| OG Title  | Space Title  | User name | Post by {Author} in {Space name} on {Site name} | Site name |
| OG Description  | Space subtext (truncated)  | User About (truncated)| Post content | Site description |
| OG Type  |Website|Website|(if) Attachment type|Website|

ArchBlood:
You should also add one for `og:description`
```php
$this->setOpenGraphDescription($description);
```

ArchBlood:
To keep this alive, would something like the following be possible with this?

```html
<meta property="og:url" content="<?= $og->setOpenGraphUrl($url); ?>" />
<meta property="og:type" content="article" />
<meta property="og:title" content="<?= $og->setOpenGraphTitle($title); ?>" />
<meta property="og:description" content="<?= $og->setOpenGraphDescription($description); ?>" />
<meta property="og:image" content="<?= $og->setOpenGraphImage($imgUrl); ?>" />
```
If this is the desired affect it shouldn't be that hard to implement into the main.php file and create a controller that can be `/* @var $og OpenGraph */`? :thinking:

Or go in a similar way that [vumanskyi/yii2-opengraph](https://github.com/vumanskyi/yii2-opengraph) has?

funkycram:
+1. Today, with Humhub 1.9, sharing a content with a link or the Social Share module on Facebook is quite bad: only the image of the space is show, but not title, description or image of the publication or other module's content.

lisandi:
+1 It isn't available in 1.10.3 either and to promote HumHub further as a social media and sharing platform it is a MUST HAVE to not lack behind other social scripts with similar functionalities. It is necessary that a featured image or the first image of a post gets shown in the shared post beside at least the title and some first lines (paragraph) of the post.

[...]

PR #5863, opened 2022-09-13
Enh #3506: Add Open Graph support to have preview of the pages on oth…

marc-farre:
Enh #3506: Add Open Graph support to have preview of the pages on other websites (Facebook, Twitter, etc.)

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**

### DIFF
--- a/protected/humhub/modules/post/controllers/PostController.php
+++ b/protected/humhub/modules/post/controllers/PostController.php
@@ -14,12 +14,14 @@ use humhub\modules\content\widgets\stream\StreamEntryOptions;
 use humhub\modules\content\widgets\stream\StreamEntryWidget;
 use humhub\modules\content\widgets\stream\WallStreamEntryOptions;
 use humhub\modules\content\widgets\WallCreateContentForm;
+use humhub\modules\file\converter\PreviewImage;
 use humhub\modules\file\handler\FileHandlerCollection;
 use humhub\modules\post\models\forms\PostEditForm;
 use humhub\modules\post\models\Post;
 use humhub\modules\post\permissions\CreatePost;
 use Yii;
 use yii\helpers\StringHelper;
+use yii\imagine\Image;
 use yii\web\HttpException;
 use yii\web\NotFoundHttpException;
 
@@ -47,6 +49,19 @@ class PostController extends ContentContainerController
         // Add metadata to view to preview the content with Open Graph protocol
         $this->view->metaTitle = $post->getOwner()->getDisplayName() . ' · ' . $this->contentContainer->getDisplayName();
         $this->view->metaDescription = StringHelper::truncate(RichText::convert($post->message, RichText::FORMAT_SHORTTEXT), 195, '[...]');
+        $previewImage = new PreviewImage();
+        foreach ($post->fileManager->findAll() as $file) {
+            if (
+                file_exists($file->store->get())
+                && $previewImage->applyFile($file) // File is an image
+            ) {
+                $originalImage = Image::getImagine()->open($file->store->get());
+                if ($originalImage && $originalImage->getSize()->getHeight() > 200 && $originalImage->getSize()->getWidth() > 200) { // 200px is the minimum size for Facebook
+                    $this->view->metaImage = $file->getUrl();
+                }
+                break;
+            }
+        }
 
         return $this->render('view', [
             'post' => $post,


ArchBlood:
Love the idea, but I'm wondering why the Post controller instead of the content container controller, wouldn't doing this make it more globalized?

funkycram:
> why the Post controller instead of the content container controller, wouldn't doing this make it more globalized?

Because we need the Post object (after find in the database) to get the title, message and image metadatas for the preview.
In the Content Container controller, we don't know which is the content.

funkycram:
About the failing checks:
✖ ChangeEmailCest: Ensure that changing email works
E InviteCest: Ensure that inviting external users by mail form works.
✖ PasswordRecoveryCest: Ensure that password recovery works
S ProfileCest: Ensure that saving the account works 
E NonceCest: Test statistic
E NonceCest: Test invalid statistic

I tested changing email, inviting, password recovery and profile saving: I had no error and it worked properly with Humhub 1.12.1.
I could not understand what is the problem. I didn't test the NonceCest error.

luke-:
@funkycram Thanks for the feature PR! 

FYI: I'm in the process of restructuring your PR a bit. Mainly that the complete meta handling sits in its own class. The current `View` class is already very large and I want to keep it as lean as possible.

Another point:  The `humhub\modules\web\pwa\widgets\LayoutHeader` is actually only for the PWA specific things and can be disabled if necessary.

See my copy of your PR here: https://github.com/humhub/humhub/pull/5868/files

funkycram:
Thanks @luke- , it's much better yes to separate the meta part in another class.
Then we need controllers to be able to set metadata, such as [the PostController](https://github.com/humhub/humhub/pull/5863/files#diff-d47476863657fd72e21ff4461f800efee7289a4c8641efa9f3169c65a39da4b6).
Perhaps we could add `View::setMeta()` method?

Should I merge your PR and finish the work or did you plan to do it?
Thanks.

luke-:
@funkycram I'm considering finishing implementing it. Currently I'm thinking about how to improve the API for the PostController (and other content controllers).

Post Controller:

```php 
$this->view->meta->setContent($post);  // For ContentContainer / Author Infos / Date Infos
$this->view->meta->setTitle($post->getOwner()->getDisplayName()); // The Container Name should be automatically appended
$this->view->meta->setDescription(RichText::convert($post->message, RichText::FORMAT_SHORTTEXT)); // Truncate by `ViewMeta`
$this->view->meta->setImage($post->fileManager->findAll()); // Auto select first image from result set
``` 
What do you think about such an API?

---
Often I see the problem, for example, when I paste from the intranet into Link in another software, that Opengraph/Meta does not work, because the meta data for the login page is displayed. e.g. https://stackoverflow.com/questions/23695745/how-to-serve-opengraph-for-a-private-page

Unfortunately, if we also want to solve this problem, it becomes more complicated. 

We somehow need to (if enabled) resolve the meta data before the permission "loginRequired" redirect and cache it to display it later on the login page.

Should we take that into consideration?  

This would definitely have to become an optional feature (and not active by default), since private information is partially accessible without a login.

funkycram:
> I'm considering finishing implementing it.
Thanks a lot!

> API for the PostController
I don't have enough experience with the API to be able to answer, but it looks good.
For the title, perhaps we could add the name of the container (e.g. the space name): `$post->getOwner()->getDisplayName() . ' · ' . $post->content->container->getDisplayName();`

About the contents of Humhubs not allowing guest access (and only in this case), indeed the login page is displayed. For other Humhubs, private contents return 404 Not found (or 401 Unauthorized if the Space is not public for guests).
In my opinion, ideally, for Humhubs not allowing guest access, content metatags for OpenGraph should give any information about the content (as the platform is private), but could be:
- Title: Name of the application
- Description: Empty
- Image: the logo of the application

If you think this is a good option, then we only need to set the meta title in [AuthController::actionLogin()](https://github.com/humhub/humhub/blob/master/protected/humhub/modules/user/controllers/AuthController.php#L93).

Now, my answer is different for internal links (links from the same Humhub instance), with the [LinkPreview](https://www.humhub.com/en/marketplace/linkpreview/) module: if think this is a problem that could be solved by the module, but ideally, the preview should work even if the Humhub is not allowing guest access.

Thanks again for this work!

luke-:
@funkycram Thanks, merged https://github.com/humhub/humhub/pull/5868 based on your work. Feel free to enhance if required.

funkycram:
> @funkycram Thanks, merged #5868 based on your work. Feel free to enhance if required.

Thanks a lot @luke- !

PR #5864, opened 2022-09-14
Use Base URL from general settings for all absolute URLs

yurabakhtin:

- [x] Feature

- [x] All tests are passing
- [x] Changelog was modified

**Other information:**
Issue: https://github.com/humhub/team_tasks/issues/250

### DIFF
--- a/protected/humhub/components/UrlManager.php
+++ b/protected/humhub/components/UrlManager.php
@@ -50,6 +50,46 @@ class UrlManager extends \yii\web\UrlManager
      */
     public function getHostInfo()
     {
-        return Yii::$app->settings->get('baseUrl') ?: parent::getHostInfo();
+        $baseUrl = Yii::$app->settings->get('baseUrl');
+
+        if (empty($baseUrl)) {
+            return parent::getHostInfo();
+        }
+
+        $data = parse_url($baseUrl);
+
+        return $data['scheme'] . '://' . $data['host'] . (isset($data['port']) ? ':' . $data['port'] : '');
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function getBaseUrl()
+    {
+        $baseUrl = Yii::$app->settings->get('baseUrl');
+
+        if (empty($baseUrl)) {
+            return parent::getBaseUrl();
+        }
+
+        $data = parse_url($baseUrl);
+
+        return $data['path'] ?? '';
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function getScriptUrl()
+    {
+        $baseUrl = Yii::$app->settings->get('baseUrl');
+
+        if (empty($baseUrl)) {
+            return parent::getScriptUrl();
+        }
+
+        $data = parse_url($baseUrl);
+
+        return ($data['path'] ?? '') . '/' . basename(parent::getScriptUrl());
     }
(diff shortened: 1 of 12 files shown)


protected/humhub/components/UrlManager.php
```diff
@@ -43,4 +44,12 @@ public function createUrl($params)
 
         return parent::createUrl($params);
     }
+
+    /**
+     * @inheritdoc
+     */
+    public function getHostInfo()
+    {
+        return Yii::$app->settings->get('baseUrl') ?: parent::getHostInfo();
```
luke-, line 53: Please note that `baseUrl` can also something like `https://localhost/folder/humhub`.

https://www.yiiframework.com/doc/api/2.0/yii-web-request#getHostInfo()-detail

`getHostInfo` should only return `https://localhost`

luke- (changes requested):
@yurabakhtin Thanks, please see my comment. Since it's a bigger change to switch all absolute URLs (not only e-email) to the baseUrl Setting, I've changed the branch to `develop`

protected/humhub/components/UrlManager.php
yurabakhtin: @luke- Please see my test when `getHostInfo()` returns `http://localhost`:

folder

luke-: This is strange, in my test cases I always got back a correct absolute URL when I return `http://localhost` there for example.

My Tests:
- With and without enabled Pretty URL in Root Folder
- With and without enabled Pretty URL in '/humhub` subfolder

As webserver I'm using NGINX. I added my WebServer Config for the subfolder setup here:
https://community.humhub.com/s/installation-and-setup/wiki/WebServer+Configurations

Maybe it is your WebServer Config?

yurabakhtin: @luke- Sorry, my previous test was not completely correct, because Base URL was defined as `http://localhost/humhub` but the request a password was done from alias `http://humhub.yb`.

test-hosts

I.e. yesterday I tested only the case `2a`.

So if we return only host as `http://localhost` then case `2a` will be broken, however this solution will fixes the case `1b`.
I think to find a solution for all cases.

luke-: @yurabakhtin Your case `2a` is very special, you assume here that HumHub can run in parallel in the root folder as well as in a subdirectory. I would just assume that the domain can change.

luke-: FYI: here is how we use the base url in the console part: https://github.com/humhub/humhub/blob/master/protected/humhub/components/console/Application.php#L80-L90

yurabakhtin: @luke- I wanted to post this comment 5 hours ago, but it seems I just forgot to press "Comment", and it was stored here:

---

Maybe it is not correctly to use a site from alias like `http://humhub.yb` when base URL is defined as `http://localhost/humhub`, but I have implemented a solution here https://github.com/humhub/humhub/pull/5864/commits/ae1886fbfa9a72389f8238a06bf2a0d2303c33ae

my tests for the code `Url::to(['/user/password-recovery/reset'], true)`:

fixed_host

luke- (approved):

Issue #3534, opened 2019-03-29
ldap user desactivat since last update


Spartacus2018:
### What steps will reproduce the problem?
Evry morning i come to my humhub and all my user from ldap are desactivate, ihave to connect in admin user and activate all user one by one

### What is the expected result?

All ldap user update or improrted for the first time have to be activated

### What do you get instead?
i got this problem since the last update HumHub 1.3.12 has just been released.

### Additional info

| Q                | A
| ---------------- | ---
| HumHub version   | Version installée actuellement : 1.3.12
| PHP version       PHP - Version - 7.0.33-0ubuntu0.16.04.3 | 
| Operating system ubuntu serveur |

> Note: Also provide the full error stacks from `Administration --> Information --> Logging` and errors from your browser console if possible.
> Note: Please use the [community](https://community.humhub.com) site for any questions on customizing HumHub.

lutzministrator:
I'm having the same problem. Fresh install 1.3.12 on Ubuntu 18.04 with php 7.2. 
Cronjobs running, LDAP users imported successfully, but deactivated assumingly by the hourly cronjob. 
Logfile says 'Disabled user[username] (4) - not found in LDAP! '

pe-en:
Same problem with my system too. HELP... 😫

Spartacus2018:
Hello 
to solve this problem u have to do this : 

> Lucas Bartholemy 29 mars 2019 - 10:27
> 
> Try to set this attribute to “objectguid” and wait an hour (or run the php yii ldap/sync command)

Found here : 
https://community.humhub.com/s/general-discussion/?contentId=184494

pe-en:
THX seems to work! 👍

lutzministrator:
I can confirm setting an ID Attribute solves the issue. Why not make that field mandatory if it's necessary for LDAP to work properly?

biznlinks:
Even with the added objectguid in the ID attribute, some users are still being disabled hourly. Assistance required urgently.

aklamke:
> Even with the added objectguid in the ID attribute, some users are still being disabled hourly. Assistance required urgently.

Please check if you have a space behind the ID attribute. This happened to my because I copy/pasted from the hint on the page. Every run of "yii ldap/sync" forced every user to be disabled. After deleting the space and rerun of "yii ldap/sync" all users came back activated.

svenjungnickel:
[...]

PR #5873, opened 2022-09-19
Fix users being disabled because operational attributes are not queried

alexhorn:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**
In OpenLDAP, the `entryUUID` attribute is an operational attribute which is only returned when explicitly queried. Before, HumHub would refuse to work with our LDAP server and disable all of the users when running `ldap/sync`. This is because `entryUUID` is correctly set as the ID attribute in the HumHub LDAP settings, however, since HumHub does not request operational attributes, it would not be able to find our users in the database and end up disabling everyone. Issue #3534 might be related to this.

This might break compatibility with Active Directory, if it can not handle `+` in the requested attribute list. Unfortunately, I do not have an AD server ready for testing.

### DIFF
--- a/protected/humhub/modules/ldap/Module.php
+++ b/protected/humhub/modules/ldap/Module.php
@@ -27,5 +27,5 @@ class Module extends \humhub\components\Module
     /**
      * @var array|null the queried LDAP attributes, leave empty to retrieve all
      */
-    public $queriedAttributes = [];
+    public $queriedAttributes = ['*', '+'];
 }


luke-:
@alexhorn Thanks for the PR but the preferred way is to add this via configuration `protected/config/common.php'.

e.g.  

```php
return [
   'modules' => [
      'ldap' => [
          'queriedAttributes' => ['*', '+'],
      ]       
   ]
];
```

alexhorn:
@luke- 
Thanks for the tip, but I think this should be fixed anyway. As it stands, the documentation suggests that configure a UUID attribute in the web interface. I would assume would mean that HumHub would actually request the configured attribute from the LDAP server, which it unfortunately doesn't.

luke-:
@alexhorn Thanks for the hint. With MS Active Directory the UUID/GUID attribute is also returned without "+". We have to check the compatibility before we change the default setting.

PR #5951, opened 2022-11-29
Add space between icon and text inside button

yurabakhtin:

- [x] Feature

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**
Issue: https://github.com/humhub/wiki-2.0-issues/issues/10#issuecomment-1330358168


luke- (approved):

yurabakhtin:
@luke- I think we should implement some new style class like `btn-icon` where the default space after icon will be removed, because if we have a button without text like this it looks not good:

icon-button

PR #5962, opened 2022-12-01
Update humhub.ui.topNavigation.js: make menu shrink and grow

aronbos:

- [ ] Bugfix
- [x] Feature
- [x] Code style update
- [x ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

Rational:
Humhub is often viewed on a table or phone. Switching between portrait en landscape views is common on this devices.
The menu in Humhub will shrink, but not grow. Switching to a different view-mode, Humhub will stay in the shortest view mode, that is a menu with a pull-down.
This commit resets the menu pull-down after a resize, which gives a much better user experience. The effect is that the menu growth as well.

The extra 3 lines in the on resize function are simple and only depend on the view ids.

Extra changes are:
- Do not find ids, just select them.
- Bring the reset in the resize event.
- Do not chain different ids and classes.
- Do not remove the next dropdown for edge-cases. This breaks the user experience on tablets.

### DIFF
--- a/static/js/humhub/humhub.ui.topNavigation.js
+++ b/static/js/humhub/humhub.ui.topNavigation.js
@@ -1,35 +1,37 @@
 humhub.module('ui.topNavigation', function (module, require, $) {
     var $topBarSecond = $('#topbar-second');
-    var $topNav = $topBarSecond.find('#top-menu-nav');
-
-    // Prevent overflow on init
-    $topBarSecond.css('overflow', 'hidden');
+    var $topNav = $('#top-menu-nav');
 
     var init = function () {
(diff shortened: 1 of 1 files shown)


luke-:
@aronbos Sorry for the delay here. Should we still merge this?

aronbos:
Yes, I am using this in my own machine without problems.

yurabakhtin:
> @yurabakhtin Can you please take a look?

@luke- As I wrote above there is a small issue on reduce window and then again to expand it. So it would be good to include the small fix.

PR #5973, opened 2022-12-08
Remove  image-orientation as it was temporary (was already moved to m…

aronbos:
…edia.less)

protected/humhub/widgets/LayoutAddons.php)

// Temporary fix until v1.6 then bundled with css pack
 Yii::$app->view->registerCss('img { image-orientation: from-image; }', [], 'fix-orient');

Which is already set in static/less/media.less)

img {
    image-orientation: from-image;
}

I propose to remove to temporary fix in LayoutAddons.php


luke-:
@aronbos Thanks for the contribution. Can you please use as Base Branch Develop.

luke- (approved):

Issue #4803, opened 2021-01-29
Add panel-info


ArchBlood:
### Is your feature request related to a problem? Please describe.
As the issue title suggests, I already see that it's possible to use `panel-success`, `panel-warning`, `panel-danger` and `panel-default`

### Describe the solution you'd like
Add the following to the theme.css file
```css
.panel-info{
    border:2px solid #5bc0de
}
.panel-info .panel-heading{
    color:#5bc0de
}
````
### Describe alternatives you've considered
I've considered adding via custom theme but overall thought it a better solution to add to the core theme.css as it's missing there

buddh4:
I think those are similar: https://getbootstrap.com/docs/4.0/components/alerts/ ?

ArchBlood:
In a way yes, the only thing these do is highlight the boarder of the panel, only thing is `panel-info` is missing

ArchBlood:
For an example
![Screenshot_1](https://user-images.githubusercontent.com/35392110/106326418-6ebe4f80-624a-11eb-86d1-0a5cde8601dc.png)

ArchBlood:
@luke- here's an example of where this could be used, which can highlight areas that a user should learn about the platform.

![Screenshot_1](https://user-images.githubusercontent.com/35392110/208322894-e085f4e8-c1f2-45d9-b536-c936d8277e11.png)

Of course, this would require that line 7 of the `tourPanel.php` be changed to `panel panel-info panel-tour` over the default one.

ArchBlood:
> @luke- here's an example of where this could be used, which can highlight areas that a user should learn about the platform.
> 
> ![Screenshot_1](https://user-images.githubusercontent.com/35392110/208322894-e085f4e8-c1f2-45d9-b536-c936d8277e11.png)
> 
> Of course, this would require that line 7 of the `tourPanel.php` be changed to `panel panel-info panel-tour` over the default one.

Through looking at the tour module I've noticed that the `Administration (Modules)` tour is broken? :Thinking:

PR #5979, opened 2022-12-16
Changed theme.css file

sveneschlbeck:
#4803

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x ] Other, please describe: Added panel info to ``theme.css``

- [ ] Yes
- [ x] No


ArchBlood (changes requested):
The changes here just add the CSS to the bottom of the file, it should be placed where the other `.panel` entries are, not only this but it only adds parts of what #4803 is interpreting.

luke-:
Thank you for your contribution.
The `theme.css` file is automatically compiled from the various LESS files and therefore cannot be edited manually.

The following file is used for this purpose:
https://github.com/humhub/humhub/blob/develop/themes/HumHub/less/build.less

sveneschlbeck:
@luke- Do you mean this file?

https://github.com/humhub/humhub/blob/develop/themes/HumHub/less/theme.less

sveneschlbeck:
Is the ``.panel-info`` supposed to just be added to the so-far empty LESS file?

felixhahnweilheim:
https://github.com/humhub/humhub/blob/master/static/less/panel.less

static/less/panel.less
```diff
@@ -206,6 +206,13 @@
     }
 }
 
+.panel-info{
+    border:2px solid #5bc0de
+}
+.panel-info .panel-heading{
+    color:#5bc0de
```
luke-, line 213: Please do not use hardcoded color values. Instead use color variables like `color:@info` 

You can find available color variables here:
https://github.com/humhub/humhub/blob/master/static/less/variables.less#L12

luke- (changes requested):

static/less/panel.less
sveneschlbeck: Thanks! Done

static/less/panel.less
```diff
@@ -206,6 +206,13 @@
     }
 }
 
+.panel-info{
```
luke-, line 209: Please add a whitespace between classname and bracket.

static/less/panel.less
```diff
@@ -206,6 +206,13 @@
...
+    border:2px solid #5bc0de
```
luke-, line 210: Please also use color code here and end the line with semicolon.

static/less/panel.less
```diff
@@ -206,6 +206,13 @@
...
+}
+.panel-info .panel-heading{
+    color: @info
```
luke-, line 213: Please end the line with semicolon.

static/less/panel.less
```diff
@@ -206,6 +206,13 @@
     }
 }
 
+.panel-info{
+    border:2px solid #5bc0de
+}
+.panel-info .panel-heading{
```
luke-, line 212: Please add a whitespace between classname and bracket.

luke- (changes requested):
Regarding the

ArchBlood (approved):
Looks good to me

luke- (approved):

PR #5980, opened 2022-12-17
Actually cache $_moduleInfo

martin-rueegg:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**The PR fulfils these requirements:**

- [ ] All tests are passing -> behaviour does not change
- [ ] New/updated tests are included -> behaviour does not change
- [ ] Changelog was modified -> Which changelog?

**Other information:**
Please let me know if I should consider/fix something in my code or where I should add this to a change log. Thank you!

### DIFF
--- a/protected/humhub/components/Module.php
+++ b/protected/humhub/components/Module.php
@@ -30,9 +30,9 @@ use yii\web\AssetBundle;
 class Module extends \yii\base\Module
 {
     /**
-     * @var array the loaded module.json info file
+     * @var array|null the loaded module.json info file
      */
-    private $_moduleInfo = null;
+    private ?array $_moduleInfo = null;
 
     /**
      * @var string The path for module resources (images, javascripts)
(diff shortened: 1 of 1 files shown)


ArchBlood:
Could you provide a little more information on which part (if any) of the the `$_moduleInfo` that isn't cached?

luke-:
@martin-rueegg Thanks for the fix!

@ArchBlood Seems we don't set the `_moduleInfo` attribute at all.

luke- (approved):

ArchBlood:
@luke- just wondering if this would be a better solution;

```php
    /**
     * Reads module.json which contains basic module information and
     * returns it as array
     *
     * @return array module.json content
     */
    protected function getModuleInfo(): ?array
    {
        if ($this->_moduleInfo !== null) {
            return $this->_moduleInfo;
        }
        $moduleJson = file_get_contents($this->getBasePath() . DIRECTORY_SEPARATOR . 'module.json');

        return $this->_moduleInfo = Json::decode($moduleJson);
    }
```
This way you don't have to change `$_moduleInfo` itself and still caches the modules installed.

luke-:
@ArchBlood Hmm, I would prefer both, adding a return type to the method and add the type to the attribute.

martin-rueegg:
I haven't added the return type to the method in order not to break backwards compatibility, as any deriving class would have to adapt to the new interface ...

But it seems there is a bit of inconsistency throughout the project as to where return types are used and where not.

luke-:
@martin-rueegg Yes, we try to add this to new methods. Maybe we can refactor a larger part automatically with tools like Rector.

martin-rueegg:
@luke- Well, refactoring can also be done with IDEs like PhpStorm. I don't see the issue within the core code. However, you will still break backwards-compatibility as modules that extend such a modified class will no longer compile.

Issue #6002, opened 2022-12-30
Link to private content of public Space results in 404


luke-:
### Is your feature request related to a problem? Please describe.

If the link to a private content of a public Space is shared to a non-member (of the Space), that user will receive a 404 error message.

### Describe the solution you'd like

Instead, it would be better to display a page, with a text like:
"This content is only available to members of this Space. Click here to become/request membership."

--- 
For content of a Private Space the generic 404 page is correct, however we should improve the text a bit here.
e.g. The requested content cannot be displayed. Either it was deleted, you mistyped it or it is not available for you.

yurabakhtin:
@luke- As I understand you tell about core content records `Poll`, but probably we should apply the same my new code `ContentContainerController->checkContentIsReadable($record)` for other external module records like `Calendar`, `Task` and etc.

PR https://github.com/humhub/humhub/pull/6005 :

I could not implement the link to request a membership because there the code `Html::encode()` before display a message, so it looks like this:

![html_encoded](https://user-images.githubusercontent.com/6995524/210209643-8fd4048d-faa7-4ba7-80a6-5355f9031f72.png)

I have decided to use the message:

![message_403](https://user-images.githubusercontent.com/6995524/210209673-f866bfba-ddd8-49d7-9f6f-b053586c5fa8.png)

for 404:

![message_404](https://user-images.githubusercontent.com/6995524/210209784-90a5e4bd-7bf6-417b-a5fc-ef8e4deb63aa.png)

luke-:
@yurabakhtin In the first case (Private Content & Public Space): I would like instead of the default Error View, a page in Space layout (with header). Maybe @Eladnarlea can also design it a bit.

The second case (Link to private Space): Please leave it as it is and don't change anything here for now.

luke-:
@Semir1212 You you please add a message here. Maybe with headline. Please note, it's not always possible to become/request membership. Maybe some illu from @Eladnarlea ?

Semir1212:
@luke- 

**Access denied**

You do not have permission to access this content, as it is reserved for members of this Space. Please become a member or apply for membership. The available options for membership will depend on the Space's settings.

Eladnarlea:
<img width="668" alt="Screen Shot 2023-03-23 at 1 02 53 PM" src="https://user-images.githubusercontent.com/105652258/227198146-70c42a83-33dc-47bd-b07c-ef44316aab42.png">

How do you like something like this? or do you want me to design it more like an info?

[...]

PR #6005, opened 2023-01-02
Display more info when content record is not readable

yurabakhtin:

- [ ] Feature

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
Issue: https://github.com/humhub/humhub/issues/6002

### DIFF
--- a/protected/humhub/modules/content/components/ContentContainerController.php
+++ b/protected/humhub/modules/content/components/ContentContainerController.php
@@ -8,17 +8,16 @@
 
 namespace humhub\modules\content\components;
 
+use humhub\components\Controller;
+use humhub\modules\content\models\ContentContainer;
 use humhub\modules\admin\permissions\ManageSpaces;
 use humhub\modules\space\models\Space;
 use humhub\modules\user\helpers\AuthHelper;
 use humhub\modules\user\models\User;
 use Yii;
+use yii\web\ForbiddenHttpException;
 use yii\web\HttpException;
-use humhub\components\Controller;
-use humhub\modules\content\models\ContentContainer;
-use humhub\modules\content\components\ContentContainerModule;
-use humhub\modules\content\components\ContentContainerActiveRecord;
-use humhub\modules\content\components\ContentContainerControllerAccess;
+use yii\web\NotFoundHttpException;
 
 /**
  * Controller is the base class of web controllers which acts in scope of a ContentContainer (e.g. Space or User).
@@ -146,6 +145,29 @@ class ContentContainerController extends Controller
             !$this->contentContainer->moduleManager->isEnabled($this->module->id)) {
             throw new HttpException(405, Yii::t('base', 'Module is not enabled on this content container!'));
         }
+    }
+
+    /**
+     * Check if the requested content record is readable for the current or given User
+     *
+     * @param ContentActiveRecord|null $record
+     * @param User|integer|null $user
+     * @return bool
+     * @throws HttpException
+     * @throws \Throwable
+     * @throws \yii\base\Exception
+     */
+    protected function checkContentIsReadable(?ContentActiveRecord $record, $user = null): bool
+    {
+        if (!($record instanceof ContentActiveRecord)) {
+            throw new NotFoundHttpException(Yii::t('base', 'The requested content cannot be displayed. Either it was deleted, you mistyped it or it is not available for you.'));
+        }
+
+        if (!$record->content->canView($user)) {
+            throw new ForbiddenHttpException(Yii::t('base', 'This content is only available to members of this Space. Please become/request membership.'));
+        }
+
+        return true;
     }
 
     /**
--- a/CHANGELOG-DEV.md
+++ b/CHANGELOG-DEV.md
@@ -7,3 +7,4 @@ HumHub Changelog (DEVELOP)
 
 - Enh #4803: Added more panel styles (`panel-info` and `panel-primary`) 
 - Enh #5972: Removed old vendor CSS prefixes (e.g. `-moz` or `-webkit`) 
+- Enh #6002: Display more info when content record is not readable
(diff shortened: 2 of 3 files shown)


protected/humhub/modules/content/components/ContentContainerController.php
```diff
@@ -144,6 +147,30 @@ protected function checkModuleIsEnabled()
...
+     * @param User|integer|null $user
+     * @return bool
+     * @throws HttpException
+     * @throws \Throwable
+     * @throws \yii\base\Exception
+     */
+    protected function checkContentIsReadable(?ContentActiveRecord $record, $user = null): bool
+    {
+        if (!($record instanceof ContentActiveRecord)) {
+            throw new NotFoundHttpException(Yii::t('base', 'The requested content cannot be displayed. Either it was deleted, you mistyped it or it is not available for you.'));
```
ArchBlood, line 163: @luke- instead of `not available` maybe change to `currently not available`,  this way it lets the user know that it's not available at this time but once things are fixed it may be available at a later date.

luke-: @Semir1212 What do you think?

Semir1212: Looks good. @luke-

luke-: @yurabakhtin Can you please adapt?

yurabakhtin: @luke- Done in commit https://github.com/humhub/humhub/pull/6005/commits/3f8754be630dfe5c9e72ed8f52586ea45b085ca4.

luke- (approved):

luke- (changes requested):
Can you please add tests for it?

luke-:
@yurabakhtin Since we have to customize every module with a content detail view for this anyway, I am just thinking if it would be cleaner from an implementation point of view to introduce an additional `ContentController` besides the `ContentContainerController`.

Features could be:

- SEO Header Tags
- Errors Messages & Access Control
- Generic Content Type Features
      - A generic "About" page for content (e.g. view Created At/Updated At) 
      - Automatic Link to this Controller from `ContentActiveRecord`
      
Challenges/Problems: 
- Generic Load of Content by Query. e.g. for a wiki page we currently use the page title otherwise often the ID.
-  How to support content without content container

What do you think about this?

yurabakhtin:
@luke- I don't understand where new `ContentController` should be used. For example, the `ContentContainerController` is used for `class PollController extends ContentContainerController`, but how the new controller will be used?

luke-:
@yurabakhtin The Idea is something like that:

**New ContentController**
```php
abstract class ContentController extends ContentContainerController
{
    /**
     * @var ContentActiveRecord
     */
    public $record;


    /**
     * @inheritdoc
     */
    public function beforeAction($action)
    {
        // Auto Load Content by query param e.g. 'content.guid`

        // Check Access Rights handle Error Pages

        if (!parent::beforeAction($action)) {
            return false;
        }

        // Register SEO Meta Tags

    }


    public function actionInfo()
    {
        // Shows a modal with general content information
        // e.g.  creator, updated, archived, ....
        return $this->render('info', ['record' => $this->record]);
    }

    public function actionDelete()
    {
        // Generic Delete Method instead ContentController::delete();
    }

    // More Generic Methods from ContentController

}
```

**Poll Model**
```php
class Poll extends ContentActiveRecord implements Searchable
{

    public static $contentController = PollDetailController::class;
```

**PollDetail Controller**
```php
class PollDetailController extends ContentController
{
    public function actionReload($id)
    {
        return $this->asJson(Stream::getContentResultEntry($this->record->content));
    }

}
```

I'm not sure yet if the benefits are great enough to be worth for such larger rewrite/change.

But with the current approach we need to add following lines in all modules content view controllers.

```
        if (!$this->checkContentIsReadable($post)) {
            return $this->renderNotAvailableContent();
        }
```

yurabakhtin:
@luke- Thanks for the samples!

Yes, I think it is a good idea. But I am not sure we should create new controllers like `PollDetailController`, instead it would be good to factorize already existings controllers `PollController`, so URLs will not changed and the old controllers will have less code.

I.e. actions `actionDelete`, `actionEdit`, `actionCreate`, `actionCreateForm` might be factorized, but probably some of them have very specific code.

---

> `public $record;`

I think this record can be initialized in `beforeAction($action)` from passed GET or POST data `$id`

> `public static $contentController = PollDetailController::class;`

I am thinking how to find the model class on call the controller. I mean of call an URL like `s/space-name/poll-detail/123` the controller `PollDetailController` is called, and we need to initialize the `$record` from some class, so the class is unknown here.
I think we should initialize the record class from the new `ContentController` side, like:

```php
class PollDetailController extends ContentController
{
    public string $recordClass = Poll::class;
}
```

luke-:
@yurabakhtin Thanks for the feedback.

When you get a chance, can you give an example of what you mean by "to factorize already existing controllers actions"? 

I agree with you, adding extra controllers, like `PollDetailController` also has drawbacks. 

Maybe we can add the `ContentController` features directly in the existing `ContentContainerController` and the implementation e.g. `PollController`  needs to implement e.g. a loading method or via  a behavior/mixin.

yurabakhtin:
@luke- 

> When you get a chance, can you give an example of what you mean by "to factorize already existing controllers actions"?

I mean instead of creating new controllers like `PollDetailController`(which will be extended from new `ContentController`) it will be better to use the existing controllers like `PollController`. Because such main module controllers already have actions `actionCreate`, `actionEdit`, `actionDelete`.

> I agree with you, adding extra controllers, like `PollDetailController` also has drawbacks.

Yes, it would be better to use the same controllers, so urls like `s/space-name/poll/123` will not be changed to `s/space-name/poll-details/123`.

> Maybe we can add the `ContentController` features directly in the existing `ContentContainerController` and the implementation e.g. `PollController` needs to implement e.g. a loading method or via a behavior/mixin.

As I understand new `ContentController` should have actions like `actionCreate`, `actionEdit`, `actionDelete` in order to work with Active Records, so I don't think we need to keep them in the `ContentContainerController` because this controller is also used for config/settings controllers where no code of managing Active Records.

PR #6012, opened 2023-01-04
View own invisible profile

yurabakhtin:

- [x] Bugfix

- [x] All tests are passing
- [x] Changelog was modified

**Other information:**
Issue: https://github.com/humhub/team_tasks/issues/269

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ HumHub Changelog
 - Fix #5996: Fix param type in strpos() (PHP 8.1)
 - Fix #5997: Possible NULL param value in Birthday field (PHP 8.1) 
 - Enh #6001: Added new `ContentActiveFixture` and migrated `PostFixture` to it
-- Fix #: View own invisible profile
+- Fix #6012: View own invisible profile
 
 1.13.0 (December 21, 2022)
 --------------------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ HumHub Changelog
 - Fix #5996: Fix param type in strpos() (PHP 8.1)
 - Fix #5997: Possible NULL param value in Birthday field (PHP 8.1) 
 - Enh #6001: Added new `ContentActiveFixture` and migrated `PostFixture` to it
+- Fix #: View own invisible profile
 
 1.13.0 (December 21, 2022)
 --------------------------
--- a/README.md
+++ b/README.md
@@ -5,29 +5,26 @@ HumHub - Putting People and Pieces together
 [![Yii2](https://img.shields.io/badge/Powered_by-Yii_Framework-green.svg?style=flat)](http://www.yiiframework.com/)
 [![CLA assistant](https://cla-assistant.io/readme/badge/humhub/humhub)](https://cla-assistant.io/humhub/humhub)
 
-**HumHub is an intuitive to use and modular designed open-source software**, used primarily as social network, knowledge database, intranet or information and communication platform.
+#### **HumHub is an intuitive to use and modular designed open-source software**, used primarily as social network, knowledge database, intranet or information and communication platform.
 
 **The software is written in PHP** and is best described by dividing into 4 main parts: **User, Spaces, Content and Modules.**
 
-**User:** All users have their own customisable profile (including name, profile picture, cover photo and personal information) and can follow and interact with each other. If wished and enabled, users can create own content, comment posts and join Spaces. Profile fields, permissions and all settings can be defined individually by the network operator (administrator).
+- **User:** All users have their own customisable profile (including name, profile picture, cover photo and personal information) and can follow and interact with each other. If wished and enabled, users can create own content, comment posts and join Spaces. Profile fields, permissions and all settings can be defined individually by the network operator (administrator).
 
-**Spaces:** Rooms or groups for any projects, departments, events or other needs. Network operators can create as many Spaces as needed and automatically map users into the desired Spaces. HumHub comes with an advanced permission and notification system (including email summaries).
+- **Spaces:** Rooms or groups for any projects, departments, events or other needs. Network operators can create as many Spaces as needed and automatically map users into the desired Spaces. HumHub comes with an advanced permission and notification system (including email summaries).
 
-**Content:** Users can create content of all kinds (posts, wiki pages, photo/video, schedule appointments, create events or tasks) depending on their permission and share it with other members in their Space. There is a multi-level comment function, versatile collaboration options and also features to report inappropriate posts and content. All Content, Spaces and Members can be easily found through various and individually definable filter and search functions, Content can be edited, deleted and archived.
+- **Content:** Users can create content of all kinds (posts, wiki pages, photo/video, schedule appointments, create events or tasks) depending on their permission and share it with other members in their Space. There is a multi-level comment function, versatile collaboration options and also features to report inappropriate posts and content. All Content, Spaces and Members can be easily found through various and individually definable filter and search functions, Content can be edited, deleted and archived.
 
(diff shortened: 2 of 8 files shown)


luke- (changes requested):
Hmm, what do you think of the following approach:

- Create new class `AbstractActiveQueryContentContainer` with abstract method `visible()` and derive `ActiveQuerySpace` and `ActiveQueryUser` from it.

- In `ContentContainerController::getContentContainerByGuid()` simply call `visible()` without condition.

- Move further logic e.g. `CanManageSpace` or `canSeeOwnProfile` to the respective `visible()` implementation of `ActiveQuerySpace` or `ActiveQueryUser`.

yurabakhtin:
@luke- I find we already have the methods `ActiveQuerySpace->visible()` and `ActiveQueryUser->visible()`, so I just moved the conditions into the methods - commit https://github.com/humhub/humhub/pull/6012/commits/559150221fe15740d64aa5ce1e183e378a099e4a.
I didn't see a reason to create the abstract class `AbstractActiveQueryContentContainer`.
Instead we could create an interface with name like `Visible` or `Filterable` if we plan to extend this later with other methods.

luke-:
@yurabakhtin Ok, an interface is good too. 

- But I would call it either `ActiveQueryContentContainerInterface` where we collect all common methods for ContentContainer (Space, User).
- or `ActiveQueryVisibilityInterface` with the `visible()` method.

It would be good if we add the user to the method signature `visible(?User): void`.

yurabakhtin:
@luke- You was right firstly when suggested to use new abstract class `AbstractActiveQueryContentContainer` because I have found a method with same code in the user and space query classes, please find the implementaion in the commit https://github.com/humhub/humhub/pull/6012/commits/1d067a65f289f5791d23cb347fe94f8eee69251a.

----
Also I have added the param `$user` for the method `visible(?User)` and replaced the code from `Yii::$app->user->can(ManageUsers::class)` to `$user->can(ManageUsers::class)` where `$user` is `Yii::$app->user->getIdentity()`, so I detected result is different for admin:

- `Yii::$app->user->can(ManageUsers::class)` === `true`
- `Yii::$app->user->getIdentity()->can(ManageUsers::class)` === `false`

I think the result should be same in both cases, so ti should be fixed.

luke-:
@yurabakhtin Do you have any idea why the behavior of the two `can()` methods is different?

If there is no reason for this, I would mark the `Yii::$app->user->can()` method as deprecated and just refer to `Yii::$app->user->identity->can()`.

It might also make sense to rename `humhub\modules\user\components\User` to `humhub\modules\user\components\WebUser`.

yurabakhtin:
@luke- 

> Do you have any idea why the behavior of the two `can()` methods is different?

I still cannot find why `Yii::$app->user->getIdentity()->can(ManageUsers::class)` returns `false` when `true` is expected.
But the dirrence is that the component `Yii::$app->user` uses `new PermissionManager(['subject' => $this->getIdentity()])`,
and the model `Yii::$app->user->getIdentity()` uses `new ContentContainerPermissionManager(['contentContainer' => $this, 'subject' => Yii::$app->user->getIdentity()])`.

Maybe for User container model we just need to use the same `PermissionManager` as we use for User component.

I don't think we should deprecate the method `Yii::$app->user->can()` because it is used in many place and it is easier to write instead of `Yii::$app->user->getIdentity()->can()`.
We just need to find a solution to make result the same. For me it looks like the `ContentContainerPermissionManager` works correctly only for Space, but for model User it should not be applied completely, so we should use the simple `PermissionManager` as it works correctly for the web component User.
Or do you remember why the `ContentContainerPermissionManager` should be used for model User?

Also I find in 3 places like this one https://github.com/humhub/humhub/blob/master/protected/humhub/modules/user/models/User.php#L837 the following using `(new PermissionManager(['subject' => $this]))->can(ManageUsers::class)`, so I decided to use this way as fix in the commit https://github.com/humhub/humhub/pull/6012/commits/d8c1c52f3c06690bf6cb03ed9bc6e26a26bab390#diff-2575cda244a6a074e97e4c57c0249ed08b61d219f392dd082f33af33435d8f79L87, and here I could implement a test for invisible user status.

---

> It might also make sense to rename `humhub\modules\user\components\User` to `humhub\modules\user\components\WebUser`.

Yes, we can rename it, I don't find it is used from external modules, so the changes will be only in the core.
Should I rename it?

luke-:
@yurabakhtin 



> @luke-
> 
> > Do you have any idea why the behavior of the two `can()` methods is different?
> 
> I still cannot find why `Yii::$app->user->getIdentity()->can(ManageUsers::class)` returns `false` when `true` is expected. But the dirrence is that the component `Yii::$app->user` uses `new PermissionManager(['subject' => $this->getIdentity()])`, and the model `Yii::$app->user->getIdentity()` uses `new ContentContainerPermissionManager(['contentContainer' => $this, 'subject' => Yii::$app->user->getIdentity()])`.
> 
> Maybe for User container model we just need to use the same `PermissionManager` as we use for User component.
> 
> I don't think we should deprecate the method `Yii::$app->user->can()` because it is used in many place and it is easier to write instead of `Yii::$app->user->getIdentity()->can()`. We just need to find a solution to make result the same. For me it looks like the `ContentContainerPermissionManager` works correctly only for Space, but for model User it should not be applied completely, so we should use the simple `PermissionManager` as it works correctly for the web component User. Or do you remember why the `ContentContainerPermissionManager` should be used for model User?
> 
> Also I find in 3 places like this one https://github.com/humhub/humhub/blob/master/protected/humhub/modules/user/models/User.php#L837 the following using `(new PermissionManager(['subject' => $this]))->can(ManageUsers::class)`, so I decided to use this way as fix in the commit [d8c1c52#diff-2575cda244a6a074e97e4c57c0249ed08b61d219f392dd082f33af33435d8f79L87](https://github.com/humhub/humhub/commit/d8c1c52f3c06690bf6cb03ed9bc6e26a26bab390#diff-2575cda244a6a074e97e4c57c0249ed08b61d219f392dd082f33af33435d8f79L87), and here I could implement a test for invisible user status.



> 
> > It might also make sense to rename `humhub\modules\user\components\User` to `humhub\modules\user\components\WebUser`.
> 
> Yes, we can rename it, I don't find it is used from external modules, so the changes will be only in the core. Should I rename it?



> @luke-
> 
> > Do you have any idea why the behavior of the two `can()` methods is different?
> 
> I still cannot find why `Yii::$app->user->getIdentity()->can(ManageUsers::class)` returns `false` when `true` is expected. But the dirrence is that the component `Yii::$app->user` uses `new PermissionManager(['subject' => $this->getIdentity()])`, and the model `Yii::$app->user->getIdentity()` uses `new ContentContainerPermissionManager(['contentContainer' => $this, 'subject' => Yii::$app->user->getIdentity()])`.
> 
> Maybe for User container model we just need to use the same `PermissionManager` as we use for User component.
> 
> I don't think we should deprecate the method `Yii::$app->user->can()` because it is used in many place and it is easier to write instead of `Yii::$app->user->getIdentity()->can()`. We just need to find a solution to make result the same. For me it looks like the `ContentContainerPermissionManager` works correctly only for Space, but for model User it should not be applied completely, so we should use the simple `PermissionManager` as it works correctly for the web component User. Or do you remember why the `ContentContainerPermissionManager` should be used for model User?
> 

I can also check if I can find the error here. But basically `ContentContainerActiveRecord::can()` should work correctly for Spaces and User.

We can keep `Yii::$app->user->can()`, but this should only be a shortcut for `Yii::$app->user->identity->can()`.

I'll take a look later...

 
> > It might also make sense to rename `humhub\modules\user\components\User` to `humhub\modules\user\components\WebUser`.
> 
> Yes, we can rename it, I don't find it is used from external modules, so the changes will be only in the core. Should I rename it?

Yes, please do this. If this has unexpected side effects, we may have to change it back.

luke- (approved):

luke-:
@yurabakhtin I'll create another PR for further changes. Thanks for this PR Fix for the master.

yurabakhtin:
@luke- Sorry, but this merged PR has one bug for guest users with code `$user->id` because `$user` is `null`, please merge new fix from the PR https://github.com/humhub/humhub/pull/6036.

luke-:
@yurabakhtin Done. thanks!

Issue #6018, opened 2023-01-05
Profile fields textarea and checkbox list are not disabled for editing


yurabakhtin:
### What steps will reproduce the problem?

Create new profile fields with unchecked option "Editable":

![editable-option](https://user-images.githubusercontent.com/6995524/210778130-71d65d3e-c33e-418a-ab8d-5fa9386929f9.png)

Go to edit the fields from account profile

### What is the expected result?

All fields should disabled.

Also if they are not "Visible" then they should hidden as well, they may be visible and available for editing only by admin.

### What do you get instead?

Textarea and Checkbox list are not disabled:

![disabled-issue](https://user-images.githubusercontent.com/6995524/210778550-936a6860-a60c-4336-b915-5f924d118e99.png)

### Additional info

| Q                | A
| ---------------- | ---
| HumHub version   | 1.13.1

yurabakhtin:
@luke- Fixed in PR https://github.com/humhub/humhub/pull/6020:

![disabled-profile-fields](https://user-images.githubusercontent.com/6995524/210937453-917e7bc7-2dee-48e2-91f0-9570b139ffea.png)

funkycram:
Could we have the possibility to disable checkbox list and radio list fields for any models (not only for profile fields)?
Eg: `$form->field($model, 'attribute_name')->checkboxList([1 => 'one', 2 => 'two'], ['disabled' => 'disabled'])`
Or should I create a separate issue?
Thanks!

luke-:
@funkycram A seperate issue would be great!

funkycram:
> @funkycram A seperate issue would be great!

Done: https://github.com/humhub/humhub/issues/6083
Thanks!

luke-:
@funkycram This issue can be closed right?

funkycram:
> @funkycram This issue can be closed right?

Yes, it works fine for profile fields!

PR #6020, opened 2023-01-05
Fix option "Editable" of the profile fields

yurabakhtin:

- [x] Bugfix
- [x] Code style update
- [x] Refactor

- [x] All tests are passing
- [x] Changelog was modified

**Other information:**
Issue: https://github.com/humhub/humhub/issues/6018


luke- (changes requested):
@yurabakhtin Thanks, can you please add some tests for it? Since the change is a bit bigger than expected and the bug is not critical, maybe we should add it to the `develop` branch.

yurabakhtin:
> @yurabakhtin Thanks, can you please add some tests for it? Since the change is a bit bigger than expected and the bug is not critical, maybe we should add it to the `develop` branch.

@luke- Commit https://github.com/humhub/humhub/pull/6020/commits/a2c464914ccedeb4fa6604e8202354340303ef7a

luke- (approved):

Issue #5699, opened 2022-05-14
Allow users to invite by link


samuk:
### Is your feature request related to a problem? Please describe.

People are unlikely to click on an anonymous/impersonal email.

### Describe the solution you'd like

When sending invites the subject and body of the email can be edited by the person doing the invitation

### Describe alternatives you've considered

None

samuk:
@luke- Would you accept a pull for this? If so would you have particular needs around how it was implemented?

luke-:
@samuk I would be a bit concerned that the emails could then be abused as spam if normal users can freely define the recipient and content.

Can you maybe make some mockups how you imagine it from the interface.

Otherwise I am open for a PR.

funkycram:
Maybe this new field could be displayed only to space admin and moderators?

samuk:
@funkycram it's the personal "Hi Julie this is a really great space, We're discussing XYZ on it and thought of you" that I think will create trust/click-throughs in the recipients.

I take Lukes's point on spam though. It might be a good use for a [badge/reputation system](https://meta.discourse.org/badges) like Discourse has, although I guess that would be a lot of work.

A moderator defined message would be better than the current message IMHO.

funkycram:
If the field is displayed only to space admin and moderators, we don't need badge reputation.
I suggest having a pre-filled field such as this:
![image](https://user-images.githubusercontent.com/23310825/169854049-862f42c7-dd19-48ee-9cfa-f196b26f5f7f.png)

samuk:
Yes looks good to me.

Having it only visible to mods/admin is less-good than having it exposed to users, but let's go ahead and do this if Luke's happy with it.

A user reputation system would be powerful in its own right, but let's discuss this elsewhere.

luke-:
What do you guys think about an invite link that you can just share personally?

Somehow I'm not a big fan of the solution with the placeholders + only for mods/admins.

samuk:
Sorry for slow reply.

I think an invite link would be a good thing to have.

I wonder if it could be presented with a mailto option? Something like;

----------

Share this link with your friends: https://somehumhubinstance.com/space/invitecode

Or send via email <a href="mailto:yourfriend@gmail.com&subject=Really%fun%social%networkl
&body=I'm%really%liking%this%new%social%space%come%join%mehttps://somehumhubinstance.com/space/invitecode</a>

Help to <a href="https://support.google.com/a/users/answer/9308783?hl=en>send mail from your Gmail</a>

[...]

PR #6087, opened 2023-02-08
Fix for Allow users to invite by link #5699

funkycram:
Fix for https://github.com/humhub/humhub/issues/5699

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**

### DIFF
--- a/protected/humhub/modules/admin/controllers/PendingRegistrationsController.php
+++ b/protected/humhub/modules/admin/controllers/PendingRegistrationsController.php
@@ -59,7 +59,8 @@ class PendingRegistrationsController extends Controller
             'searchModel' => $searchModel,
             'types' => [
                 null => null,
-                PendingRegistrationSearch::SOURCE_INVITE => Yii::t('AdminModule.base', 'Invite'),
+                PendingRegistrationSearch::SOURCE_INVITE => Yii::t('AdminModule.base', 'Invite by email'),
+                PendingRegistrationSearch::SOURCE_INVITE_BY_LINK => Yii::t('AdminModule.base', 'Invite by link'),
                 PendingRegistrationSearch::SOURCE_SELF => Yii::t('AdminModule.base', 'Sign up'),
             ]
         ]);
--- a/protected/humhub/modules/admin/models/PendingRegistrationSearch.php
+++ b/protected/humhub/modules/admin/models/PendingRegistrationSearch.php
@@ -76,7 +76,7 @@ class PendingRegistrationSearch extends Invite
         $query->andFilterWhere(['like', 'username', $this->getAttribute('originator.username')]);
         $query->andFilterWhere(['like', 'user_invite.email', $this->email]);
         $query->andFilterWhere(['like', 'user_invite.language', $this->language]);
-        $query->andFilterWhere(['like', 'source', $this->source]);
+        $query->andFilterWhere(['source' => $this->source]);
 
         return $dataProvider;
     }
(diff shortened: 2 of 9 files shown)


luke-:
@funkycram  The option `enableEmailField` is intended for the admin interface or automatic registrations (via LDAP/OAuth) and does not validate the email address. 

We need to find another workflow here.

a.) The easiest way: The initial secret link URL does not lead to the 'registration' directly instead to a new dialog where the user first only has to enter the email. Then he receives an e-mail with a link, validates his e-mail address there and the `/registration` starts as before.

b.) We introduce a new dialogue, directly after the login/registration. Similar to TwoFactor Auth. There, users without an email address are asked to enter an email address and verify it directly. (This step could be omitted if E-Mail Required is set to False). The disadvantage here would be that users could not reset their password until they have set  a validated e-mail.

c.) We integrate an e-mail validation directly into the registration form. 

d.) We only allow the "Register by Link" feature when `$userModule->emailRequired` is set to `false`.

funkycram:
Thanks @luke- for these proposals. 

As the registration process is already a little complex, I think it's better to separate the registration by link. So I took option B.

It works also if User::emailRequired is set to false (I redirect to the registration form).

I also separated the different registration types here (and fixed the value which was already `1` in the Source column):
image

What to you think about this PR?

luke-:
@funkycram Thanks, looks good, that's how I thought of option A. :-)

Can you perhaps add a few (acceptance) tests for this?

Especially for such features  with different edge cases (e.g. `emailRequired` true/false) it would be good to have  such things covered by tests.

funkycram:
I have no experience in test management...
However, I can try, but I have limited time, because @samuk is the one who pay me for this job and I already spend quite some time one it...

As far as I can see, I could add something like this in https://github.com/humhub/humhub/blob/develop/protected/humhub/modules/user/tests/codeception/acceptance/InviteCest.php:

```php

    /**
     * @param AcceptanceTester $I
     * @throws \Exception
     */
    public function testDashboardInviteByLink(AcceptanceTester $I)
    {
        $I->wantTo('ensure that inviting external users by link works.');

        $I->amUser();
        DirectoryMemberPage::openBy($I);

        $I->click('Invite new people');
        $I->waitForText('Invite new people', null, '#globalModal');
        $I->click('Invite by link');
        $I->waitForText('You can invite external users who are currently not registered via link.');
    }
```

But I understand that the main process you want to test is the registration part. But I cannot find these tests in `acceptance`. It's more in https://github.com/humhub/humhub/blob/develop/protected/humhub/modules/user/tests/codeception/functional/RegistrationCest.php

luke- (approved):

luke-:
@funkycram  I've added some basic tests here: https://github.com/humhub/humhub/commit/dbf39aa49707bfd1a1b6d973f1aac51360b43ffb

I have removed the support when no email address is required. This added too much  special handling for this edge case (e.g. space ID in session with approval).  See here: https://github.com/humhub/humhub/commit/f860b27f2f182b8113fed6dc6bf03d15e12740bf

Can you please check this.

funkycram:
@luke- A very big thank you for the work you have done!
And I see that it was quite a job!
Next time, I will plan the time to spend to add or update the automatic tests.
I did a quick test and invitations by link works well.
Thanks again! Marc.

PR #6116, opened 2023-02-17
Scheduled content publishing

yurabakhtin:

- [x] Feature

- [x] All tests are passing
- [x] Changelog was modified

**Other information:**
Issue: https://github.com/humhub/team_tasks/issues/274

### DIFF
--- a/protected/humhub/modules/content/controllers/ContentController.php
+++ b/protected/humhub/modules/content/controllers/ContentController.php
@@ -12,19 +12,18 @@ use humhub\components\behaviors\AccessControl;
 use humhub\components\Controller;
 use humhub\modules\content\models\Content;
 use humhub\modules\content\models\forms\AdminDeleteContentForm;
+use humhub\modules\content\models\forms\ScheduleOptionsForm;
 use humhub\modules\content\Module;
 use humhub\modules\content\notifications\ContentDeleted;
 use humhub\modules\content\permissions\CreatePublicContent;
 use humhub\modules\content\widgets\AdminDeleteModal;
 use humhub\modules\stream\actions\StreamEntryResponse;
 use Yii;
-use yii\base\BaseObject;
 use yii\base\Exception;
 use yii\base\InvalidConfigException;
 use yii\web\BadRequestHttpException;
 use yii\web\ForbiddenHttpException;
 use yii\web\HttpException;
-use yii\web\NotAcceptableHttpException;
 use yii\web\NotFoundHttpException;
 use yii\web\Response;
 
@@ -437,5 +436,20 @@ class ContentController extends Controller
         }
 
         return $this->asJson($json);
+    }
+
+    public function actionScheduleOptions($id = null)
+    {
+        $scheduleOptions = new ScheduleOptionsForm([
+            'content' => $id ? Content::findOne($id) : null
+        ]);
+
+        if ($scheduleOptions->load(Yii::$app->request->post())) {
+            $scheduleOptions->save();
+        }
+
+        return $this->renderAjax('scheduleOptions', [
+            'scheduleOptions' => $scheduleOptions
+        ]);
     }
 }
--- a/protected/humhub/modules/content/controllers/ContentController.php
+++ b/protected/humhub/modules/content/controllers/ContentController.php
@@ -446,12 +446,16 @@ class ContentController extends Controller
             'content' => $id ? Content::findOne($id) : null
         ]);
 
-        if ($scheduleOptions->load(Yii::$app->request->post())) {
-            $scheduleOptions->save();
+        if ($scheduleOptions->load(Yii::$app->request->post()) && $scheduleOptions->save()) {
+            // Disable in order to don't focus the date field because modal window will be closed anyway
+            $disableInputs = true;
+        } else {
+            $disableInputs = !$scheduleOptions->enabled;
         }
 
         return $this->renderAjax('scheduleOptions', [
-            'scheduleOptions' => $scheduleOptions
+            'scheduleOptions' => $scheduleOptions,
+            'disableInputs' => $disableInputs
         ]);
     }
 }
--- a/protected/humhub/modules/content/controllers/ContentController.php
+++ b/protected/humhub/modules/content/controllers/ContentController.php
@@ -440,6 +440,8 @@ class ContentController extends Controller
 
     public function actionScheduleOptions($id = null)
     {
+        $this->forcePostRequest();
+
         $scheduleOptions = new ScheduleOptionsForm([
             'content' => $id ? Content::findOne($id) : null
         ]);
--- a/protected/humhub/modules/content/controllers/ContentController.php
+++ b/protected/humhub/modules/content/controllers/ContentController.php
@@ -446,9 +446,9 @@ class ContentController extends Controller
             'content' => $id ? Content::findOne($id) : null
         ]);
 
-        if ($scheduleOptions->load(Yii::$app->request->post()) && $scheduleOptions->save()) {
+        if ($scheduleOptions->load(Yii::$app->request->post())) {
             // Disable in order to don't focus the date field because modal window will be closed anyway
-            $disableInputs = true;
+            $disableInputs = $scheduleOptions->save();
         } else {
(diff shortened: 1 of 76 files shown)


protected/humhub/modules/content/migrations/m230217_112400_content_scheduled_at.php
```diff
@@ -0,0 +1,29 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function safeUp()
+    {
+        $this->safeAddColumn(Content::tableName(), 'scheduled_at', $this
+            ->dateTime()
+            ->null()
+            ->after('updated_by'));
```
luke-, line 19: please before `created_at`.

protected/humhub/modules/content/widgets/ScheduleLink.php
```diff
@@ -0,0 +1,53 @@
...
+    public ContentActiveRecord $contentRecord;
+
+    /**
+     * @inheritdoc
+     */
+    public function run()
+    {
+        $content = $this->contentRecord->content;
+
+        if ((int) $content->state !== Content::STATE_SCHEDULED) {
```
luke-, line 35: Allow option also for drafts

protected/humhub/modules/content/controllers/ContentController.php
```diff
@@ -394,4 +396,25 @@ public function actionNotificationSwitch()
 
         return $this->asJson($json);
     }
+
+    public function actionScheduleOptions($id = null)
+    {
+        $this->forcePostRequest();
+
```
luke-, line 403: Is there any protection? We also need to ensure this is only possible for New or Draft Contents

luke- (changes requested):
- It should be possible to schedule also Drafts and it would be good to change the scheduling
- Can you please create an issue in Team Tasks with all new messages and screens. So Semir or Hubert can check the wordings and grammar.
- We also should support for this feature in the News and Polls module

protected/humhub/modules/content/migrations/m230217_112400_content_scheduled_at.php
yurabakhtin: @luke- Done here https://github.com/humhub/humhub/pull/6116/commits/abaa10a0d6c14a094a5f515bc7fb3d02751f5a65#diff-f8af3f993c3fdd6bb6cb7e8d3ec9149506e2e999ad0517073eb97cad7f7c8201L19.

protected/humhub/modules/content/widgets/ScheduleLink.php
yurabakhtin: @luke- Done here https://github.com/humhub/humhub/pull/6116/commits/abaa10a0d6c14a094a5f515bc7fb3d02751f5a65#diff-5180ebc7566a79028c0c165f7079fa892457d06405c0ec8a3067fe7f1453f504L35.

But please note currently it is possible only to change state from `draft` to `scheduled`, but cannot revert this back, because if uncheck the option "Enable schedule date time" then content state will be updated to `published`.
Is it ok, if we should implement a selector for a state when the checkbox is unchecked?

protected/humhub/modules/content/controllers/ContentController.php
yurabakhtin: @luke- Yes, it works well for New and Draft contents. I just decided to this this POST protection because in almost all cases we need to call this as POST request.
The action is loaded now only by modal window, and I decided to use js code `modal.post`(POST request) instead of `modal.load`(GET request).
Please review what cases we have:

- load a modal form first time for new created content: here it is possible to use GET
- submit the modal form: only POST request
- load the modal form next times for the same new content in order to do changes: here we must use only POST request, in order to load date and time from previous submitted form(need POST request for php code `ScheduleOptionsForm::load($_POST)`)
- load it for existing/stored content: possible to use GET request
- submit the changed options for the existing content: only POST request

i.e. in all cases we use the same js code which can use either `modal.load` or `modal.post` and it is easy to use in all cases only one way, so I selected to use `modal.post` in all cases and it is why I decide to use here the POST protection because I don't think the action will may be called in some cases as GET request.

yurabakhtin:
@luke- 

> - Can you please create an issue in Team Tasks with all new messages and screens. So Semir or Hubert can check the wordings and grammar.

https://github.com/humhub/team_tasks/issues/278

yurabakhtin:
@luke- 

> - We also should support for this feature in the News and Polls module

I have created issues:

- https://github.com/humhub/news/issues/56
- https://github.com/humhub/polls/issues/119

protected/humhub/modules/content/widgets/ScheduleLink.php
luke-: Hmm, what do you think about: If the user is uncheck the scheduling checkbox on a existing scheduled content, the content will be marked as draft?

protected/humhub/modules/content/controllers/ContentController.php
luke-: @yurabakhtin Thanks for the explaination. Ok, the GET/POST methods are ok. I was more concerned about whether the user should be allowed to view and edit the content. Or is that validated by the model?

protected/humhub/modules/content/widgets/ScheduleLink.php
yurabakhtin: @luke- Yes, it seems it is more sense to set state to `draft`, ok I will do this.

protected/humhub/modules/content/controllers/ContentController.php
yurabakhtin: @yurabakhtin 

> Or is that validated by the model?

Do you mean to add some additional checking like `$content->canEdit()` for this action?

luke-: @yurabakhtin Is there already any check, that the user can edit the draft or schedulded content?

yurabakhtin: @luke- No yet. I started new method `Content::canChangeState($state)`, it is used only now in the method `Content::setState()`, and the method `canChangeState()` checks only for allowed state values, e.g. if wrong state value is passed from request.

I thought to extend this method later to check if current user is a content owner then allow to change state to serveral values like `draft`, `scheduled`, but I am not sure what states should be allowed and what denied depending on user permissions, could you advise me here?

Or may be we should integrate the state checking inside the existing method `Content::canEdit()`?

yurabakhtin: @luke- Ok, done here https://github.com/humhub/humhub/pull/6116/commits/3f75a583b84335312330bd5f0c3ce56fe4101eb7#diff-3df0e8a05df0ffddaaac0a53aed588b00cdcbf899d0c1973caa3bd5d18247082R406.

protected/humhub/modules/content/widgets/StateBadge.php
```diff
@@ -18,23 +24,34 @@ class StateBadge extends Widget
...
-                ['class' => 'label label-danger label-state-deleted']
-            );
+        switch ($this->model->content->state) {
+            case Content::STATE_DRAFT:
+                return Html::tag('span', Yii::t('ContentModule.base', 'Draft'),
+                    ['class' => 'label label-danger label-state-draft']
+                );
+            case Content::STATE_SCHEDULED:
+                $scheduledDateTime = new DateTime($this->model->content->scheduled_at, new DateTimeZone('UTC'));
+                return Html::tag('span', Yii::t('ContentModule.modules', 'Scheduled at {dateTime}', [
```
luke-, line 44: We should also sue `ContentModule.base` as message category here.

luke- (changes requested):
@yurabakhtin One more thing I noticed. Can we merge it after that?

protected/humhub/modules/content/widgets/StateBadge.php
yurabakhtin: @luke- Fixed in the commit https://github.com/humhub/humhub/pull/6116/commits/87b919b497c35ebe54d17175820441db0a3e8323

luke- (approved):

PR #6160, opened 2023-03-11
Admin menu : set active entry by searching by ID instead of by URL

funkycram:
In the admin layouts, for the Spaces and Settings menu entries of the main Admin menu, the related menu entry is marked as active.
The entry is searched by URL.
I suggest to search it by ID (see [here](https://github.com/humhub/humhub/blob/develop/protected/humhub/modules/admin/widgets/AdminMenu.php#L82)).

For Humhub, while URL are not changed, this PR won't change anything.
But it would be useful for modules, if the URL is changed.

Thanks!

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**

### DIFF
--- a/protected/humhub/modules/admin/views/layouts/setting.php
+++ b/protected/humhub/modules/admin/views/layouts/setting.php
@@ -1,5 +1,5 @@
 <?php
-\humhub\modules\admin\widgets\AdminMenu::markAsActive(['/admin/setting']);
+\humhub\modules\admin\widgets\AdminMenu::markAsActive('settings');
 ?>
 
 <?php $this->beginContent('@admin/views/layouts/main.php') ?>
--- a/protected/humhub/modules/admin/views/layouts/space.php
+++ b/protected/humhub/modules/admin/views/layouts/space.php
@@ -1,5 +1,5 @@
 <?php
-\humhub\modules\admin\widgets\AdminMenu::markAsActive('space');
+\humhub\modules\admin\widgets\AdminMenu::markAsActive('spaces');
 ?>
 
 <?php $this->beginContent('@admin/views/layouts/main.php') ?>
--- a/protected/humhub/modules/admin/views/layouts/space.php
+++ b/protected/humhub/modules/admin/views/layouts/space.php
@@ -1,5 +1,5 @@
 <?php
-\humhub\modules\admin\widgets\AdminMenu::markAsActive(['/admin/space']);
+\humhub\modules\admin\widgets\AdminMenu::markAsActive('space');
 ?>
 
 <?php $this->beginContent('@admin/views/layouts/main.php') ?>


luke-:
@funkycram I don't fully understand the reason, isn't it more reliable to highlight/find the navigation entry based on the URL? Are IDs absolutely necessary for menu entries?

funkycram:
I think searching by URL and by ID are both reliable as [the ID](https://github.com/humhub/humhub/blob/develop/protected/humhub/modules/admin/widgets/AdminMenu.php#L82) will nether change and the URL nether changes either on Humhub core.

But if we want to change the URL of the entry with a module, then the entry is not marked as active anymore. But by searching by ID it's sill marked active.

Why this need for me? 2 reasons:

1) People have difficulties to find the modules configuration because they need to go to Admin -> Modules, search the module among a long list and click on "Configure".

So for some CUZY modules where the settings can be often changed, I add a tab in the Admin -> Settings section:
image

If the user is allowed to manage module but not settings, the module changes the URL of the Settings:
image

2) On some Humhub instances, I need to allow some sub-admin users to configure modules but not disable or uninstall them. I don't want them to access to the main Humhub settings.
So the solution 1) is perfect for this usecase. I only have to disable the Modules manager page via a custom module.

Sorry, it's a little complex to explain my use cases.
If you think searching by ID is a bad ID, no problem, I'll find another solution.
Thanks!

luke-:
@funkycram Ok got it, then let's switch to ids here.

Btw. We're going to redesign the the module administration in the next version to make Module Configurations more visible.

Little preview:
Modules which are not installed yet, are displayed in the old design.
image

luke- (approved):

ArchBlood:
> @funkycram Ok got it, then let's switch to ids here.
> 
> Btw. We're going to redesign the the module administration in the next version to make Module Configurations more visible.
> 
> Little preview:
> Modules which are not installed yet, are displayed in the old design.
> image
> 
> 
> 

A suggestion when it comes to the module configuration, wouldn't it be better to force the user to use a pop-up model that then displays all of said module's configures?

funkycram:
@luke- Awesome this redesign! I really like it much more (left menu present, activated module on top, not all available modules displayed on the same page).
Do you think it would be an idea (if yes, I'll create a separated issue) to split the "Manage module" group permission in 2:
- Configure module  / Can access to the modules' settings and default activation on spaces and users
- Manage module installation / Can install, uninstall, activate and deactivate modules
Alternatively, the "Manage Settings" permission will allow to configure modules too.
As deactivating a module is really dangerous (all data lost).

@ArchBlood Modal boxes are great, but then you cannot give direct URL to someone. And sometimes, in the module configuration page, you want to have modal box too, and Humhub doesn't support 2 modal boxes layers.

luke-:
I'm a little concerned that it will be too many permissions and then become very confusing, but I can understand the use case.

Feel free to open an issue on this and it would definitely be helpful.

PR #6180, opened 2023-03-18
Fix #5537: Space Starting Page gets lost when changing URL

### LINKED ISSUE #5537: Space: Starting Page gets lost when changing URL

1. Enter a starting page for Members
2. Define a Starting Page for Guests shouldn't be Stream (Default to see the problem!)
3. Save
4. now go again and change the path segment to something else
5. Save

The already defined Starting Pages should not be changed when the Path segment gets changed.

### What do you get instead?
as you can see both - Members and Guests - will have their Starting Pages reset to default. The correct previous entries are lost.

### Additional info

| Q                | A
| ---------------- | ---
| HumHub version   | 1.10.3 - newestupdates via GITmaster?
| PHP version      |  8.0
| Operating system | Ubuntu 20.04 LTS

Screen Shot 2022-02-04 at 11 52 22

https://community.humhub.com/file/file/download?guid=757ee244-e600-41cd-bd6d-c9f7cc5c0ff

gevorgmansuryan:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**


protected/humhub/modules/space/models/AdvancedSettings.php
```diff
@@ -130,6 +130,24 @@ public function save(): bool
         $settings = $this->space->getSettings();
 
         $this->space->url = $this->url;
+
+        if ($this->space->isAttributeChanged('url')) {
```
luke-, line 134: Instead of saving the URL, would it perhaps be possible to save the route of the `index` page?

luke- (changes requested):

protected/humhub/modules/space/models/AdvancedSettings.php
luke-: Only if this is possible to implement

gevorgmansuryan: @luke- it's not only index page, so We need database migration to change all saved links to routes
A little bit change \humhub\modules\ui\menu\MenuLink and usages to have route also (currently it's only stored  href as string). This need to be done in core and in all modules which adds links to space left menu

PR #6190, opened 2023-03-23
Clear Module Settings Cache after deleting settings from database

### LINKED ISSUE #6189: Module settings survive deactivation in cache
- Install an optional module that allows settings to be stored, e.g. [HumHub Updater](https://marketplace.humhub.com/module/updater/description)
- Set a setting to a non-default value, e.g. for the updater, set the channel to "beta"
- Deactivate the module
- Re-activate the module
- Check the value you've set (it will be as you set it before, e.g. "beta")
- Clear the cache
- Check the value again (it will now be the default value, e.g. "stable")

- After reactivation, the setting should be again set to it's default value (e.g. "stable" in the above example)
- Deleting the cache should not make any difference

### What do you get instead?
- The setting is as it was before deactivating the module.
- In order to get the default value, you need to clear the cache first

### Additional info

| Q                | A
| ---------------- | ---
| HumHub version   | 1.13.2
| PHP version      | 8.1
| Operating system | Alpine Linux on docker on Ubuntu on WSL

martin-rueegg:

- Bugfix

- No

- [x] It's submitted to the the `master` branch, as it _is_ a hotfix
- [x] Fix #6189
- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
Happy to get feedback and make adjustments, if required.

### DIFF
--- a/CHANGELOG-DEV.md
+++ b/CHANGELOG-DEV.md
@@ -7,6 +7,7 @@ HumHub Changelog
 - Fix #6192: Where Group::getAdminGroupId() would sometimes return int, sometimes string
 - Enh #6260: Improve migration class
 - Fix #6199: Module manager Add types to properties
+- Fix #6189: Module settings survive deactivation in cache
 - Enh #6236: Logging: Show log entries from migrations with category migration
 - Fix #6216: Spaces icon in admin menu
 - Fix #6229: Bug on saving forms: Zend OPcache API is restricted by "restrict_api"
--- a/protected/humhub/components/SettingActiveRecord.php
+++ b/protected/humhub/components/SettingActiveRecord.php
@@ -1,0 +1,61 @@
+<?php
+
+/**
+ * @link      https://www.humhub.org/
+ * @copyright Copyright (c) 2023 HumHub GmbH & Co. KG
+ * @license   https://www.humhub.com/licences
+ */
+
+namespace humhub\components;
+
+use Yii;
+use yii\db\ActiveRecord;
+
+/**
+ * BaseSetting
+ *
+ * @since  1.13.2
+ * @author Martin Rüegg <martin.rueegg@metaworx.ch>
+ */
+abstract class SettingActiveRecord extends ActiveRecord
+{
+    /**
+     * @const array List of fields to be used to generate the cache key
+     */
+    protected const CACHE_KEY_FIELDS = ['module_id'];
+
+    /**
+     * @const string Used as the formatting pattern for sprintf when generating the cache key
+     */
+    protected const CACHE_KEY_FORMAT = 'settings-%s';
+
+    public static function deleteAll($condition = null, $params = [])
+    {
+        // get a grouped list of cache entries that are going to be deleted, grouped by static::CACHE_KEY_FIELDS
+        $containers = self::find()
+            ->where($condition, $params)
+            ->groupBy(static::CACHE_KEY_FIELDS)
+            ->select(static::CACHE_KEY_FIELDS)
+            ->all();
+
+        // going through that list, deleting the respective cache
+        array_walk($containers, static function (ActiveRecord $rec) {
+            $key = static::getCacheKey(...array_values($rec->toArray()));
+            Yii::$app->cache->delete($key);
+        });
+
+        // proceed to delete the records from the database
+        return parent::deleteAll($condition, $params);
+    }
+
+    /**
+     * @param string $moduleId  Name of the module to create the cache key for
+     * @param mixed  ...$values Additional arguments, if required by the static::CACHE_KEY_FORMAT
+     *
+     * @return string The key used for cache operation
+     */
+    public static function getCacheKey(string $moduleId, ...$values): string
+    {
+        return sprintf(static::CACHE_KEY_FORMAT, $moduleId, ...$values);
+    }
+}
--- a/protected/humhub/components/SettingsManager.php
+++ b/protected/humhub/components/SettingsManager.php
@@ -105,7 +105,7 @@ class SettingsManager extends BaseSettingsManager
     /**
      * @inheritdoc
      */
-    public function get($name, $default = null)
+    public function get(string $name, $default = null)
     {
         if ($this->isFixed($name)) {
             return Yii::$app->params['fixed-settings'][$this->moduleId][$name];
--- a/protected/humhub/libs/BaseSettingsManager.php
+++ b/protected/humhub/libs/BaseSettingsManager.php
@@ -8,6 +8,7 @@
 
 namespace humhub\libs;
 
+use humhub\components\SettingActiveRecord;
 use Yii;
 use yii\base\Component;
 use yii\base\Exception;
@@ -24,21 +25,20 @@ use yii\helpers\Json;
  */
 abstract class BaseSettingsManager extends Component
 {
-
     /**
-     * @var string module id this settings manager belongs to.
+     * @var string|null module id this settings manager belongs to.
      */
-    public $moduleId = null;
+    public ?string $moduleId = null;
 
     /**
      * @var array|null of loaded settings
      */
-    protected $_loaded = null;
+    protected ?array $_loaded = null;
 
     /**
      * @var string settings model class name
      */
-    public $modelClass = 'humhub\models\Setting';
+    public string $modelClass = 'humhub\models\Setting';
 
     /**
      * @inheritdoc
@@ -94,11 +94,10 @@ abstract class BaseSettingsManager extends Component
     /**
      * Can be used to set object/arrays as a serialized values.
      *
-     *
      * @param string $name
      * @param mixed $value array or object
      */
-    public function setSerialized($name, $value)
+    public function setSerialized(string $name, $value)
     {
         $this->set($name, Json::encode($value));
     }
@@ -109,7 +108,7 @@ abstract class BaseSettingsManager extends Component
      * @param string $name
      * @param mixed $default the setting value or null when not exists
      */
-    public function getSerialized($name, $default = null)
+    public function getSerialized(string $name, $default = null)
     {
         $value = $this->get($name, $default);
         if (is_string($value)) {
@@ -126,20 +125,22 @@ abstract class BaseSettingsManager extends Component
      * Returns value of setting
      *
      * @param string $name the name of setting
+     *
      * @return string|null the setting value or null when not exists
      */
-    public function get($name, $default = null)
+    public function get(string $name, $default = null)
     {
-        return isset($this->_loaded[$name]) ? $this->_loaded[$name] : $default;
+        return $this->_loaded[$name] ?? $default;
     }
 
     /**
      * Returns the value of setting without any caching
      *
      * @param string $name the name of setting
+     *
      * @return string the setting value or null when not exists
      */
-    public function getUncached($name, $default = null)
+    public function getUncached(string $name, $default = null): ?string
     {
         $record = $this->find()->andWhere(['name' => $name])->one();
         return ($record !== null) ? $record->value : $default;
@@ -150,7 +151,7 @@ abstract class BaseSettingsManager extends Component
      *
      * @param string $name
      */
-    public function delete($name)
+    public function delete(string $name)
     {
         $record = $this->find()->andWhere(['name' => $name])->one();
         if ($record !== null) {
@@ -166,8 +167,6 @@ abstract class BaseSettingsManager extends Component
         if (isset($this->_loaded[$name])) {
             unset($this->_loaded[$name]);
         }
-
-        $this->invalidateCache();
     }
 
     /**
@@ -180,7 +179,7 @@ abstract class BaseSettingsManager extends Component
             $this->_loaded = [];
             $settings = &$this->_loaded;
 
-            array_map(function ($record) use (&$settings) {
+            array_map(static function ($record) use (&$settings) {
                 $settings[$record->name] = $record->value;
             }, $this->find()->all());
 
@@ -212,9 +211,11 @@ abstract class BaseSettingsManager extends Component
      *
      * @return string the cache key
      */
-    protected function getCacheKey()
+    protected function getCacheKey(): string
     {
-        return 'settings-' . $this->moduleId;
+        /** @var SettingActiveRecord $modelClass */
+        $modelClass = $this->modelClass;
+        return $modelClass::getCacheKey($this->moduleId);
     }
 
     /**
@@ -222,7 +223,7 @@ abstract class BaseSettingsManager extends Component
      */
     protected function createRecord()
     {
-        $model = new $this->modelClass;
+        $model = new $this->modelClass();
         $model->module_id = $this->moduleId;
 
         return $model;
--- a/protected/humhub/models/Setting.php
+++ b/protected/humhub/models/Setting.php
@@ -8,8 +8,8 @@
 
 namespace humhub\models;
 
+use humhub\components\SettingActiveRecord;
 use Yii;
-use yii\db\ActiveRecord;
 use yii\base\Exception;
 
 /**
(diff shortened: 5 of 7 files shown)


luke-:
@martin-rueegg Thanks!

martin-rueegg:
The reasoning for `\humhub\libs\BaseSetting::getCacheKey()` was to have a single source of truth. Rather than creating the key both in the `BaseSetting` as well as in the `\humhub\libs\BaseSettingsManager::getCacheKey()` ...

Also, I tried to do it as flexible as reasonably possible, so that future child classes could still implement the same design pattern.

ArchBlood:
@luke- I would like to note that most of the Setting model is deprecated and we should find a better way for a workaround than using this class, possible solution being migrated to a new class or separating into two classes to keep backwards compatible for v1.1 migration.

martin-rueegg:
> @luke- I would like to note that most of the Setting model is deprecated

Thanks, @ArchBlood. Could you give some indication in what sense the Settings model is deprecated or where to find information on how to do things the right was. As we are in the progress of developing a new module and are currently relating on the Settings model (through the SettingsManager) quite intensely, it would be good to know what to watch out for.

ArchBlood:
You can find which parts of the Setting model that are deprecated by looking for parts that read `@deprecated since version 1.1`

ArchBlood:
As for the `get` method, you can get a module `id` by using `Yii::$app->getModule($moduleId);`, `$moduleId` being the module `id`.

martin-rueegg:
> You can find which parts of the Setting model that are deprecated by looking for parts that read `@deprecated since version 1.1`

Thanks! So basically _every_ method is deprecated, unless it is required for the `ActiveRecord` to work, or it is `Temporary for 1.1. migration`. 🤣 There is only one exception: `Setting::isInstalled()` ...

Hence the correct approach would be to use the `Setting::$value` property, rather than the `Setting::get()` and `Setting::set()` methods, is that correct? As such, the `\humhub\models\Setting` class itself is _not_ deprecated, is that also correct?

If that's the case, I do not see why the proposed changes would mess with the deprecation of the other methods, as
- I _am_ introducing a new base class
- Only overriding a method of the underlying `ActiveRecord`, namely `ActiveRecord::deleteAll()`, 
  plus adding a _new_ static method to get the `getCacheKey()`, because it seemed to be the most logical place to get that information stored/from. :-)

Is there something I am missing here?

However, I notice that the namespace for the new `BaseSetting`, i.e. `humhub\libs` is poorly chosen, and should probably rather be `humhub\models`. Happy to change that, if Lucas sees the implementation to otherwise fit.

No rush though.

protected/humhub/libs/BaseSetting.php
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * @param string $moduleId  Name of the module to create the cache key for
+     * @param        ...$values Additional arguments, if required by the static::CACHE_KEY_FORMAT
+     *
+     * @return string The key used for cache operation
+     */
+    public static function getCacheKey(
+        string $moduleId,
+        ...$values
+    ): string {
```
luke-, line 69:
```suggestion
    public static function getCacheKey(string $moduleId, ...$values): string {
```

protected/humhub/libs/BaseSetting.php
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * @const string Used as the formatting pattern for sprintf when generating the cache key
+     */
+    protected const CACHE_KEY_FORMAT = 'settings-%s';
+
+    public static function deleteAll(
+        $condition = null,
+        $params = []
+    ) {
```
luke-, line 38:
```suggestion
    public static function deleteAll($condition = null, $params = []) {
```

protected/humhub/libs/BaseSetting.php
```diff
@@ -0,0 +1,73 @@
...
+ * BaseSetting
+ *
+ * @since  1.13.2
+ * @author Martin Rüegg <martin.rueegg@metaworx.ch>
+ */
+abstract class BaseSetting
+    extends ActiveRecord
+{
+
+    // constants
```
luke-, line 24:
```suggestion
```

protected/humhub/libs/BaseSetting.php
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * BaseSetting
+ *
+ * @since  1.13.2
+ * @author Martin Rüegg <martin.rueegg@metaworx.ch>
+ */
+abstract class BaseSetting
+    extends ActiveRecord
+{
```
luke-, line 22:
```suggestion
abstract class BaseSetting extends ActiveRecord
{
```

protected/humhub/libs/BaseSetting.php
```diff
@@ -0,0 +1,73 @@
...
+use Yii;
+use yii\db\ActiveRecord;
+
+/**
+ * BaseSetting
+ *
+ * @since  1.13.2
+ * @author Martin Rüegg <martin.rueegg@metaworx.ch>
+ */
+abstract class BaseSetting
```
luke-, line 20:
```suggestion
abstract class SettingActiveRecord
```

For base ActiveRecord classes, we currently always add the suffix "ActiveRecord". Normally we also add these in the `components` namespace, but in this case `libs` is also fine.

luke- (changes requested):
@martin-rueegg Thanks, looks fine. A few code style requests...  :-)

martin-rueegg:
> Thank you, looks good! Do you have any idea why the tests fail? Otherwise, I can also take a closer look too.

Nope, seems not have anything to do with the changes, does it?
https://github.com/humhub/humhub/actions/runs/4807191287/jobs/8555632086?pr=6190#step:18:1408

However, I'm currently trying to get my test environment up and running.

luke-:
@martin-rueegg I guess that the failed test has something to do with the PR.

This seems to work like before:
https://github.com/humhub/humhub/blob/master/protected/humhub/modules/user/tests/codeception/acceptance/LoginCest.php#L72-L78
https://github.com/humhub/humhub/blob/master/protected/humhub/modules/user/models/User.php#L990-L994

Maybe some problem with caching?

martin-rueegg:
> @martin-rueegg I guess that the failed test has something to do with the PR.
> 
> This seems to work like before: https://github.com/humhub/humhub/blob/master/protected/humhub/modules/user/tests/codeception/acceptance/LoginCest.php#L72-L78 https://github.com/humhub/humhub/blob/master/protected/humhub/modules/user/models/User.php#L990-L994
> 
> Maybe some problem with caching?

I've been spending most of my afternoon on testing this stuff. But on my local system, with a vanilla database, I do get ton's of errors. None of which has to do with my changes, but many with js.

I find the documentation on testing confusing. And I have no clue how to meaningfully step-debugging the tests.

### What I've done to try to get my test environment running was ...

#### codeception

This did not work for me (https://docs.humhub.org/docs/develop/testing/#test-environment-setup)
``` bash
composer global require "codeception/codeception=2.0.*" "codeception/specify=*" "codeception/verify=*"
```
due to unsolvable dependencies. But the following did:
``` bash
composer global require "codeception/codeception=2.*" "codeception/specify=*" "codeception/verify=*"
```
(Version `2.*` instead of `2.0.*`).

#### DB
``` sql
-- new empty database
CREATE DATABASE `humhub_test` CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
```

https://community.humhub.com/s/module-development/wiki/Test+Cheatsheet and
https://docs.humhub.org/docs/develop/testing/#test-environment-setup have different instructions regarding db connection the configuration. But I assume it's not really relevant, apart from the later being under version control.

I used the `env.php` and set up `.env`, too. Sourcing it.

Again, the two documentations differ in respect to in which directory to run the migrations, but it does not seem to make a difference, so I did it in the `@webroot`
``` bash
php "$HUMHUB_TEST_YII" migrate/up --includeModuleMigrations=1 --interactive=0
```

Then I seeded the database with:
``` bash
php $PHP_XDEBUG "$HUMHUB_TEST_YII" installer/auto
Install Site:

```
Note the empty result after "Install Site". Is that a problem?

#### Selenium
I used the docker version (from the [cheat sheet](https://community.humhub.com/s/module-development/wiki/Test+Cheatsheet))
``` bash
docker run --detach --net=host --shm-size="2g" selenium/standalone-chrome
```
rather than the java version from [test documentation](https://docs.humhub.org/docs/develop/testing/#run-acceptance-tests)
``` bash
java -Dwebdriver.chrome.driver=chromedriver.exe -jar selenium-server-standalone-2.53.0.jar
```

and the test server:
``` bash
 cd "$HUMHUB_PATH" && /usr/bin/php7.4 --server 127.0.0.1:8080 index-test.php
```

#### Tests
And finally started the tests with
``` bash
source .env

cd "$HUMHUB_PATH" \
 && rm -rf node_modules/ \
 && rm -rf protected/vendor/ \
 && composer install \
 && npm install \
 && git checkout HEAD -- protected/vendor/.gitignore

(rm -rf protected/runtime \
 && git checkout HEAD -- protected/runtime/.gitignore \
 && pushd $HUMHUB_PATH/protected/humhub/tests \
 && /usr/bin/php7.4 $HUMHUB_VENDOR_BIN/codecept run -v)
```

But already the activity tests fail:

[activity]: tests from /home/mdr/projects/humhub/htdocs/protected/humhub/tests/../modules/activity/tests

```
Activity.acceptance Tests (1) ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Modules: WebDriver, \tests\codeception\_support\WebHelper, \tests\codeception\_support\DynamicFixtureHelper
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
E ActivityStreamCest: Test test post creation activity (0.89s)


....


here were 2 errors:

---------
1) ActivityStreamCest: Test test post creation activity
 Test  ../../activity/tests/codeception/acceptance/ActivityStreamCest.php:testSimpleStream

  [Facebook\WebDriver\Exception\JavascriptErrorException] javascript error: $ is not defined
  (Session info: chrome=111.0.5563.146)


Scenario Steps:

 9. $I->executeJS("$('#contentForm_message')[0].click();") at ../../../tests/codeception/_support/AcceptanceTester.php:358
 8. // I am going to create a new post
 7. $I->amOnPage("/space/space?sguid=5396d499-20d6-4233-800b-c6c86e5fa34c") at ../../../tests/codeception/_support/AcceptanceTester.php:123
 6. $I->waitForElementVisible("#wallStream",30) at ../../../tests/codeception/_support/AcceptanceTester.php:233
 5. $I->click("#login-button") at ../../../tests/codeception/_pages/LoginPage.php:27
 4. $I->fillField("Login[password]","123qwe") at ../../../tests/codeception/_pages/LoginPage.php:26

#1  /home/mdr/projects/humhub/htdocs/protected/vendor/php-webdriver/webdriver/lib/Exception/WebDriverException.php:112
#2  /home/mdr/projects/humhub/htdocs/protected/vendor/php-webdriver/webdriver/lib/Remote/HttpCommandExecutor.php:359
#3  /home/mdr/projects/humhub/htdocs/protected/vendor/php-webdriver/webdriver/lib/Remote/RemoteWebDriver.php:605
#4  /home/mdr/projects/humhub/htdocs/protected/vendor/php-webdriver/webdriver/lib/Remote/RemoteWebDriver.php:349
#5  Codeception\Module\WebDriver->executeJS
#6  /home/mdr/projects/humhub/htdocs/protected/humhub/modules/activity/tests/codeception/_support/_generated/AcceptanceTesterActions.php:2723
#7  /home/mdr/projects/humhub/htdocs/protected/humhub/tests/codeception/_support/AcceptanceTester.php:358
#8  /home/mdr/projects/humhub/htdocs/protected/humhub/tests/codeception/_support/AcceptanceTester.php:144
#9  /home/mdr/projects/humhub/htdocs/protected/humhub/modules/activity/tests/codeception/acceptance/ActivityStreamCest.php:28
#10 stream\acceptance\ActivityStreamCest->testSimpleStream
Artifacts:

Png: /home/mdr/projects/humhub/htdocs/protected/humhub/modules/activity/tests/codeception/_output/stream.acceptance.ActivityStreamCest.testSimpleStream.fail.png
Html: /home/mdr/projects/humhub/htdocs/protected/humhub/modules/activity/tests/codeception/_output/stream.acceptance.ActivityStreamCest.testSimpleStream.fail.html

---------

```

Really not sure what to do about this javascript error. Seems like jquery is not loaded?

It also happens with the `master` and the `develop` branches.

Any hint would be appreciated.

luke-:
@martin-rueegg Thats very strange. I can't explain the error message with the missing JQuery.

The global Composer installation might be the problem. Normally I always use the Codeception version which is available in the `protected/vendor/bin` directory and managed via `composer.json`. Meanwhile we also use Codeception 4. I think we should remove this step from the docs.

I just ran the following steps to start a fresh empty HumHub test environment with successful tests:
https://gist.github.com/luke-/fb3c565f630ae7d7c2ab4c51d6ea23aa

martin-rueegg:
Ok, I can now reproduce the issue locally. Will see if I manage to debug it.

luke-:
@martin-rueegg Thanks for the update. Let me know if I can help!

luke- (approved):

martin-rueegg:
Found and fixed the culprit! :-D

Now, in order to create a test, I need to add data to the contentcontainter_setting table. To do that, a user is required.

- `setting` data is loaded from `protected/humhub/tests/codeception/fixtures/data/setting.php`.
   I thought to use this as an example to create ``setting` data is loaded from `protected/humhub/tests/codeception/fixtures/data/contentcontainer_setting.php`
- `user` data is loaded from `protected/humhub/modules/user/tests/codeception/fixtures/data/user.php`,
   which is in a module.

Hence the following questions:
- Do I need to specify which data is loaded at the beginning of testing?
- Is module data also automatically loaded, or
- can I specify to load the above `user` data or do I need to copy it (with one user only) to `protected/humhub/tests/codeception/fixtures/data/user.php` (and `contentcontainer`)?

luke-:
> Found and fixed the culprit! :-D
> 
> Now, in order to create a test, I need to add data to the contentcontainter_setting table. To do that, a user is required.
> 
> * `setting` data is loaded from `protected/humhub/tests/codeception/fixtures/data/setting.php`.
>   I thought to use this as an example to create ``setting`data is loaded from`protected/humhub/tests/codeception/fixtures/data/contentcontainer_setting.php`
> * `user` data is loaded from `protected/humhub/modules/user/tests/codeception/fixtures/data/user.php`,
>   which is in a module.
> 
> Hence the following questions:
> 
> * Do I need to specify which data is loaded at the beginning of testing?
> * Is module data also automatically loaded, or
> * can I specify to load the above `user` data or do I need to copy it (with one user only) to `protected/humhub/tests/codeception/fixtures/data/user.php` (and `contentcontainer`)?
> * Is it ok to create a unit test, if I want to test just one function, or does it need to be another kind of test, since db data is involved?

Unfortunately, I am not the test expert either. But basically the following standard fixtures should be always loaded: 
https://github.com/humhub/humhub/blob/master/protected/humhub/tests/codeception/_support/DynamicFixtureHelper.php#L121

Per suite or unit test it should be possible to define additional fixtures.

It should be also possible to load fixture data from other modules (e.g. user).  I would prefer to always work with the same defined HumHub test example users, instead of define own per test/suite/module.

A unit test would be sufficient in this case.

I hope I could help a little bit. :-)

martin-rueegg:
Ok, I'm going to make a new PR as there are now tests coming ...

luke-:
@martin-rueegg Sorry, was to fast with merging :-)

martin-rueegg:
Yes, next time I'll add WIP to the title, so you'd know. Unfortunately, I had squashed my chantes locally, so I need to sort them out now ... :-)

PR #6195, opened 2023-03-25
Fix error when no modules set

martin-rueegg:
https://docs.humhub.org/docs/admin/requirements lists php 7.4 as minimum requirement. Hence, the array properties should actually be arrays.

This caused log targets to be disabled due to the `ModuleManager::getModules()` throwing an error when `ModuleManager::$modules` would be null.

- Bugfix

- No, as there is no class that extends `ModuleManager`. Otherwise, the type `array` could be left away but just initializing the `$modules = []`

Do you want me to create an issue for this and adding a changelog entry?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ HumHub Changelog
 
 1.13.2 (Unreleased)
 -------------------------
+- Fix #6195: Fix ModuleManager disabling logging when now modules set
 - Fix #5965: Suppress log warning 'Invalid session auth key attempted for user'
 - Fix #6084: Automatic LDAP user registration broken when not all req. attributes provided
 - Fix #6104: Fix update user with not existing group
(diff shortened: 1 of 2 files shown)


luke-:
@martin-rueegg In the current master branch (`v.13`) PHP 7.3 is still intentionally set, because some internal build processes required this.

But with the soon coming version `1.14` this is `composer.json` PHP platform value already increased.  You are welcome to add the type hints in the `ModulManager` in the current `develop` branch. I will then merge it in the upcoming beta.2 version.  Thanks!

luke-:
Would be great, to have an Changelog Entry for all changes. If there is no Issue# you can also use the PR# number.

martin-rueegg:
Thanks, @luke- for the explanation.

Changes made accordingly.

luke- (approved):

PR #6201, opened 2023-03-28
Fix #6192: Where Group::getAdminGroupId() would sometimes return int, sometimes string

### LINKED ISSUE #6192: `Group::getAdminGroupId()` returns sometimes `int`, sometimes `string`

1. Make sure, the following setting is deleted:
    ```SQL
    DELETE FROM setting WHERE module_id = 'user' AND name = 'group.adminGroupId';
    ```
    Of course, you can review it first! ;-)
    ```SQL
    SELECT * FROM setting WHERE module_id = 'user' AND name = 'group.adminGroupId';
    ```
2. call [`Group::getAdminGroupId()`](https://github.com/humhub/humhub/blob/bb0d8a5f1f923148ffcd51124511604f8316e003/protected/humhub/modules/user/models/Group.php#L234) and check the type of the result. It will be `int`. As a side-effect, the id is stored in the `settings` table.
3. call [`Group::getAdminGroupId()`](https://github.com/humhub/humhub/blob/bb0d8a5f1f923148ffcd51124511604f8316e003/protected/humhub/modules/user/models/Group.php#L234) again and check the type of the result. It will be `string`, as this time it's been loaded from the `settings` table, rather than the `group` table

It would be good to get a consistent result. IDs would be expected to be of type `int`.

### Possible solutions

A quick fix would be adding `(int)` to the return line would do the job.

``` php
public static function getAdminGroupId()
{
	$adminGroupId = Yii::$app->getModule('user')->settings->get('group.adminGroupId');
	if ($adminGroupId == null) {
		$adminGroupId = self::getAdminGroup()->id;
		Yii::$app->getModule('user')->settings->set('group.adminGroupId', $adminGroupId);
	}
	return (int) $adminGroupId;
}
```
Best, the return type of the function would be `int` too, but that might break backward-compatibility.

Happy to create a PR, if that is appreciated.

### Additional considerations

However, using a setting only makes sense, if there could be multiple admin groups, which in fact does not seem possible (from the front-end, not the database, as there is no setting to make a group an admin group).

As such, it would be enough to simply cache the result, but not also adding it to the settings. Yet, removing this redundant setting would also break backward-compatibilty. However, the setting should be signed deprecate. As it will only be available once the `Group::getAdminGroupId()` was called. Using the the setting to cache the value is overkill and could cause problems in case of ambiguous values. Also, clearing the cache would not reset the setting.

### Additional info

| Q                | A
| ---------------- | ---
| HumHub version   | 1.13.1
| PHP version      | 8.2

martin-rueegg:

- Bugfix
- Refactor

- Yes, potentially for external module developers who extended `\humhub\modules\user\models\Group`
- No for this repository

- Fix #6192 
- [ ] All tests are passing
- New/updated tests are included
- Changelog was modified

### DIFF
--- a/CHANGELOG-DEV.md
+++ b/CHANGELOG-DEV.md
@@ -3,6 +3,7 @@ HumHub Changelog
 
 1.15.0 (Unreleased)
 -------------------
+- Fix #6192: Where Group::getAdminGroupId() would sometimes return int, sometimes string
 - Enh #6260: Improve migration class
 - Fix #6199: Module manager Add types to properties
 - Enh #6236: Logging: Show log entries from migrations with category migration
--- a/protected/humhub/modules/user/models/Group.php
+++ b/protected/humhub/modules/user/models/Group.php
@@ -231,10 +231,10 @@ class Group extends ActiveRecord
         return self::findOne(['is_admin_group' => '1']);
     }
 
-    public static function getAdminGroupId()
+    public static function getAdminGroupId(): int
     {
         $adminGroupId = Yii::$app->getModule('user')->settings->get('group.adminGroupId');
-        if ($adminGroupId == null) {
+        if ($adminGroupId === null) {
             $adminGroupId = self::getAdminGroup()->id;
             Yii::$app->getModule('user')->settings->set('group.adminGroupId', $adminGroupId);
         }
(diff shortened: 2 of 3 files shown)


martin-rueegg:
- Is there some documentation about how to run tests locally? I run
  ```bash
   php vendor/bin/codecept bootstrap
   php vendor/bin/codecept run
   ```
   Where the former said that the module `codeception/module-phpbrowser` would be required and I installed that. But then `run` still said that no tests where executed ...
- It is also unclear to me, if the cache is deleted by the test setup. Or if the cache state/runtime dir is uncontrolled.

martin-rueegg:
> @martin-rueegg Maybe this helps https://community.humhub.com/s/module-development/wiki/Test+Cheatsheet

Looks like I have no access to the given space. Benutzername ist `metaworx`, in case you could add me.

But duckduck helped me find the official documentation, which helped:
https://docs.humhub.org/docs/develop/testing/

Thanks!

luke- (approved):

Issue #5718, opened 2022-05-23
Placeholder in Select2 Search Input


luke-:
Add "search..." placeholder into the text box
![image](https://user-images.githubusercontent.com/4736168/169772473-413040df-3ac4-4366-a8ac-c675243f7da6.png)

Would be good if we could integrate such a placeholder text into the various Select2 DropDowns.

See: https://github.com/humhub/team_tasks/issues/224

yurabakhtin:
@luke- I see on your screenshot the field Country looks like Select2 dropdown with input field, but I see it as simple html select element on `develop` branch:

![country_field](https://user-images.githubusercontent.com/6995524/228305530-78365d33-8290-47a4-91cc-da4992f96670.png)

Should I modify the field into Select2 Dropdown firstly and then implement the placeholder for the field?

luke-:
@yurabakhtin this is strange. can you please check why in `develop` the `country` field is no longer a select2. If there is no good reason there, please convert that back to a Select2.

yurabakhtin:
@luke- Sorry, it was my bug from here https://github.com/humhub/humhub/pull/6020/files#diff-f0dddc2afb07bab05bf804dea03a3f7e2ce5c6b2a244e7e772af4b9a8412eae6R107, I forgot to merge options from `CountrySelect` because it is extended from `Select` with additional options.
Fixed here https://github.com/humhub/humhub/pull/6208/commits/0e26915f13ea0e3d70198f8a99ec0fbb495464bc#diff-f0dddc2afb07bab05bf804dea03a3f7e2ce5c6b2a244e7e772af4b9a8412eae6R107.

yurabakhtin:
@luke- Placeholder for the Country field has been added here https://github.com/humhub/humhub/pull/6208/commits/ad0ce2d5b523fb92adac0c47f34d74563726b78f#diff-4bc3a1ae05bf3a9510b1b2e24e7556f4f77a90b00fbe8ac341302caa3a0b8b32R103-R104.

Please note if a placeholder is enabled for Select2 widget then original empty option "Please select:" is hidden, and the placeholder is displayed only at the place:

![country_placeholder](https://user-images.githubusercontent.com/6995524/228443653-7d8b3b93-ca1b-4865-a125-8ff16596a524.png)

In order to select an empty value user should click here:

![empty_select](https://user-images.githubusercontent.com/6995524/228444056-439e4267-796d-4c6f-9b73-9131745feb92.png)

[...]

PR #6208, opened 2023-03-29
Use Select2 plugin for all drop-down list fields

yurabakhtin:

- [x] Bugfix (Fix for https://github.com/humhub/humhub/issues/5718#issuecomment-1487320074)
- [x] Feature

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
Issue: https://github.com/humhub/humhub/issues/5718


yurabakhtin:
@luke- 

> - All Select2 with search field should have "Search" as placeholder by default (unless it is overwritten via Data).

To implement this it would be have new field `->select2()` as I suggested `$form->field($model, 'language')->select2($languages);`

> - I don't like the solution that we need a separate PHP asset class, a separate javascript file etc. for this. Is there nothing more lightweight for this?

I see we have `humhub\assets\Select2Asset` which includes one JS file and includes `humhub\assets\Select2StyleAsset` which includes one CSS file. I don't understand why these 2 assets are splitted, maybe because only the `Select2StyleAsset` should be included into the `AppAsset` without JS file, but I am not sure why JS don't need there, do you know?
I.e. the 2 assets may be combined.
I would like to include new JS file `select2-searchInputPlaceholder.min.js` include into the `Select2Asset` but I don't think it is possible because the Select2 files are loaded from the source path `@npm/select2/dist/js`.

> - Seems you only committed the .min version, at least we need both JS file for future maintenance

Ok, I will add the original file.

luke- (approved):

Issue #3737, opened 2019-11-14
Better Module/Disable and Install behavior


buddh4:
Currently disabling a module completely removes the modules tables and data and removes the module from `module_enabled`.

It should be possible to disable a module without flushing the data. The following issues need to be considered:

-  Search
- Content (Stream)
- Integrity
- Events

If a module update or installation (migration) fails, the module should be automatically set to disabled until the issues are resolved.

Ideally, there should be a migration log category, in order to troubleshoot the issues.

starbock:
This one is pretty important, so glad to see it already being addressed.  For example, some regulatory compliance issue emerges that REAQUIRES taking some functionality off-line (e.g., Files module) while remediation steps are taken, then the module can be brought back on-line.  Deleting all content prevents this process.  Another example could be an upgrade breaking some module functionality, requiring it to be taken off-line while a fix is implemented;  without the ability to disable-without-delete it is operationally difficult to manage through this temporary condition.  There are lots of real-world cases like that, thus appreciate consideration to implement this capability.

martin-rueegg:
We too see this as a requirement and have so fare implemented it in our module internally. But it would certainly be better suited - and handled - by the core.

In fact I would see the current "installed-but-not-activated" state as something like this. And that only "uninstalling" the module would actually delete the content.

Happy to contribute to this.

@luke- ?

luke-:
The main problem here is that a module can have many records like notifications / stream entries / activities in the database. We somehow find a way to skip these records when a module is "paused".

 As with @martin-rueegg  discussed earlier a `class_map` table could be also helpful here, the table could have the following attributes `id`, `class_name`, `active`.

When a module is paused, all classes with the namespace are changed to `active=0`. On stream/activity/notification queries we can easily filter out non-active classes this way.

luke-:
Yes, `module_id` would really be better here, even if a join is more...

PR #6217, opened 2023-03-31
[WIP][RFC] Proof-of-concept for drop-in-replacement of class names in table fields

martin-rueegg:
This PR is WORK-IN-PROGRESS / PROOF OF CONCEPT / REQUEST FOR COMMENT

It is a test implementation of the feature discussed in #3737. The PR for now has the sole purpose to allow interested parties to review and discuss.

# Feature description

### Class Column Replacement

In principle, this patch allows for any table that is currently using fields that contain class names to replace them with a foreign key to the `class_map` table.

The proof of concept is shown with the following two tables:

- `content` contains the field `object_model`. This is replaced with `object_class_id` (which could as well be `object_model_id`, if desired). The default queries from the `\humhub\modules\content\models\Content` model will, however, still return a "virtual" field `object_model`. This field is transparently "translated" into `object_class_id`.
 The table has been chosen due to its major importance and central usage.
- `notification` contains _two_ fields, `class` and `source_class`, which are "translated" into `class_id` and `source_class_id` respectively.
  The table has been chosen since it contains two fields that are being linked. It shows that also this is possible.

Then, the performance can be further improved (also at a later stage), to match against class names, but their respective IDs directly. An example of which can be seen in `\humhub\modules\stream\models\filters\ContentTypeStreamFilter::apply`.

### Record exclusion / Module pausing
**Apart from the translation, the result of queries on these tables will return only those records, whose module has not been disabled** (or `paused` as it's currently coined to avoid confusion with `enable/disable` or `active/inactive`. A module is "paused" under the following condition:
- the table `module_enabled` contains an entry with the given `module_id` and the `is_paused` field is set to `0`

It is foreseen and easily implemented to also have the same effect through these conditions:
- `config.php` returns an array including `['isPaused' => true]`
(all that is required for this to work is that the setting is written to the `module_enabled` table.)

### Potential compatibilty issues

Compatibility problems could arise, where a query against a table with the given class_map reference is created without the use of its model. A case of which is seen e.g. in `\humhub\modules\stream\models\filters\DefaultStreamFilter::filterFile()` or in `\humhub\modules\content\models\ContentType::getContentTypes`.

### Further observations

There are some things that seem weird to me:
- Modules that are present in the file system, get still loaded (at least on the modules page) and an alias is reserved
- There are some inconsistencies in treating duplicate `module_id`s
- Class names, class namespaces, table names and all that stuff can be adapted to better fit the projects code style or be more meaningful or less confusing...
 

# PR Admin

- Feature
- Refactor

- No, bud need s further checking

- [ ] If yes, ToDo; please describe the impact and migration path for existing applications:

- [ ] All tests are passing
- [ ] Changelog was modified


protected/humhub/libs/ClassMapCacheModule.php
```diff
@@ -0,0 +1,40 @@
...
+namespace humhub\libs;
+
+class ClassMapCacheModule
+{
+    //  public properties
+    public ?int   $id = null;
+    public string $class_name;
+    public string $moduleId;
+
+    public function __construct(
```
ArchBlood, line 12: All one line?
```php
public function __construct($values,  string $class_name = '', string $moduleId = '')
```

protected/humhub/components/ModuleManager.php
```diff
@@ -146,79 +121,73 @@ public function registerBulk(array $configs)
...
+            $moduleInfo->isPaused     |= ($moduleInfo->isRegistered && $registration->isPaused);
         }
 
         // Not enabled and no core/installer module
-        if (!$isCoreModule && !in_array($config['id'], $this->enabledModules)) {
+        if (!$moduleInfo->isCoreModule && !$moduleInfo->getIsActive()) {
             return;
         }
 
-        // Handle Submodules
```
ArchBlood, line 188: Has submodule handler been moved somewhere else for better use?

protected/humhub/libs/ClassMapCacheModule.php
martin-rueegg: Hi @ArchBlood, thank you for taking the time to have a look at this and please accept my apologies for not having followed the project's code style guidelines. Of course, happy to apply those if the PR in general is welcomed. Do you use only CSFixer? Or is there also a repo for shared `.idea`-settings?

protected/humhub/components/ModuleManager.php
martin-rueegg: Since the `ModuleInfo->modules` property is an already-initialized array, it is not required to do the check in line 189 anymore.

Whether or not the module is a core module is also determined during instantiation of `ModuleInfo` and stored as a property thereof.

luke-:
@martin-rueegg Thanks for the PR. It looks promising. Please give me some time to take a closer look and give feedback. In the meantime, Happy Easter! :-)

protected/humhub/components/ModuleManager.php
```diff
@@ -364,25 +342,25 @@ public function getEnabledModules($options = [])
      */
     public function hasModule($id)
     {
-        return (array_key_exists($id, $this->modules));
+        return ($module = $this->modules[$id] ?? null) && !$module->isMissing && $module->getClassExists();
     }
 
     /**
-     * Returns weather or not the given module id belongs to an core module.
+     * Returns weather or not the given module id belongs to a core module.
```
ArchBlood, line 349: Can this be fixed to say `whether` as currently this is the wrong word used, I'm surprised that this has gone unchecked for so long.

protected/humhub/components/ModuleManager.php
```diff
@@ -416,7 +447,7 @@ public function flushCache()
     }
 
     /**
-     * Checks the module can removed
+     * Checks whether the module can be removed
```
ArchBlood, line 450: `Checks whether or not the module can be removed`

martin-rueegg:
As it turns out, any module can store class names of other modules in their own tables. As do the core modules, e.g. `comment` and `like`.

If a module is deactivated, and those records are not cleaned in the `Module::disable()` or `uninstall.php` routines, then this results in unforeseen crashes.

I guess the issue addressed by this PR should be considered significant. And it becomes increasingly important the more module (developers) there are.

Maybe some note regarding this issue should be added to https://docs.humhub.org/docs/develop/modules-base-class#disable

Is it worth creating a separate issue for this?

ArchBlood:
> As it turns out, any module can store class names of other modules in their own tables. As do the core modules, e.g. `comment` and `like`.
> 
> If a module is deactivated, and those records are not cleaned in the `Module::disable()` or `uninstall.php` routines, then this results in unforeseen crashes.
> 
> I guess the issue addressed by this PR should be considered significant. And it becomes increasingly important the more module (developers) there are.
> 
> Maybe some note regarding this issue should be added to https://docs.humhub.org/docs/develop/modules-base-class#disable
> 
> Is it worth creating a separate issue for this?
> 

It's worth mentioning in detail on the documentation site for new developers.

luke-:
Thanks for the PoC. Now I'm finally getting around to taking a closer look at the PR. Thank you for your patience.

I like the approach with the virtual attributes since it provides very high compatibility, but it also adds some "magic" and  a lot of complexity (e.g. wrapper of the Yii Model core methods like `getAttribute`, `__set`, ...) .

It would be interesting to check what effects there would be if we completely do without such Virutal Attributes or integrate them only temporarily.

Another interesting refactoring could be to create another table `polymorphic_relation` with attributes (id, class_map_id, primary_key) in addition to the `class_map` table. This could be used to bundle logic even further. 

Another question, I didnt found the migration for the ClassMap table, but do we really need to save the `source_file` here? Is this an absolute path?

martin-rueegg:
Thanks @luke- for looking into this and your appreciative feedback!

> I like the approach with the virtual attributes since it provides very high compatibility, but it also adds some "magic" and a lot of complexity (e.g. wrapper of the Yii Model core methods like `getAttribute`, `__set`, ...) .
> 
> It would be interesting to check what effects there would be if we completely do without such Virutal Attributes or integrate them only temporarily.

A more Yii-like approach might be to use a behavior. But I've not really looked into those much to see what you can and can't do. 
Also, a related record to the class map table might be a solution. I have created a solution that does eager loading of a related record right from the beginning which reduces SQL load significantly as there is one single SQL for both the main record and the child record. Works only for 1:1-relations though. But that would be the case here.

> Another interesting refactoring could be to create another table `polymorphic_relation` with attributes (id, class_map_id, primary_key) in addition to the `class_map` table. This could be used to bundle logic even further.

As a matter of fact I've been thinking about this today as well. In fact, I'd love to have a global GUID table with the attributes:

- gid (Global ID), unsigned int, autointeger, primary key
- guid (Global Unique ID), varchar, unique index
- class_map_id
- object_id, nullable signed int

The idea behind this is to be able to find a record using the GID or GUID which exists only once (GID per db, GUID possibly globally).

Existing tables would add a gid column, which would be an FK to the guid table. An existing guid field could be replaced in a similar approach to the class_map_id.

New tables would no longer implement an id field, but the gid field instead. 

Referencing tables would no longer build an FK to the id field, but the gid field, or directly to the gid of the guid table. Particulary, if it does implement a class_map_id/object_id combination which could then be replaced with one single field and as such add fk-checking/updating/deleting on db-level.

> Another question, I didnt found the migration for the ClassMap table, but do we really need to save the `source_file` here? Is this an absolute path?

Blimey, you're absolutely right! Going to add it! :-D

martin-rueegg:
> Another question, I didnt found the migration for the ClassMap table, but do we really need to save the `source_file` here? Is this an absolute path?

I've now changed the path to be relative, rather than absolute. But other than storing for future reference it's not used in any way in the current code. I forgot the rational behind storing it. I guess just my experience as a sysadmin ... :-)

martin-rueegg:
> I like the approach with the virtual attributes since it provides very high compatibility, but it also adds some "magic" and a lot of complexity (e.g. wrapper of the Yii Model core methods like `getAttribute`, `__set`, ...) .
> 
> It would be interesting to check what effects there would be if we completely do without such Virutal Attributes or integrate them only temporarily.

The "virtual properties", btw, are only used for backwards compatibility. New code should use the class_map_id, rather than `object_model` or `class`. As such we could issue a deprecation warning.

Talking about deprecation: I wanted to suggest to add a deprecation table, that is only used in debug mode or if activated by the configuration. It would store deprecation messages only once. It could help developers find deprecations. Nor sure this is a good idea. But if we do it, then the use of these virtual properties could issue such a warning.

Also, I don't really see an issue with overriding `getAttribute` or `__set` or alike. However, to make the code a bit more understandable, we could do something like [Improve readability of overriding virtual attribute methods ](https://github.com/humhub/humhub/pull/6217/commits/52c44273cf761826ee654dc7df30bef7222ca50c).

luke-:
## Global ID

The idea with a Global ID (name to be defined) system is very very interesting. 

Every record (e.g. Post,  User,  Space, ...) that should get such a GID must e.g. extend from "GlobalActiveRecord" (or also implement Trait, Behavior, ...) and must add an additional column called "gid" (incl. FK relation).  Then the "GlobalActiveRecord" could then automatically creates an entry in the new "Global ID" table.

Then we could e.g. replace:
- `content.object_model` & `content.object_id` ==>  `content.record_gid`.
- `notification.source_class` & `notification.source_pk` ==> `notification.source_gid`. 
- ... and in other tables ('file', 'activity', 'like', 'comment', 'user_follow', ...)

We should check if we make this migrations in steps, as it requires a bunch of refactoring & testing.

Table columns of `global_id` (name to be defined) could have following fields
- id                                       (gid in referencing tables)
- guid
- class_map_id
- ref_pk                              (primary key of referecing model/class)
- url_slug                           (future usage, to implement permalinks...)
- cached_class_name     (maybe we could also store the class name here to avoid extra lookup as some sort of  "caching")
- cached_is_paused        (maybe we could store a `is_paused` value here as well, to avoid second join)

## Class Map

After the "Global ID" concept is already used in many places (e.g. content), the class mapping is only necessary in a few situations directly: 

- Notification 
- Activity
- Global ID  


---

Maybe it would make sense to combine such a key change with something like a "Global ID".

martin-rueegg:
> ## Global ID
> 
> The idea with a Global ID (name to be defined) system is very very interesting.

Isn't it?! :-)
 
> Every record (e.g. Post, User, Space, ...) that should get such a GID must e.g. extend from "GlobalActiveRecord" (or also implement Trait, Behavior, ...) and must add an additional column called "gid" (incl. FK relation). Then the "GlobalActiveRecord" could then automatically creates an entry in the new "Global ID" table.

Exactly! It would also help with parameters when getting content via pjax. It's enough to send a `guid`, or an array thereof, and it will be clear what record to load and how to do it. So if, for example, you are requesting the `guid` of an attached file or the owning record, it does not matter. The code can find the `guid` and determine if it's the `file` and then load the owner, or find the owner, and then load the `file`.


> Then we could e.g. replace:
> 
>     * `content.object_model` & `content.object_id` ==>  `content.record_gid`.
> 
>     * `notification.source_class` & `notification.source_pk` ==> `notification.source_gid`.
> 
>     * ... and in other tables ('file', 'activity', 'like', 'comment', 'user_follow', ...)

Absolutely. Would make things much easier and reference can be ensured with FK. Orphans should no longer happen.

> We should check if we make this migrations in steps, as it requires a bunch of refactoring & testing.

Yes, I suggest to do that. Maybe with the tables already involved?


> Table columns of `global_id` (name to be defined) could have following fields
> 
>     * id                                       (gid in referencing tables)
>     * guid
>     * class_map_id
>     * ref_pk                              (primary key of referecing model/class)

I don't think this should be required, as it is a must that the referenced model does have a gid column with an FK to the `object.id` (could `object` work as a table name?). Hence, the ref_pk would simply create redundancy with no additional benefit.

>     * url_slug                           (future usage, to implement permalinks...)
👍

>     * cached_class_name     (maybe we could also store the class name here to avoid extra lookup as some sort of  "caching")
>     * cached_is_paused        (maybe we could store a `is_paused` value here as well, to avoid second join)

I would strongly suggest to refrain from this. Yes, they could be implemented as FK's with ON UPDATE CASCADE, but a) this cannot be relied on, and b) the tables are so small, that a join is very cheap.

Otherwise, a view could be an alternative. But Yii does not really use that and updatable views are not necessarily supported by all DBMS


> ## Class Map
> 
> After the "Global ID" concept is already used in many places (e.g. content), the class mapping is only necessary in a few situations directly:
> 
>     * Notification
>     * Activity
>     * Global ID
👍 


> Maybe it would make sense to combine such a key change with something like a "Global ID".
👍

martin-rueegg:
Just found, that `\yii\db\ActiveRecord::deleteAll()` is not so easy hackable ...

Hence, all these calles MUST be updated, e.g. https://github.com/humhub/humhub/pull/6217/commits/0f3be617f1877dc7492124a3c99eb64d492590ac#diff-2ce54e20db1929cf8cb630bfda8d6dc150218d04360c117770c76451061bb15fR175

luke-:
> Just found, that `\yii\db\ActiveRecord::deleteAll()` is not so easy hackable ...
> 
> Hence, all these calles MUST be updated, e.g. [0f3be61#diff-2ce54e20db1929cf8cb630bfda8d6dc150218d04360c117770c76451061bb15fR175](https://github.com/humhub/humhub/commit/0f3be617f1877dc7492124a3c99eb64d492590ac#diff-2ce54e20db1929cf8cb630bfda8d6dc150218d04360c117770c76451061bb15fR175)

We should use `deleteAll()` quite rarely anyway, since no events or model methods are triggered here. 
So should not be a big deal. But if there are potentially large amounts of records, we should rely on background jobs.

luke-:
> > Every record (e.g. Post, User, Space, ...) that should get such a GID must e.g. extend from "GlobalActiveRecord" (or also implement Trait, Behavior, ...) and must add an additional column called "gid" (incl. FK relation). Then the "GlobalActiveRecord" could then automatically creates an entry in the new "Global ID" table.
> 
> Exactly! It would also help with parameters when getting content via pjax. It's enough to send a `guid`, or an array thereof, and it will be clear what record to load and how to do it. So if, for example, you are requesting the `guid` of an attached file or the owning record, it does not matter. The code can find the `guid` and determine if it's the `file` and then load the owner, or find the owner, and then load the `file`.

Yes, that would provide interesting options. But such changes would be then something for later development phases :-)

> > Then we could e.g. replace:
> > ```
> > * `content.object_model` & `content.object_id` ==>  `content.record_gid`.
> > 
> > * `notification.source_class` & `notification.source_pk` ==> `notification.source_gid`.
> > 
> > * ... and in other tables ('file', 'activity', 'like', 'comment', 'user_follow', ...)
> > ```
> 
> Absolutely. Would make things much easier and reference can be ensured with FK. Orphans should no longer happen.

:+1: 

> > We should check if we make this migrations in steps, as it requires a bunch of refactoring & testing.
> 
> Yes, I suggest to do that. Maybe with the tables already involved?

With the tables from this PR? I don't know if it wouldn't be better to start even lower, e.g. first only with notifications, activities or content...

> 
> > Table columns of `global_id` (name to be defined) could have following fields
> > ```
> > * id                                       (gid in referencing tables)
> > * guid
> > * class_map_id
> > * ref_pk                              (primary key of referecing model/class)
> > ```
> 
> I don't think this should be required, as it is a must that the referenced model does have a gid column with an FK to the `object.id` (could `object` work as a table name?). Hence, the ref_pk would simply create redundancy with no additional benefit.

Thats right, we could probably really go without the `ref_pk`.   

> 
> > ```
> > * cached_class_name     (maybe we could also store the class name here to avoid extra lookup as some sort of  "caching")
> > * cached_is_paused        (maybe we could store a `is_paused` value here as well, to avoid second join)
> > ```
> 
> I would strongly suggest to refrain from this. Yes, they could be implemented as FK's with ON UPDATE CASCADE, but a) this cannot be relied on, and b) the tables are so small, that a join is very cheap.
> 
> Otherwise, a view could be an alternative. But Yii does not really use that and updatable views are not necessarily supported by all DBMS
> 

Ok, I know that's not the cleanest way. Lets start without any  `cache_` hacks :-)

PR #6224, opened 2023-04-11
Fix #6220: User Soft Delete doesn't remove third party auth references

### LINKED ISSUE #6220: User Soft Delete doesn't remove third party auth references

- Create a user using OAuth2 or other
- Soft delete this user
- Try to login using OAuth2 again

- User logged in

### What do you get instead?

- Error "Invalid user status!"

### Solution

We should ensure `user_auth.user_id` and `user. authclient_id` is removed when user is soft deleted.

yuriimaz:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**


protected/humhub/modules/user/controllers/AuthController.php
```diff
@@ -211,6 +211,36 @@ public function onAuthSuccess(BaseClient $authClient)
...
+    /**
+     * Do log in user
+     *
+     * @param $user
+     * @param $authClient
+     * @param $redirectUrl
+     * @param $isOAuth
+     * @return array
+     */
+    protected function doLogin($user, $authClient, $redirectUrl, $isOAuth = false)
```
luke-, line 223: Make we could mark the visibility of this method to `private`? 

Also the name of the param `$isOAuth` is a bit misleading. If I understand correctly, this is about all non "local" auth methods? (e.g. LDAP, OAuth2, SAML, ...)

protected/humhub/modules/user/services/AuthClientService.php
```diff
@@ -64,10 +64,11 @@ public function getUser(): ?User
...
      * This method will be called after login or by cron sync.
      *
-     * @param User $user
-     * @return boolean succeed
+     * @param User|null $user
+     * @param bool $isOAuth
+     * @return bool succeed
      */
-    public function updateUser(User $user = null): bool
+    public function updateUser(User $user = null, bool $isOAuth = false): bool
```
luke-, line 71: Reading the `isOAuth` param, please see the comment above.

luke- (changes requested):
Thanks for the PR!

But can it be that with this it is no longer possible to disable e.g. an OAuth2 user. It would be important to be able to disable e.g. unwanted users for a time.

luke-:
@yuriimaz Hmm strange, seems there is already:
https://github.com/humhub/humhub/blob/master/protected/humhub/modules/user/models/User.php#L521
https://github.com/humhub/humhub/blob/master/protected/humhub/modules/user/models/User.php#L515

protected/humhub/modules/user/controllers/AuthController.php
yuriimaz: Yes, I will think then how to do it better.

protected/humhub/modules/user/services/AuthClientService.php
yuriimaz: Ok

yuriimaz:
No, not really, in this case the user's status plays a role - when we delete user, user status is mistakenly set to disabled, but it should have been change to soft deleted. Check out my commit with the changes.

protected/humhub/modules/user/models/forms/AccountDelete.php
```diff
@@ -68,12 +68,13 @@ public function performDelete()
             return false;
         }
 
-        $this->user->status = User::STATUS_DISABLED;
+        $this->user->status = User::STATUS_SOFT_DELETED;
         $this->user->save();
 
+        Auth::deleteAll(['user_id' => $this->user->id]);
```
luke-, line 74: Do we really need this? If it is deleted anyway, during the `softDelete()` in ActiveJob?

protected/humhub/modules/user/services/AuthClientService.php
```diff
@@ -105,6 +106,18 @@ public function updateUser(User $user = null): bool
             if (count($user->profile->getDirtyAttributes()) !== 0 && !$user->profile->save()) {
                 Yii::warning('Could not update user profile (' . $user->id . '). Error: '
                     . VarDumper::dumpAsString($user->profile->getErrors()), 'user');
+
+                return false;
+            }
+        } elseif ($isNoLocalAuth && $user->status === User::STATUS_SOFT_DELETED) {
+            // Change status from soft deleted to enabled
+            // if user logged in with no local auth method
+            $user->setAttribute('status', User::STATUS_ENABLED);
```
luke-, line 115: It should not be possible to recover "Soft Deleted" user accounts. Instead, a new account should be created for the user.

Presumably the cause of the original bug report was that a "soft deleted" user tried to log in before the ActiveJob run. I.e. the current behavior should be correct, sorry for the confusion.

protected/humhub/modules/user/controllers/AuthController.php
```diff
@@ -211,6 +211,36 @@ public function onAuthSuccess(BaseClient $authClient)
...
+    /**
+     * Do log in user
+     *
+     * @param User $user
+     * @param BaseClient $authClient
+     * @param array $redirectUrl
+     * @param bool $isNoLocalAuth
+     * @return array
+     */
+    protected function doLogin($user, $authClient, $redirectUrl, $isNoLocalAuth = false)
```
luke-, line 223: We could mark this method as private?

luke- (changes requested):

protected/humhub/modules/user/controllers/AuthController.php
yuriimaz: Yes

protected/humhub/modules/user/services/AuthClientService.php
yuriimaz: I removed it

protected/humhub/modules/user/controllers/AuthController.php
```diff
@@ -223,20 +252,11 @@ protected function login($user, $authClient)
...
-            }
-            (new AuthClientService($authClient))->updateUser($user);
 
-            if ($success = Yii::$app->user->login($user, $duration)) {
-                Yii::$app->user->setCurrentAuthClient($authClient);
-                $redirectUrl = Yii::$app->user->returnUrl;
-            }
+        if ($user->status == User::STATUS_ENABLED) {
+            [$success, $redirectUrl] = $this->doLogin($user, $authClient, $redirectUrl);
+        } elseif ($user->status == User::STATUS_SOFT_DELETED && $authClient->getId() !== 'local') {
```
luke-, line 258: @yuriimaz Since it is not possible for SoftDeleted users to be reactivated, we can remove this as well, can't we?

Anyway I would keep the  `doLogin` method, because this way the code is a bit cleaner.

luke- (changes requested):

protected/humhub/modules/user/controllers/AuthController.php
yuriimaz: Ok, removed

luke- (approved):

luke- (approved):

PR #6261, opened 2023-04-26
Enh #10: ui.showMore - ReadMore "collapseAt" Default

yuriimaz:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**


luke-:
@yuriimaz Thanks for the PR. However, it is implemented a bit more extensive than planned.

A custom setting and input in the admin interface is not necessary for this.

It is sufficient to add in `module\post\Module.php` a class attribute `$collapsedPostHeight = 300`.
In the rare case that an admin wants to change this, this can then be done via the configuration file. 

Can you please reduce the PR accordingly.

Otherwise, the PR looks good, thanks!

yuriimaz:
Done

protected/humhub/modules/ui/Module.php
```diff
@@ -41,13 +40,19 @@ class Module extends \humhub\components\Module
         'stream' => 'bars'
     ];
 
+    /**
+     * @var int collapsed post block height
+     * @since 1.15
+     */
+    public int $collapsedPostHeight = 300;
```
luke-, line 47: Can you please add this to the Module class of the Post module. Since this settings is only used by the Post module.

If we would have a global default for all collapse content then we could add it here.

luke- (changes requested):
Thanks looks good. One small issue left.
Please also add this change to the changelog.

protected/humhub/modules/ui/Module.php
yuriimaz: Done

luke- (approved):

PR #6268, opened 2023-04-28
Enh setModalLoader

gevorgmansuryan:
https://github.com/humhub/news/issues/58


luke-:
@gevorgmansuryan Since the method is marked as deprecated.  Wouldn't it be better to use an alternative in the module?

PR #6270, opened 2023-04-28
Improve settings handling and add tests

### LINKED ISSUE #6266: BaseSettingsManager::deleteAll() does use prefix as wildcard
The comment to the function reads: 

``` php
/**
 * @param string|null $prefix if set only delete settings with given name prefix (e.g. theme.)
 **/
```

As such, the parameter $prefix is understood to be a prefix only. One would expect an SQL along the lines 
``` SQL
WHERE name LIKE 'prefix%'
```
(or given the dot mentioned in the comment, even `'prefix.%`). However, the code does not use it as a prefix:

https://github.com/humhub/humhub/blob/43e866772ca200d904e6c5afd7ff40cf777959a9/protected/humhub/libs/BaseSettingsManager.php#L251

The 3rd parameter to the like condition is missing, and not `%` manually added! Hence the SQL created will be 
``` SQL
WHERE name LIKE '%prefix%
```

The docs [read](https://www.yiiframework.com/doc/api/2.0/yii-db-queryinterface#where()-detail):

>    **like:** operand 1 should be a column or DB expression, and operand 2 be a string or an array representing the values that the column or DB expression should be like. For example, ['like', 'name', 'tester'] will generate name LIKE '%tester%'. (...)
>  **Sometimes, you may want to add the percentage characters to the matching value by yourself, you may supply a third operand false to do so. For example, ['like', 'name', '%tester', false] will generate name LIKE '%tester'.**

How shall we deal with this discrepancy?

1. Change the comment and the name of the parameter to reflect the actual behavior
2. Change the behavior to reflect the description in the comment, but potetially break someone's code who relied on this bug
3. Create an option to provide your own `%`, in which case the value is used as-is, otherwise make it `prefix%`

(It showed up while creating tests for #6190)

### LINKED ISSUE #6267: SettingsManager::flushContentContainer() only clears the collection in the current instance, not the underlying cache
https://github.com/humhub/humhub/blob/43e866772ca200d904e6c5afd7ff40cf777959a9/protected/humhub/components/SettingsManager.php#L56-L63

martin-rueegg:
Sequel to https://github.com/humhub/humhub/pull/6190:

- Fix #6267
- Enh #6271
- Enh #6272
- Fix #6266
- Enh #6270


luke-:
As a note, here is a failed module test against the Develop Branch. I haven't looked at it myself yet. 
https://github.com/humhub/calendar/actions/runs/4850778096/jobs/8644702182?pr=375

luke- (approved):

luke-:
@martin-rueegg Thanks for the great work. Really very clean & neat implementation!

martin-rueegg:
> As a note, here is a failed module test against the Develop Branch. I haven't looked at it myself yet. https://github.com/humhub/calendar/actions https://github.com/humhub/calendar/actions/runs/4850778096/jobs/8644702182?pr=375

Sorry, I had a look into it and realized that we also need to add tests for the `content container settings manager` in the content module. I intended to do that but then forgot.

Unfortunately, I was not able to spot the error right away as I had to dive into the Cest testing procedure and the composition of the pages and workflows.

If I manage I will look into it later today/this week.

luke-:
@martin-rueegg Thanks for the confirmation. It's good to know that the problem is related to the PR. I haven't had time to look closer at it myself, my hope was the PR fixes the calendar problem so I've already merged the PR. :-)

Would be great if you find some time for this - but it's not urgent at all... Let me know when I can help!

PR #6277, opened 2023-04-30
Enh/bring back pjax

gevorgmansuryan:
https://github.com/humhub/news/issues/58


luke-:
@gevorgmansuryan Can you please describe in a little more detail what the PR does.

e.g. What does it change, what does it make possible?

gevorgmansuryan:
@luke- it makes a possible to use full original Yii's Pjax - https://www.yiiframework.com/doc/api/2.0/yii-widgets-pjax
like this:

use humhub\widgets\Pjax;;

Pjax::begin();
echo GridView::widget([...]);
Pjax::end();

it adds exclude-from-pjax-client class to pjax container and custom pjax implementation from our js framework will just ignore this container.
It gives us possibility to use full Yii functions eg. GridView sorting, filtering, pagination, Forms ajax validations, etc...

luke-:
@yurabakhtin Can you take a look at this please? Do you see any problems here for our existing PJAX implementation?

yurabakhtin:
> @yurabakhtin Can you take a look at this please? Do you see any problems here for our existing PJAX implementation?

@luke- I think it will not create new problems, becuase as I see the previous class just was moved from `humhub\widgets\Pjax` into new `humhub\widgets\PjaxLayoutContent`, i.e. it was a simple renaming. And the new and old are used in single place.

But I am not sure that we really need the renaming in core, because new `humhub\widgets\Pjax` is used currently only in single place in the news module. Maybe only one reason is to have the same class `Pjax` name as it is in the yii core.

luke- (approved):

luke-:
@yurabakhtin Thanks for the review. I think a separate PJAX class for HumHub class is useful, especially if there are some customizations here like "Exclude from layout".

@gevorgmansuryan Thanks for the PR!

protected/humhub/widgets/Pjax.php
```diff
@@ -1,60 +1,19 @@
...
-     */
     public function init()
     {
-        $this->clientOptions['pushRedirect'] = true;
-        $this->clientOptions['replaceRedirect'] = true;
-        $this->clientOptions['cache'] = false;
-        $this->clientOptions['timeout'] = 5000;
+        $this->options = ArrayHelper::merge($this->options, [
+            'class' => ['exclude-from-pjax-client'],
+        ]);
```
martin-rueegg, line 15: Should that not be something like

```php
        $this->options['class'] = 'exclude-from-pjax-client ' . $this->options['class'] ?? '';
```

In order not to overwrite classes specified in the `Pjax::widget/begin` call?

gevorgmansuryan: @martin-rueegg used `ArrayHelper::merge`, it will merge classes from options

martin-rueegg: @gevorgmansuryan well, I guess the situation is more complex:

- `ArrayHelper::merge()` will indeed merge them if the classes are provided as an array with individual elements.
- However, if `$this->options['class']` is a string, then the the value from the second array will overwrite the string.

I do think this should be considered and merged accordingly. My solution does address the later case, but not the former.

Issue #6258, opened 2023-04-25
Mentioning Notification in Draft


luke-:
In a post draft, the Mentioning Notifications are already triggered when saving.

This should only happen after the publish.

yurabakhtin:
@luke- Fixed in PR https://github.com/humhub/humhub/pull/6288.

yurabakhtin:
@luke- I have one question here https://github.com/humhub/humhub/pull/6288#issuecomment-1539762977.

luke-:
@yurabakhtin Sorry, please see my reply

PR #6288, opened 2023-05-03
Send notification to mentioned user only after publishing of a content

yurabakhtin:

- [x] Bugfix

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
Issue: https://github.com/humhub/humhub/issues/6258


luke-:
@yurabakhtin What happens after restore deleted content? Can you please add some tests for this situations?

yurabakhtin:
@luke- 

> What happens after restore deleted content?

I think nothing, because if some notification was already sent then it is not sent twice, but if the mentioning notification was not sent then it can be sent first time only on the Publish action as I have described here https://github.com/humhub/humhub/pull/6288/files#diff-3df0e8a05df0ffddaaac0a53aed588b00cdcbf899d0c1973caa3bd5d18247082R376, or it can be done after next saving of a Post because it is called from as `MentioningExtension::onPostProcess()` <- `RichText::postProcess()` <- `Post::afterSave()`.

> Can you please add some tests for this situations?

Ok, I will add them.

luke-:
@yurabakhtin Ok thanks for clarify. What is about other notifications e.g. "Inform users about new content"  / "New content" or what happens if somebody creates a comment on a "Scheduled" content. Maybe we should disable Comments/Likes for Scheduled Content completly if there are further problems.

luke-:
> @luke-
> 
> > What is about other notifications e.g. "Inform users about new content" / "New content" or what happens if somebody creates a comment on a "Scheduled" content.
> 
> Yes, in previous version we run the processing related to new content only after update state from Draft to Published, so I have fixed that here [01b7385#diff-b3de8a7bb84547e2de7081f31795f491fb051a7df7042d02d991041ee0235a51R242](https://github.com/humhub/humhub/commit/01b7385dfaa51de8034b9a37d6a777600d94ab1d#diff-b3de8a7bb84547e2de7081f31795f491fb051a7df7042d02d991041ee0235a51R242) in order to do this on each update to Publish state from any other state.
> Also on creating of new content we have the option "Notify members", and I see it can work only right after submit the form, because the users are stored only in the var `Content::$notifyUsersOfNewContent` which is not stored in DB, so on publish a Draft or a Scheduled content we cannot know the users list and they cannot be notified in such cases. I.e. we should either store the notified users per Content record, or disable this option for not Published content.

If we can implement this  in the UI, I would prefer to disable the Notify User option for now, on drafts/scheduling.
 
> I think we should implement new method like `Content->hasState(Content::STATE_PUBLISHED)` because I have detected an issue: after save a Draft Content the menu item "Publish draft" is hidden. Fixed here [01b7385#diff-2c7f5125956ea366a36488ba4ce215cb860de73c049764d95def6ef8cfafd9b0R23](https://github.com/humhub/humhub/commit/01b7385dfaa51de8034b9a37d6a777600d94ab1d#diff-2c7f5125956ea366a36488ba4ce215cb860de73c049764d95def6ef8cfafd9b0R23), because right after save we have a string value like `'10'` in the field `state`, but after get from DB we have there an integer value `10`. I suggest to replace all code where we have now `$content->state === Content::STATE_*` with `$content->hasState(Content::STATE_*)`. Inside the new method we could compare with `(int) $content->state === Content::STATE_*`. Should I do this fix right in this PR or in another?

Since we are currently externalizing more logic into services. Does it make sense to implement a `ContentStateService` that implements logic like state changes etc. cleanly?

You can use the same PR for it.

> 
> > Maybe we should disable Comments/Likes for Scheduled Content completly if there are further problems.
> 
> Yes, I think it make sense to disable all addons(Comments, Likes and others) for not published Content, because it seems such Content in almost all cases are visible only for owner.

Ok

luke-:
> @luke-
> 
> > Since we are currently externalizing more logic into services. Does it make sense to implement a `ContentStateService` that implements logic like state changes etc. cleanly?
> > You can use the same PR for it.
> 
> I have created new `ContentStateService` in the commit [700924a](https://github.com/humhub/humhub/commit/700924ad2ad45ab006181475cea956acb79aaf6e).

Looks good thanks! We should also consider that Modules adding own states.
 
> 
> Please note I have moved the `Content->setState()` into the new Service, and I deprecated the method, because it is used in external modules. Should I update the modules in order to delete the method `Content->setState()` completely?

Please create PRs for the modules. And we'll delete this method in v1.15 (`develop`).

> 
> Currently we have many uses like `$content->getStateService()->is(Content::STATE_PUBLISHED)`, I suggest to add new menthods like `content->getStateService()->isPublished()`, `content->getStateService()->isDeleted()` and etc. Do you agree this?
> 

Sounds good. 


> Also we could implement new methods like `content->getStateService()->publish()` or `content->getStateService()->delete()` instead of `content->getStateService()->setState(Content::STATE_PUBLISHED)`. I think code will be more readable, but please correct me if we should not make the Service class bigger.
> 

No problem, the service class can  become larger. 

> Maybe we also should move the contants `STATE_PUBLISHED` from `Content` model into the Service, but it would be good to do this only after creating the alias methods like `isPublished()`, `publish()` and etc. in order to don't request in the code `ContentStateService::STATE_PUBLISHED`.

I would keep the constants in the model. Since they are used there...

yurabakhtin:
@luke- Commit https://github.com/humhub/humhub/pull/6288/commits/d48e91f1fc96e42f29141f59a2a91911b765073a for new methods like `$content->getStateService()->isPublished()`, `$content->getStateService()->publish()` and etc.

> I would keep the constants in the model. Since they are used there...

Ok.

---

> Please create PRs for the modules. And we'll delete this method in v1.15 (develop).

- https://github.com/humhub/calendar/pull/378
- https://github.com/humhub/cfiles/pull/178
- https://github.com/humhub/news/pull/62
- https://github.com/humhub/recycle-bin/pull/9
- https://github.com/humhub/wiki/pull/263

yurabakhtin:
@luke- 

> We should also consider that Modules adding own states.

Currently we have the method `ContentStateService->getAllowedStates()` - https://github.com/humhub/humhub/pull/6288/files#diff-c23e19c0f8c902738de418f315684f7e731e079ea5144627e8e017ee76db89a9R26-R34.

1. We could implement there an event trigger to append new states from module side.
2. This class can be extended from module side.
3. Maybe we should move the method `ContentStateService->getAllowedStates()` into the class `ContentActiveRecord` and name it as `getAllowedContentStates`, so it can be overridden from module model object easier.

yurabakhtin:
@luke- 

> > > Maybe we should disable Comments/Likes for Scheduled Content completly if there are further problems.
> > 
> > 
> > Yes, I think it make sense to disable all addons(Comments, Likes and others) for not published Content, because it seems such Content in almost all cases are visible only for owner.
> 
> Ok

Commit https://github.com/humhub/humhub/pull/6288/commits/e4ee70384331f5658d49e9fbbc46fb54f61b30ce - Comments and Likes are disabled for not published content.

luke-:
> @luke-
> 
> > We should also consider that Modules adding own states.
> 
> Currently we have the method `ContentStateService->getAllowedStates()` - https://github.com/humhub/humhub/pull/6288/files#diff-c23e19c0f8c902738de418f315684f7e731e079ea5144627e8e017ee76db89a9R26-R34.
> 
> 1. We could implement there an event trigger to append new states from module side.

That sounds like the most flexible solution.

I would probably trigger the events from `ContentStateService`.
Here ywe could either add new events at `init` or possibly manipulate them at `GetAllowedStates`.


> 2. This class can be extended from module side.
> 3. Maybe we should move the method `ContentStateService->getAllowedStates()` into the class `ContentActiveRecord` and name it as `getAllowedContentStates`, so it can be overridden from module model object easier.

 Would only work for custom `ContentActiveRecords` but not for all, right? Maybe we could also support this.

luke- (approved):

Issue #6344, opened 2023-05-22
Poll creation doubles the answers


Reuns:
### What steps will reproduce the problem?
Create a new Poll with 3 answer (actually it doesn't matter how many)
### What is the expected result?
Have a poll with 3 possible answer
### What do you get instead?
Have a poll with 6 possible answer
### Additional info
Issue on Mobile and Desktop
https://github.com/humhub/humhub/assets/19529531/8b5f5271-9885-4c67-9447-76256f583111

| Q                | A
| ---------------- | ---
| HumHub version   | 1.14.1
| Module version      |  Polls 1.3.0
| PHP version      |  8.1
| Operating system |

> Note: Also provide the full error stacks from `Administration --> Information --> Logging` and errors from your browser console if possible.
> Note: Please use the [community](https://community.humhub.com) site for any questions on customizing HumHub.

ArchBlood:
@luke- I can confirm this is an issue with HumHub v1.14.2 and Poll v1.3.0

### Space Polls
![Screenshot_1](https://github.com/humhub/humhub/assets/35392110/353d17d5-68af-470b-8095-c3d15019d8cd)

### User Polls
![Screenshot_3](https://github.com/humhub/humhub/assets/35392110/678e289b-07f0-4428-a842-f9d6bbc1aaf1)

I also noticed that when the Poll is created the title field isn't cleared after saved?

yurabakhtin:
@Reuns @ArchBlood Thanks for the report!

@luke- Fixed in PR https://github.com/humhub/polls/pull/123.
I think this bug appeared after I implemented new method `ContentActiveRecord->afterStateChange()` that is called in `Content->afterSave()`, so `ContentActiveRecord->afterSave()` is called twice, maybe in future we should implement better solution to don't call it twice.
I mean this code https://github.com/humhub/humhub/commit/6d65da3#diff-b3de8a7bb84547e2de7081f31795f491fb051a7df7042d02d991041ee0235a51R245-R247 should be called once, probably we need some flag-variable to avoid this.

luke-:
@yurabakhtin Thanks for the fix.

Can we maybe just do an `updateAttributes` in the `afterStateChange`?

yurabakhtin:
@luke- I have fixed it in core PR https://github.com/humhub/humhub/pull/6359 in order to avoid similar issue in other modules.

PR #6359, opened 2023-05-29
Avoid double call of `afterSave` on creating of a published content

yurabakhtin:

- [x] Bugfix

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
Fix issue https://github.com/humhub/humhub/issues/6344 that was created in the PR https://github.com/humhub/humhub/pull/6288 here: https://github.com/humhub/humhub/commit/6d65da3#diff-b3de8a7bb84547e2de7081f31795f491fb051a7df7042d02d991041ee0235a51R245-R247.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@ HumHub Changelog
 1.14.3 (Unreleased)
 ----------------------
 - Fix #6345: Fix updating of post on wall stream
-- Fix #6351: Error when config `defaultReloadableScripts` is not array 
+- Fix #6351: Error when config `defaultReloadableScripts` is not array
+- Fix #6359: Avoid double call of `afterSave` on creating of a published content
 
 1.14.2 (May 22, 2023)
 ----------------------
--- a/protected/humhub/modules/content/models/Content.php
+++ b/protected/humhub/modules/content/models/Content.php
@@ -241,9 +241,10 @@ class Content extends ActiveRecord implements Movable, ContentOwner, SoftDeletab
             if ($this->getStateService()->isPublished()) {
                 // Run process for new content(Send notifications) only after publishing the Content
                 $this->processNewContent();
-                // Also run process for parent object in order to send notifications like mentioning users
-                if ($model instanceof ActiveRecord) {
-                    $model->afterSave($insert, $changedAttributes);
+                // Also run process for parent object in order to send notifications like mentioning users,
+                // for case when it was not published on creating
+                if (!$insert && $model instanceof ActiveRecord) {
+                    $model->afterSave(false, []);
                 }
             }
 


luke- (approved):

funkycram:
@yurabakhtin Isn't there a better solution than running twice the afterSave() of the model? This can create bugs in some models having specific processes in afterSave().
Maybe triggering an event here (see my comment in the code above) which can be used by the notification module?

yurabakhtin:
@funkycram 

> Maybe triggering an event here (see my comment in the code above) which can be used by the notification module?

Sorry, I don't find your comment above, could you please give a link or screenshot of the comment? Maybe you should publish your comment somehow to make it visible for me.

> Isn't there a better solution than running twice the afterSave() of the model? This can create bugs in some models having specific processes in afterSave().

Yes, I can agree the solution is not ideal and may creates some bugs because we already have several related to this code.
I try to explain why I have implemented this solution and probably together we will find some better way:

- here https://github.com/humhub/humhub/blob/master/protected/humhub/modules/user/models/Mentioning.php#L139-L140 - `!$record->content->getStateService()->isPublished()` - we did to don't send a mentioning notifications if a Content is not published at the save time, for example, when Post is created as `Draft` or `Scheduled`,
- The method `Mentioning::mention()` runs inside `RichText::postProcess()` e.g. on `Post::afterSave()` and for many other external modules like here https://github.com/humhub/news/blob/main/models/News.php#L181
- I.e. we should run the mentioning notifications on change state to `Published`
- may be we should update all models like "Post", "News" and etc. to run the `RichText::postProcess()` also on `Post->afterStateChange()`, but it requires to update many modules.

@luke- It seems we have the issues here only related to the mentioning notification.
I see all mentions are stored in the DB table `user_mentioning` and we send the notification only on `Mentioning::afterSave()`.
What if we will create a new cron job hourly to send the notifications there instead of the `Mentioning::afterSave()`. We could check the `Content->isPublished()` inside the cron job, so no need to check this before creating a `Mentioning` record, and then no need to run `$model->afterSave(false, []);` inside `Content->afterSave()`.
But we need in new column `user_mentioning.is_sent = 1`. All existing records will have `1` here because they all were sent.

luke-:
@yurabakhtin 
> @luke- It seems we have the issues here only related to the mentioning notification. I see all mentions are stored in the DB table `user_mentioning` and we send the notification only on `Mentioning::afterSave()`. What if we will create a new cron job hourly to send the notifications there instead of the `Mentioning::afterSave()`. We could check the `Content->isPublished()` inside the cron job, so no need to check this before creating a `Mentioning` record, and then no need to run `$model->afterSave(false, []);` inside `Content->afterSave()`. But we need in new column `user_mentioning.is_sent = 1`. All existing records will have `1` here because they all were sent.

Isn't this related to the discussion here: https://github.com/humhub/humhub/issues/6433#issuecomment-1645523043
When we ensure via `wasPublished()` to not fire Created Activities / Mentioning Stuff twice, this should solves the problem?

yurabakhtin:
@luke- I think it is a different case, because a mentioning record should be created on each updating, i.e. we should send the mentioning notification to users on each case when content was modified, but we should send them only when content is published.
It works now like this:

- Create new Draft Post with text `@user1 some text` => don't create and don't send the mentioning notification
- Publish the Post => create and send the mentioning notification to `user1`
- Update the Published Post to text `@user1 some text @user2 another text` => create+send the mentioning notif to `user2`
- Delete the Post, update text with any new user mentioning => don't send any mentioning notifications because not Published

If we will implement a sending of the mentioning notification by new cron job it will works like this:

- Create new Draft Post with text `@user1 some text` => only create but don't send the mentioning notification
- Publish the Post => send the mentioning notification to `user1` on run new cron job hourly
- Update the Published Post to text `@user1 some text @user2 another text` => create the mentioning notif for `user2` and send by the cron job later
- Delete the Post, update text with any new user mentioning => create the mentioning notifs for all new mentioned users and send later by cron when it will be published/restored, or they will be deleted hardly if the record will be deleted hardly as well.

funkycram:
Thanks very much @yurabakhtin for your investigation and the time spent on this problem. I understand the difficulty but I'm currently missing time to go deep enough into it and give you a proper answer, sorry.

luke-:
@yurabakhtin  I like the idea with the new column `user_mentioning.is_sent` or `user_mentioning.notified`. But the hourly cron adds a too big delay.

Isn't there any event e.g. StateChange with could be used?

PR #6379, opened 2023-06-08
Normalize integer columns after load data to ActiveRecord

yurabakhtin:

- [x] Bugfix
- [x] Feature

- [x] All tests are passing
- [x] Changelog was modified

**Other information:**
Issue: https://github.com/humhub/sharebetween/issues/37#issuecomment-1573230608


protected/humhub/components/ActiveRecord.php
```diff
@@ -216,4 +217,45 @@ public function createValidators()
...
+
+    /**
+     * Check when the column value can be normalized to correct format,
+     * e.g. string formatted value '123' should be converted to integer value 123
+     *
+     * @param ColumnSchema|null $column
+     * @param mixed $value
+     * @return bool
+     */
+    public function columnValueCanBeNormalized(?ColumnSchema $column, $value): bool
```
luke-, line 249: I would prefer `private` here.

protected/humhub/components/ActiveRecord.php
```diff
@@ -216,4 +217,45 @@ public function createValidators()
...
+     */
+    public function setAttributes($values, $safeOnly = true)
+    {
+        if (is_array($values)) {
+            $schema = static::getTableSchema();
+
+            foreach ($values as $name => $value) {
+                // Make sure integers are stored with correct data type
+                $column = $schema->getColumn($name);
+                if ($this->columnValueCanBeNormalized($column, $value)) {
```
luke-, line 232: In some situations `$column` could also be `null` if it is a non-database field. Maybe we should check this.

luke- (changes requested):

protected/humhub/components/ActiveRecord.php
yurabakhtin: Yes, it can be `null` and I did this check inside the method `columnValueCanBeNormalized()` https://github.com/humhub/humhub/pull/6379/files#diff-b0143204392e87bf27a8e461e98be6aba41b271e27f0186c04afb80e6cade955R251.

protected/humhub/components/ActiveRecord.php
yurabakhtin: Ok, commit https://github.com/humhub/humhub/pull/6379/commits/8ed30f4d459405c8d4b24baa1392a24b4713f168

luke- (approved):

PR #6392, opened 2023-06-15
Performance improvements part 2

gevorgmansuryan:
https://github.com/humhub/humhub-internal/issues/24

### DIFF
--- a/protected/humhub/modules/content/models/Content.php
+++ b/protected/humhub/modules/content/models/Content.php
@@ -12,6 +12,7 @@ use humhub\components\ActiveRecord;
 use humhub\components\behaviors\GUID;
 use humhub\components\behaviors\PolymorphicRelation;
 use humhub\components\Module;
+use humhub\modules\activity\models\Activity;
 use humhub\modules\admin\permissions\ManageUsers;
 use humhub\modules\content\components\ContentActiveRecord;
 use humhub\modules\content\components\ContentContainerActiveRecord;
@@ -26,6 +27,7 @@ use humhub\modules\content\permissions\CreatePublicContent;
 use humhub\modules\content\permissions\ManageContent;
 use humhub\modules\content\services\ContentStateService;
 use humhub\modules\notification\models\Notification;
+use humhub\modules\post\models\Post;
 use humhub\modules\search\libs\SearchHelper;
 use humhub\modules\space\models\Space;
 use humhub\modules\user\components\PermissionManager;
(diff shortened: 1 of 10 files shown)


gevorgmansuryan:
Tested on MacOS, M1
Env: Docker with php7.4, mariadb (10GB RAM, 8GB PHP memory)
Data:
  content: 1000000
  post: 1000000
  activity: 1000000
  file: 10000

/dashboard/stream
  before: 1143ms / 134 queries (71 duplicated)
  after: 476ms / 47 queries (16 duplicated)

/dashboard/activity-stream
  before: 526ms / 149 queries (101 duplicated)
  after: 286ms / 52 queries (17 duplicated)

/space/stream
  before: 3743ms / 184 queries (93 duplicated)
  after: 3589ms / 70 queries (24 duplicated)

/space/stream/activity
  before: 2005ms / 134 queries (81 duplicated)
  after: 1599ms / 57 queries (17 duplicated)

protected/humhub/modules/content/models/Content.php
```diff
@@ -780,6 +782,16 @@ public function getContentContainer()
         return $this->hasOne(ContentContainer::class, ['id' => 'contentcontainer_id']);
     }
 
+    public function getPost()
```
luke-, line 785: @gevorgmansuryan I don't think it's a good idea to add relations like `getPost()` or `getActivity()`.

- There are also many other content types like `WikiPage` or `Polls` which cannot be covered by this.
- The relation matches also Activites/Wiki with the Post ID 
- E.g. a Acitivty only query, would also try to load Posts e.g. - and become more complex.

I know such heavy polymorphic relation usage is really hard to map and not very elegant.

As described in the previous PR we want to improve the base of this relations. Maybe we wait for this before adding such relations?

luke- (changes requested):

protected/humhub/modules/file/components/FileManager.php
```diff
@@ -112,10 +113,18 @@ public function findAll()
      */
     public function findStreamFiles($showInStream = true)
     {
-        if ($showInStream) {
-            return $this->find()->andWhere(['show_in_stream' => 1])->all();
+        if ($this->record->isRelationPopulated('files')) {
```
luke-, line 116: Can you please provide a unit test for this?

luke- (changes requested):
@gevorgmansuryan Thanks looks good. Please see my remark.

protected/humhub/modules/file/components/FileManager.php
gevorgmansuryan: @luke- it's just relation loading, do we really need unit test for this?

luke-: Hmm it's more about this:

```
            return ArrayHelper::getValue(
                ArrayHelper::index($this->record->files, null, 'show_in_stream'),
                +$showInStream,
                []
            );
```

gevorgmansuryan: Yeah, seems complex ) but it's just filters show_in_stream = true|false files from all files list. I decided to write this in one line instead of several foreachs. I think we don't need unit for this, it's just filtering, there no any functionality or business logic here. But if you want i'll add test for sure.

luke-: @gevorgmansuryan Thanks for the explaination. Ok, it's fine as it is.

luke- (approved):

PR #6447, opened 2023-07-19
Cannot change Homepage of space in PHP 8.x

gevorgmansuryan:
https://github.com/humhub/rest/issues/132


luke-:
@gevorgmansuryan Thanks
@yurabakhtin Can you please take a look into this?

yurabakhtin:
@gevorgmansuryan yes, I confirm your PR fixes the case when column `space.url` was `NULL` and the form "Advanced Space Settings" is updated, maybe it is actual for some old spaces without defined `url`.
But how is this fix related to the linked issue https://github.com/humhub/rest/issues/132 what I find in this PR description, maybe you wanted to link another issue here?

I have also tested the issue from https://github.com/humhub/rest/issues/132 and cannot reproduce this.
As I understand there is a problem on deleting a Space with enabled module "Polls", so on run `humhub\modules\polls\Module::disableContentContainer(Space)` the user gets the error here https://github.com/humhub/humhub/blob/master/protected/humhub/modules/content/components/ContentContainerSettingsManager.php#L80:

```php
protected function find()
{
    return parent::find()->andWhere(['contentcontainer_id' => $this->contentContainer->contentContainerRecord->id]);
}
```

I.e. I see the `Space` is passed correctly but then it is missed somewhere.
I tested this with deleting from the form "Adancaed Space Settings" `/space/manage/default/advanced` and from REST API side.

@luke- If this is a bug then should it be merged into `master`?

luke-:
@yurabakhtin Yes, `master` would be the best branch for this.

yurabakhtin:
> Yes, `master` would be the best branch for this.

@luke- I see many changed files after rebase to `master`:

rebase-master

I am not sure that all these changes will not be applied after merging, so probably we should create a new PR based on `master` v1.14.

Is it task for me or for @gevorgmansuryan?

luke-:
@gevorgmansuryan Can you please rebase/re-create this PR to master?

gevorgmansuryan:
@luke- created new branch with PR https://github.com/humhub/humhub/pull/6534

PR #6375, opened 2023-06-05
Performance Improvements

gevorgmansuryan:
https://github.com/humhub/humhub-internal/issues/24

PR #6457, opened 2023-07-24
Fix regression with Membership cache

martin-rueegg:

- Bugfix for regression introduced in [Performance Improvements](https://github.com/humhub/humhub/commit/9fabc0e2d40b580527a57eb32f3a4a29903ca3f2#diff-784788c31ac7b8ee4174e073efbfc818d7791d600260533363a35a5548670685R233-R235)  (#6375)

Requesting space membership does work, but the buttons on the space list are not updated. Therefore, the issue only happens in case the space membership is requested from the spaces list
http://humhub.local/index.php?r=space%2Fspaces

- No

- All tests are passing -> **This regression is not covered by a test**
  - **to be done:** testing the request to space membership not only from the space page, but also from the spaces list. See `RequestMembershipCest`. => #6477
- [x] Changelog was modified

### DIFF
--- a/CHANGELOG-DEV.md
+++ b/CHANGELOG-DEV.md
@@ -10,6 +10,7 @@ HumHub Changelog
 - Fix #112: Reorder Table Rows
 - Fix #6476: Fix module disabling in queue
 - Enh #6469: Implement conditions for `fixed-settings` in config
+- Fix #6457: Regression with membership cache. Also move cache to `Membership::findMembership`.
 
 1.15.0-beta.1 (July 31, 2023)
 -----------------------------
--- a/protected/humhub/modules/live/tests/codeception/unit/LegitimationTest.php
+++ b/protected/humhub/modules/live/tests/codeception/unit/LegitimationTest.php
@@ -42,7 +42,10 @@ class LegitimationTest extends HumHubDbTestCase
         $module = Yii::$app->getModule('live');
         $user = User::findOne(['id' => 1]);
         $space1 = Space::findOne(['id' => 1]);
-        $space1->addMember($user->id);
+
+        static::assertTrue($space1->addMember($user->id));
+        static::assertInstanceOf(Membership::class, Membership::findOne(['user_id' => 1, 'space_id' => 1]));
+
         $legitimations = $module->getLegitimateContentContainerIds($user, false);
 
         static::assertCount(1, $legitimations[Content::VISIBILITY_OWNER]);
--- a/protected/humhub/modules/space/behaviors/SpaceModelMembership.php
+++ b/protected/humhub/modules/space/behaviors/SpaceModelMembership.php
@@ -208,6 +208,7 @@ class SpaceModelMembership extends Behavior
         if ($membership != null) {
             $membership->group_id = Space::USERGROUP_ADMIN;
             $membership->save();
+            Yii::$app->runtimeCache->delete(__CLASS__ . '::getMembership' . $userId . '-' . $this->owner->id);
             return true;
         }
 
@@ -305,6 +306,8 @@ class SpaceModelMembership extends Behavior
 
         $membership->save();
 
+        Yii::$app->runtimeCache->delete(__CLASS__ . '::getMembership' . $userId . '-' . $this->owner->id);
+
         ApprovalRequest::instance()->from($user)->about($this->owner)->withMessage($message)->sendBulk($this->getAdminsQuery());
     }
 
@@ -375,6 +378,8 @@ class SpaceModelMembership extends Behavior
         if (!$membership->save()) {
             throw new Exception('Could not save membership!' . print_r($membership->getErrors(), 1));
         }
+
+        Yii::$app->runtimeCache->delete(__CLASS__ . '::getMembership' . $userId . '-' . $this->owner->id);
 
         if ($sendInviteNotification) {
             $this->sendInviteNotification($userId, $originatorId);
@@ -474,6 +479,8 @@ class SpaceModelMembership extends Behavior
             return false;
         }
 
+        Yii::$app->runtimeCache->delete(__CLASS__ . '::getMembership' . $userId . '-' . $this->owner->id);
+
         MemberEvent::trigger(Membership::class, Membership::EVENT_MEMBER_ADDED, new MemberEvent([
             'space' => $this->owner, 'user' => $user
         ]));
@@ -541,6 +548,8 @@ class SpaceModelMembership extends Behavior
     private function handleRemoveMembershipEvent(Membership $membership, User $user)
     {
         unset($this->_memberships[$user->id]);
+
+        Yii::$app->runtimeCache->delete(__CLASS__ . '::getMembership' . $user->id . '-' . $this->owner->id);
 
         // Get rid of old notifications
         ApprovalRequest::instance()->from($user)->about($this->owner)->delete();
--- a/protected/humhub/modules/space/behaviors/SpaceModelMembership.php
+++ b/protected/humhub/modules/space/behaviors/SpaceModelMembership.php
@@ -41,11 +41,6 @@ class SpaceModelMembership extends Behavior
     private $_spaceOwner = null;
 
     /**
-     * @var array cached membership results
-     */
-    private $_memberships = [];
-
-    /**
      * Checks if given userId is Member of this Space.
      *
      * @param integer $userId
@@ -208,7 +203,6 @@ class SpaceModelMembership extends Behavior
         if ($membership != null) {
             $membership->group_id = Space::USERGROUP_ADMIN;
             $membership->save();
-            Yii::$app->runtimeCache->delete(__CLASS__ . '::getMembership' . $userId . '-' . $this->owner->id);
             return true;
         }
 
@@ -220,23 +214,11 @@ class SpaceModelMembership extends Behavior
      *
      * If none Record is found, null is given
      *
-     * @return Membership the membership
+     * @return Membership|null the membership
      */
-    public function getMembership($userId = null)
+    public function getMembership($userId = null): ?Membership
     {
-        if ($userId instanceof User) {
-            $userId = $userId->id;
-        } elseif (!$userId || $userId == '') {
-            $userId = Yii::$app->user->id;
-        }
-
-        if (!isset($this->_memberships[$userId])) {
-            $this->_memberships[$userId] = Yii::$app->runtimeCache->getOrSet(__METHOD__ . $userId . '-' . $this->owner->id, function() use ($userId) {
-                return Membership::findOne(['user_id' => $userId, 'space_id' => $this->owner->id]);
-            });
-        }
-
-        return $this->_memberships[$userId];
+        return Membership::findMembership($this->owner->id, $userId);
     }
 
     /**
@@ -305,8 +287,6 @@ class SpaceModelMembership extends Behavior
         ]);
 
         $membership->save();
-
-        Yii::$app->runtimeCache->delete(__CLASS__ . '::getMembership' . $userId . '-' . $this->owner->id);
 
         ApprovalRequest::instance()->from($user)->about($this->owner)->withMessage($message)->sendBulk($this->getAdminsQuery());
     }
@@ -378,8 +358,6 @@ class SpaceModelMembership extends Behavior
         if (!$membership->save()) {
             throw new Exception('Could not save membership!' . print_r($membership->getErrors(), 1));
         }
-
-        Yii::$app->runtimeCache->delete(__CLASS__ . '::getMembership' . $userId . '-' . $this->owner->id);
 
         if ($sendInviteNotification) {
             $this->sendInviteNotification($userId, $originatorId);
@@ -479,8 +457,6 @@ class SpaceModelMembership extends Behavior
             return false;
         }
 
-        Yii::$app->runtimeCache->delete(__CLASS__ . '::getMembership' . $userId . '-' . $this->owner->id);
-
         MemberEvent::trigger(Membership::class, Membership::EVENT_MEMBER_ADDED, new MemberEvent([
             'space' => $this->owner, 'user' => $user
         ]));
@@ -547,9 +523,7 @@ class SpaceModelMembership extends Behavior
      */
     private function handleRemoveMembershipEvent(Membership $membership, User $user)
     {
-        unset($this->_memberships[$user->id]);
-
-        Yii::$app->runtimeCache->delete(__CLASS__ . '::getMembership' . $user->id . '-' . $this->owner->id);
+        Membership::unsetCache($this->owner->id, $user->id);
 
         // Get rid of old notifications
         ApprovalRequest::instance()->from($user)->about($this->owner)->delete();
--- a/protected/humhub/modules/space/models/Membership.php
+++ b/protected/humhub/modules/space/models/Membership.php
@@ -62,11 +62,6 @@ class Membership extends ActiveRecord
     const USER_SPACES_CACHE_KEY = 'userSpaces_';
     const USER_SPACEIDS_CACHE_KEY = 'userSpaceIds_';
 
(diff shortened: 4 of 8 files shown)


luke-:
@gevorgmansuryan Can you please take a look?

martin-rueegg:
One problem here is that most tests run without active cache.

I have now added tests for both the cached as well as the uncached calls. However, I've not set up to use the cache during the test.

martin-rueegg:
The particular situation is in `\humhub\modules\space\behaviors\SpaceModelMembership::addMember`:

In https://github.com/humhub/humhub/blob/9a28225efd9e1e6464c261d7685866e84c3200d9/protected/humhub/modules/space/behaviors/SpaceModelMembership.php#L427 which is called from here https://github.com/humhub/humhub/blob/9a28225efd9e1e6464c261d7685866e84c3200d9/protected/humhub/modules/space/controllers/MembershipController.php#L95

... the membership is retrieved (and not found) and hence a `null` value is stored in the cache.

When then later, back in https://github.com/humhub/humhub/blob/9a28225efd9e1e6464c261d7685866e84c3200d9/protected/humhub/modules/space/controllers/MembershipController.php#L97 or rather in the respective https://github.com/humhub/humhub/blob/9a28225efd9e1e6464c261d7685866e84c3200d9/protected/humhub/modules/space/controllers/MembershipController.php#L312 yet again in https://github.com/humhub/humhub/blob/master/protected/humhub/modules/space/behaviors/SpaceModelMembership.php#L63 the membership is again loaded, it gets the `null` from the cache, rather than the meanwhile created record.

martin-rueegg:
Not quite sure, why the [LegitimationTest::testSpaceMemberCanSeePrivateContent](https://github.com/humhub/humhub/actions/runs/5645932798/job/15292766597?pr=6457) failed now. I can't reproduce the error locally. Here the test succeeds. 

See if clearing the legitimate content container id cache will solve the issue. But then again, this only makes sense, if the cache would be active during the test ...

martin-rueegg:
Ok, another interesting observation:

When I run the entire test suite, the LegitimationTest fails:

    protected/humhub/tests$ php $HUMHUB_VENDOR_BIN/codecept run -v

However, when I only run the tests from the affected module, they succeed:

    protected/humhub/modules/live/tests$ php $HUMHUB_VENDOR_BIN/codecept run -v

Maybe something that should be looked at, too?

martin-rueegg:
Regarding the caching of the Membership, I doubt that there is any performance improvement in it's original implementation. In fact, I believe the following line simply create overhead and increase complexity:

https://github.com/humhub/humhub/blob/9a28225efd9e1e6464c261d7685866e84c3200d9/protected/humhub/modules/space/behaviors/SpaceModelMembership.php#L232-L236

What this code is basically doing:

1. It checks if the object exists in the array of the local member variable
2. If not, it checks, if it's in the cache (which is an array cache by default, unless it's a console process, then it's simply a dummy cache)
3. If it's not there, it looks up the database and
   1. Saves the Membership/null in the cache (if not dummy)
   2. Saves the Membership/null in the local member variable
4. Returns the Membership/null.

As the local member variable has the same live time as the runtime cache (if it's an array cache) will never be an entry in the cache, unless the space object (and hence the `SpaceModelMembership` behavior is recreated. In which case, the local member variable does not make much sense.

The approach I took was to use a static variable in the Membership class. Which definitively has the same live span as the array cache (unless I'm mistaken). In which case ...

- ... using only the static variable is more performant
- ... using only the cache is more configurable (as the runtime cache can be influenced by the config)

Either way, I'd suggest to only use either an local array _or_ the cache, but not both.

martin-rueegg:
Ok, so the reason why the test failed when running the full suite was that the array did not get cleared during the test setup, while on the other hand the database table got emptied. Hence, the Membership object was still available in the array (so it was not re-created in the db) and hence, when then getting the posts it would not be visible due to the missing Membership.

In my previous message I've laid out, why it doesn't make much sense to have _both_ the array cache _and_ the `runtimeCache` as they are basically doing the same (in this situation), the later just with additional overhead.

Still, for being consistent with the rest of the application _and_ for the benefit of being able to configure/influence the cache behaviour, I've now opted to remove the "local" array, and exclusively use the `runtimeCache` instead.

If, however, @luke- would prefer to use the array here in any case, then we would have to make sure

- the array is initialized empty when loading the class, i.e. in Membership.php
- Since in the test suite the class seem _not_ to be reloaded, it would also have to be cleared wherever the Membership table gets cleared.
- Alternatively, the array must be moved back into an object instance, as it was before, but losing the benefit of using the cache throughout the application.
-

luke-:
@martin-rueegg Thank you for your help here. I also see no reason (except maybe in the Console scenario) to keep both "caches" (`runtimeCache` and static variable). So for me, your solution with the `runtimeCache` is fine. Consistency is a good argument here.

One small point, if possible, I would prefer when the method `Membership::findMembership(Space $space, User $user): ?membership` would have strict types and does not allow `int` argument.  What do you think?

Probably it really makes sense to enable caching during tests? Should we create a new issue for this?

@gevorgmansuryan Can you please also take a look into this? Could there possibly be similar issues in your recent `runtimeCache` implementation?

martin-rueegg:
> One small point, if possible, I would prefer when the method `Membership::findMembership(Space $space, User $user): ?membership` would have strict types and does not allow `int` argument. What do you think?

Thank you for asking for my opinion.

I personally would suggest to create at least two corresponding methods for the User and Space classes (potentially with caching, too):

- `User::findUser($user): ?User`
- `User::findUserAsId($user): ?int`
- `Space::findSpace($space): ?Space`
- `Space::findSpaceAsId($space): ?int`

The `User::findUser` and `Space::findSpace` would do the real work, the other two only get the `$object->id` from the main methods, unless the result is `null`.

The code in
https://github.com/metaworx/humhub/blob/fbc093a9358bb02fe4d8dcdc1309f74396855f85/protected/humhub/modules/space/models/Membership.php#L413-L425
would then be moved into the respective methods.

This would allow to simplify all these countless occurrences of similar code to be reduced to one function call and you know exactly that you either get a User object - or a UserID respectively - or `null`, no matter what you put in (int, User, or null). In case of an invalid data type, an exception is thrown.

Happy to create a PR for this, if that would be in your interest.


> Probably it really makes sense to enable caching during tests? Should we create a new issue for this?

To be honest: this will be a bit of a task. Yes, I would strongly suggest that cache is tested. Because hardly any site runs without cache. However, it's a bit tricky, as the cache would have to be deleted from one test to the next to have a reproducible environment. But then again, the first call to a function might result in a different result than the second, due to the caching. And in many cases, things are only tested once, of course. So all tests that (indirectly) invoke a function that uses cache would need adaption and have to call things twice.

An alternative could be to create a special testing cache which, whenever a value is retrieved that has been cached, retrieve it again from the callback and compares the two results.

But yes, that would have definitively to go in a separate issue.

> @gevorgmansuryan Can you please also take a look into this? Could there possibly be similar issues in your recent `runtimeCache` implementation?

Also, I'd appreciate if this PR could be merged and a new issue for this investigation could be done. It's easier for me to get ahead with my stuff if this PR is merged as-is and I'm no longer stuck with this in my other tests ...

luke-:
> > One small point, if possible, I would prefer when the method `Membership::findMembership(Space $space, User $user): ?membership` would have strict types and does not allow `int` argument. What do you think?
> 
> Thank you for asking for my opinion.
> 
> I personally would suggest to create at least two corresponding methods for the User and Space classes (potentially with caching, too):
> 
> * `User::findUser($user): ?User`
> * `User::findUserAsId($user): ?int`
> * `Space::findSpace($space): ?Space`
> * `Space::findSpaceAsId($space): ?int`
> 
> The `User::findUser` and `Space::findSpace` would do the real work, the other two only get the `$object->id` from the main methods, unless the result is `null`.
> 
> The code in https://github.com/metaworx/humhub/blob/fbc093a9358bb02fe4d8dcdc1309f74396855f85/protected/humhub/modules/space/models/Membership.php#L413-L425 would then be moved into the respective methods.
> 
> This would allow to simplify all these countless occurrences of similar code to be reduced to one function call and you know exactly that you either get a User object - or a UserID respectively - or `null`, no matter what you put in (int, User, or null). In case of an invalid data type, an exception is thrown.
> 
> Happy to create a PR for this, if that would be in your interest.
> 

So if we could introduce two methods:

- `Membership::findMembership(Space $space, User $user): ?self`
- `Membership::findMembershipByIds(int $spaceId, int $userId): ?self`

I'm fine with this too.

But not sure if we need `User::findUser` or User::findUserAsId`. There is already a method `User::findByPk()`. Or did I missed something here?

> > Probably it really makes sense to enable caching during tests? Should we create a new issue for this?
> 
> To be honest: this will be a bit of a task. Yes, I would strongly suggest that cache is tested. Because hardly any site runs without cache. However, it's a bit tricky, as the cache would have to be deleted from one test to the next to have a reproducible environment. But then again, the first call to a function might result in a different result than the second, due to the caching. And in many cases, things are only tested once, of course. So all tests that (indirectly) invoke a function that uses cache would need adaption and have to call things twice.
> 
> An alternative could be to create a special testing cache which, whenever a value is retrieved that has been cached, retrieve it again from the callback and compares the two results.
> 
> But yes, that would have definitively to go in a separate issue.

I've created an issue here: https://github.com/humhub/humhub/issues/6462

> 
> > @gevorgmansuryan Can you please also take a look into this? Could there possibly be similar issues in your recent `runtimeCache` implementation?
> 
> Also, I'd appreciate if this PR could be merged and a new issue for this investigation could be done. It's easier for me to get ahead with my stuff if this PR is merged as-is and I'm no longer stuck with this in my other tests ...

I'm quite sure we can merge this PR. But it would be good if @gevorgmansuryan can take a look into it before.

martin-rueegg:
> > > One small point, if possible, I would prefer when the method `Membership::findMembership(Space $space, User $user): ?membership` would have strict types and does not allow `int` argument. What do you think?
> > 
> > 
> > Thank you for asking for my opinion.
> > I personally would suggest to create at least two corresponding methods for the User and Space classes (potentially with caching, too):
> > 
> > * `User::findUser($user): ?User`
> > * `User::findUserAsId($user): ?int`
> > * `Space::findSpace($space): ?Space`
> > * `Space::findSpaceAsId($space): ?int`
> > 
> > The `User::findUser` and `Space::findSpace` would do the real work, the other two only get the `$object->id` from the main methods, unless the result is `null`.
> > The code in https://github.com/metaworx/humhub/blob/fbc093a9358bb02fe4d8dcdc1309f74396855f85/protected/humhub/modules/space/models/Membership.php#L413-L425 would then be moved into the respective methods.
> > This would allow to simplify all these countless occurrences of similar code to be reduced to one function call and you know exactly that you either get a User object - or a UserID respectively - or `null`, no matter what you put in (int, User, or null). In case of an invalid data type, an exception is thrown.
> > Happy to create a PR for this, if that would be in your interest.
> 
> So if we could introduce two methods:
> 
>     * `Membership::findMembership(Space $space, User $user): ?self`
> 
>     * `Membership::findMembershipByIds(int $spaceId, int $userId): ?self`
> 
> 
> I'm fine with this too.
> 
> But not sure if we need `User::findUser` or User::findUserAsId`. There is already a method `User::findByPk()`. Or did I missed something here?

If you're referring to [User::findIdentity()](https://github.com/humhub/humhub/blob/9a28225efd9e1e6464c261d7685866e84c3200d9/protected/humhub/modules/user/models/User.php#L329), than yes, that's kind of what I meant. And it is not typed at all, neither with the input, not the output.

I would personally allow `User|int|string|null` for input. There are so many places in the code, where this could be used:

https://github.com/humhub/humhub/blob/9a28225efd9e1e6464c261d7685866e84c3200d9/protected/humhub/modules/user/models/Group.php#L273
https://github.com/humhub/humhub/blob/9a28225efd9e1e6464c261d7685866e84c3200d9/protected/humhub/modules/user/models/Group.php#L345
https://github.com/humhub/humhub/blob/9a28225efd9e1e6464c261d7685866e84c3200d9/protected/humhub/modules/user/models/Group.php#L355-L356
https://github.com/humhub/humhub/blob/9a28225efd9e1e6464c261d7685866e84c3200d9/protected/humhub/modules/user/models/Group.php#L385-L386
https://github.com/humhub/humhub/blob/9a28225efd9e1e6464c261d7685866e84c3200d9/protected/humhub/modules/user/models/User.php#L580
https://github.com/humhub/humhub/blob/9a28225efd9e1e6464c261d7685866e84c3200d9/protected/humhub/modules/user/models/UserFilter.php#L31-L35

And there are potentially many more if considering modules.

Having those functions, for the Membership class we only require one as the rest is taken care of in the User and Space classes.

What is the advantage of having a strongly typed input, if the type is verified in the function and throws an exception, as invoking the function with a wrong type would?

martin-rueegg:
> @gevorgmansuryan Can you please also take a look into this? Could there possibly be similar issues in your recent `runtimeCache` implementation?

As a matter of fact, there are.

For example here: https://github.com/humhub/humhub/blob/9a28225efd9e1e6464c261d7685866e84c3200d9/protected/humhub/modules/user/models/User.php#L331-L333

Also in the User model, there is no reset of the cache when a record is updated. This is partly solved in #6463. However, it is quite complicated, as the records can be manipulated by not only the following functions (which can be intercepted to delete the cache) but by any sql query behind the scenes. This is certainly something to keep in mind. Also, the functions marked with an asterisk below, do potentially manipulate multiple records. Since the cache stores the entries with a hashed key, there is no way to selectively flush let's say all User entries. Instead, the entire cache needs to be flushed. An alternative approach could be to create a dedicated RuntimeCache class that allows to 

- configure the storing backend (defaulting to the current setting)
- possibly allows to configure includes and excludes (to not have to entirely disable the cache in console settings)
- maybe keep track of the plain-text keys, so that flushing with a wildcard would be possible

The following model methods affect records (may be incomplete):

- `model::save()` => `model::afterSave()`
- `model::delete()` => `model::afterDelete()`
- `model::deleteAll()`*
- `model::updateAll()`*
- `model::updateAllCounters()`*

*) Functions marked with an asterisk do potentially manipulate multiple records. Hence, the entire cache needs to be flushed. See also above.

See https://github.com/humhub/humhub/pull/6463/files#diff-65dd9f5389478b955980a0410c155bd62e7cdc27f28eebc2f34d0c6bde4741ca for possible workarounds.

martin-rueegg:
There is another situation that needs consideration:

- updating a `content` record should clear the associated record cache
- updating a `contentcontainer` rechord should also clear the parent record

This would require to make the current naming convention for cache key (ie. `$model_class#$pk`/`$object_class#$object_id` mandatory in order for the detail records (`content`/`contentcontainer`) to delete their parent records.

martin-rueegg:
@gevorgmansuryan Is there any progress on reviewing this?

martin-rueegg:
@gevorgmansuryan thanks for having a look into this. By "main PR" you're referring to this one (#6457)?

Could you elaborate on why a `behavior` would suit better than the version using a `trait`? Since caching does not happen automatically, but by doing it explicitly, would it not also suit to do it by intercepting the delete and update methods, which in most cases do not even need to be touched if the trait and the enhanced RuntimeCache implementation from #6482 are used.

gevorgmansuryan:
> By "main PR" you're referring to this one (#6457)?
@martin-rueegg yes

I'll take a look to https://github.com/humhub/humhub/pull/6482

martin-rueegg:
@gevorgmansuryan thank you. Looking forward for your feedback.

@luke- shall we then merge this PR into `develop`, and from there into `statable` and `next`, so we can move forward to review the PRs that are building on top of this one?

luke- (approved):

luke-:
@martin-rueegg 

> @luke- shall we then merge this PR into `develop`, and from there into `statable` and `next`, so we can move forward to review the PRs that are building on top of this one?

Done. Thank you for the fix.

Issue #5713, opened 2022-05-20
Disabling modules should be done in a background job


funkycram:
### Is your feature request related to a problem? Please describe.

If you disable a module having many entries in the database, disabling the module with the browser can create a timeout which could corrupt the database.

### Describe the solution you'd like

Execute the module deactivation in a background job.

luke-:
https://github.com/humhub/humhub/issues/5713

funkycram:
@gevorgmansuryan is there a PR related to these commits?
`#5713` is not yet merged in https://github.com/humhub/humhub/blob/develop/CHANGELOG-DEV.md

luke-:
@funkycram FYI, I've just merged the PR into `develop`

funkycram:
Works perfectly! Thanks!

funkycram:
@luke- @gevorgmansuryan I think we need a new method in `humhub\components\Module` to allow disabling things immediately (and not with cron), such as `$this->disableTheme();`.

See https://community.humhub.com/comment/perma?id=45889

funkycram:
Thanks @yurabakhtin . Yes, it's a much better approach than adding an extra method!

luke-:
> @luke- @funkycram @gevorgmansuryan Fixed in PR #6476:
> 
> I think the line https://github.com/humhub/humhub/pull/6179/files#diff-6e6e0e5fd7c6c82bc558a3f673725b3c6286b2ff9afe548347820b1b2fe8c009R116 - `Yii::$app->moduleManager->disable($module);` was added there by mistake because it looks like previous line `$module->disable()` was splitted to:
> 
> 1. queue - `Yii::$app->moduleManager->getModule($this->moduleId)->disable()`
> 2. controller - `Yii::$app->moduleManager->disable($module)`
> 
> I don't find a reason to keep the 2nd point because it is also called inside the end of the 1st code.
> 
> I also decided to inform user with the message after deactivate a module:
> 

Probably the line was added because otherwise after deactivation, the module is still listed as activated before the background job is finished.

As an idea: In the "Admin -> Modules" list, could we show the module also as `disabled` when:

```
if (QueueHelper::isQueued(new DisableModuleJob(['moduleId' => $moduleId]))) {
```

yurabakhtin:
> As an idea: In the "Admin -> Modules" list, could we show the module also as `disabled` when:

[...]

PR #6476, opened 2023-08-03
Fix module disabling in queue

yurabakhtin:

- [x] Bugfix

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
Issue: https://github.com/humhub/humhub/issues/5713#issuecomment-1663464004


luke- (approved):

ArchBlood:
@luke- looking at this closer, it doesn't seem like the job clears leftover cache for a disabled module like it did before switching to background jobs, is that intended or is this now the module developer's responsibility to implement into their `disable()` function now? 🤔

luke-:
@ArchBlood In this case, the cache in the PHP APC may not be deleted correctly.

ArchBlood:
> @ArchBlood In this case, the cache in the PHP APC may not be deleted correctly.

It's rather weird to be honest, using the flush setting does work as intended, it's just for background jobs it doesn't work.

PR #6478, opened 2023-08-03
Add pseudo test class to allow population of DB with standard test data

martin-rueegg:

- Other, please describe:

Create an easy way to populate the database with the standard test data set, without clearing it afterwards.

Normal use will not do anything other than succeed and hence not consume any time
```
php $HUMHUB_VENDOR_BIN/codecept run unit LoadDbTest
```

However, if run with the environment variable HUMHUB_DB_INITIALIZE set, then the default fixtures will be loaded and not unloaded after the test.
```
HUMHUB_DB_INITIALIZE= php $HUMHUB_VENDOR_BIN/codecept run unit LoadDbTest
```

This class could also be extended and then the data only loaded for certain tests within the class. But for now it helps to test the codebase from the web-interface easily.

- No

- [x] All tests are passing
- [x] Changelog was modified

### DIFF
--- a/CHANGELOG-DEV.md
+++ b/CHANGELOG-DEV.md
@@ -3,6 +3,7 @@ HumHub Changelog
 
 1.15.0-beta.2 (Unreleased)
 --------------------------
+- Enh #6478: Add pseudo test class to allow population of DB with standard test data
 - Enh #6480: Convert assert* and db* methods to static, in line with general usage pattern
 - Enh #6505: Introduce Application interface; now also fire the `onInit` event when the web application has initialized
 - Fix #6502: Link notification for pending space approval to manage page
--- a/protected/humhub/tests/codeception/unit/LoadDbTest.php
+++ b/protected/humhub/tests/codeception/unit/LoadDbTest.php
@@ -1,0 +1,86 @@
+<?php
+
+/*
+ * @link      https://www.humhub.org/
+ * @copyright Copyright (c) 2023 HumHub GmbH & Co. KG
+ * @license   https://www.humhub.com/licences
+ */
+
+namespace humhub\tests\codeception\unit;
+
+use Codeception\Exception\InjectionException;
+use Codeception\Exception\ModuleException;
+use Codeception\Lib\Console\Output;
+use Codeception\Lib\ModuleContainer;
+use Codeception\Module\Yii2;
+use Codeception\Test\Metadata;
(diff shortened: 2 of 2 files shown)


luke-:
Hmm, sorry, I don't understand exactly what the class is supposed to be used for.

Ideally there would be a reference implementation in HumHub Core or Official Modules so it easier to understand how it works.

martin-rueegg:
Hil @luke- 

This _is_ actually the use case.

I've been intensively running tests recently in regards to the caching issues and other PRs I'm working on.

However, the error messages from the tests, particularly from the Cests are not very helpful to find the error in many cases. Unit and functional tests can usually be debugged from the command-line using xdebug. However, for Cests that is _not_ possible.

But in order to reproduce the exact steps of the Cest, the test data set is required in the database. This, however, will only be available once you run the test again and it fails. If your failing sub-test is followed by a succeeding sub-test, you need to specify the subtest to make it the last test failing and preserving the data.

Hence, this test doesn't test anything. But it uses the built-in mechanism to seed the db with the test data. So if you run it with the given environment variable as indicated in the introduction of this PR, your DB will be seeded with the exact test data and you're ready to go and test the thing in your browser and with your full debugging stack (front-end and back-end).

If you do not add the env variable, the test doesn't test anything but does succeed in order to not make the test suite fail.

I hope I was able to better explain the use case now. It would be helpful to have this in the develop code-base just for test-debugging purposes. It has no side effect and doesn't affect production code in any way.

luke-:
@martin-rueegg Thank you for the explanation.

If I understand it correctly, this unit tests only loads the database with fixtures so that you can manually debug based on this data afterwards?

We can merge this then with pleasure, however, I would like to add a short explanation with short usage hint in the class header e.g.

```php
/**
 * Test Debugging DbLoad Helper
 * Use this test to fill up the database with fixture test data to manually debug
 *
 * Example usage:
 * ```
 * HUMHUB_DB_INITIALIZE= php $HUMHUB_VENDOR_BIN/codecept run unit LoadDbTest
 * ```
 * 
 * @since 1.15
 */
```

martin-rueegg:
@luke- yes, that is exactly the purpose! I find it really handy do find errors that have been uncovered by a Cest.

Thank you for your suggestion. Good idea!

Have rebased the PR and added the class header.

luke- (approved):

PR #6480, opened 2023-08-03
Convert assert* and db* methods to `static`, in line with general usage pattern

martin-rueegg:

- Refactor

- No

- [x] Changelog was modified

**Other information:**

### DIFF
--- a/CHANGELOG-DEV.md
+++ b/CHANGELOG-DEV.md
@@ -3,6 +3,7 @@ HumHub Changelog
 
 1.15.0-beta.2 (Unreleased)
 --------------------------
+- Enh #6480: Convert assert* and db* methods to static, in line with general usage pattern
 - Enh #6505: Introduce Application interface; now also fire the `onInit` event when the web application has initialized
 - Fix #6502: Link notification for pending space approval to manage page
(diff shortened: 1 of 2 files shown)


luke-:
@martin-rueegg Thanks for the PR. 

A question, since the methods are used in some modules: Do you know if it is safe when we call these new static methods, continue using the instance? Or is there a warning/deprecation here?

Actually, it would be best to refactor the modules as well...

martin-rueegg:
AFAIK, @luke-, the call to a static method works also in the instance notation (i.e. `$this->method()`, rather than `static::method()` or `$object::method()`). I was only notified in PhpStorm to use static call notation in tests, but they succeeded anyway.

But yes, updating the modules (and the tests themselves) might be a good (although not urgent) approach.

martin-rueegg:
@luke- 

> Actually, it would be best to refactor the modules as well...

Shall we create an issue for the above?

... and merge this PR?

luke- (approved):

luke-:
@martin-rueegg Thanks, I've just merged this PR. I'll create an internal tasks for refactoring the modules, but since a static method can also called by instance, there is no problem here.

martin-rueegg:
Awesome. Thank you!

Issue #6486, opened 2023-08-07
Add possibility to always deactivated a module in spaces or users


funkycram:
### Is your feature request related to a problem? Please describe.

Example with the wiki module.
You want to allow activating the module in spaces but not for users.

### Describe the solution you'd like

![image](https://github.com/humhub/humhub/assets/23310825/4b48cdf8-264b-4472-bdcd-3023ae94abfb)

funkycram:
@luke- if you agree with this idea, I can submit a PR. Humhub 1.15 or 1.16?

luke-:
Yes, an "Not available" option would be good.

But I am not sure if the "Defaults" modal is the right place here.

There is also a similar feature here: https://github.com/humhub/humhub/pull/3415
However, we need to revise this PR again.

funkycram:
Yes, I agree the "set as default" is not appropriate for making the module not available on a container types. 

However, it might add complexity to add another page for that.

Perhaps we could rename the "Set as default module" modal box title with something like "Activation in spaces and for user" and then rename the "Deactivated" and "Activated" options with "Deactivated by default" and "Activated by default"?

luke-:
For now, I think it's fine to add the "Not available" option to the "Default" modal.

If we add a way to restrict modules by container (e.g. Module Wiki for Space A only), we can move this new option to this dialog.

funkycram:
OK, fine. So I'll add a check of what's chosen into [ContentContainerModule::hasContentContainerType()](https://github.com/humhub/humhub/blob/develop/protected/humhub/modules/content/components/ContentContainerModule.php#L75).

How should I deal with already activated modules in containers?

Should I:
- just hide them and keep the data if available again later
- or should I disable them and execute the `ContentContainerModule::disableContentContainer()` in which case I need to add warning when saving the form if "Not available" is selected?

funkycram:
Option 1 (just hide) is a little tricky to do because I cannot filter easily content related to "not available" modules in `ContentContainerStreamFilter::apply()`, unless I add a column in the `content` table to indicate the module ID of the content.

If I had this module ID, I could filter this way:
```php
        $enabledModuleIds = array_map(static function ($contentContainerModule) {
            return $contentContainerModule;
        }, $this->container->moduleManager->getAvailable());
        $this->query->andWhere(['content.module_id' => $enabledModuleIds]);
```

@luke- What do you think about this idea? Or should I do option 2 (disable the module and delete data)?

luke-:
[...]

PR #6365, opened 2023-05-31
Enh #6335: Pending approvals: possibility to send a message

marc-farre:
Enh #6335: Pending approvals: possibility to send a message
https://github.com/humhub/humhub/issues/6335

**What kind of change does this PR introduce?** (check at least one)

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

**Does this PR introduce a breaking change?** (check one)

- [ ] Yes
- [x] No

If yes, please describe the impact and migration path for existing applications:

**The PR fulfills these requirements:**

- [x] It's submitted to the `develop` branch, _not_ the `master` branch if no hotfix
- [x] When resolving a specific issue, it's referenced in the PR's description (e.g. `Fix #xxx[,#xxx]`, where "xxx" is the Github issue number)
- [ ] All tests are passing
- [x] New/updated tests are included
- [x] Changelog was modified

If adding a **new feature**, the PR's description includes:
- [x] A convincing reason for adding this feature (to avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it)

**Other information:**

PR #6499, opened 2023-08-11
Enh/6486 add module not available option for containers

marc-farre:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] All tests are passing
- [x] Changelog was modified

**Other information:**

Issue https://github.com/humhub/humhub/issues/6486


protected/humhub/modules/stream/models/filters/ModuleStreamFilter.php
```diff
@@ -0,0 +1,61 @@
...
+use humhub\modules\space\models\Space;
+use humhub\modules\user\models\User;
+use ReflectionClass;
+use Yii;
+
+class ModuleStreamFilter extends StreamQueryFilter
+{
+    public function apply(): void
+    {
+        if ($this->streamQuery->container instanceof ContentContainerActiveRecord) {
```
yurabakhtin, line 23: @funkycram My original solution was this:

```php
$this->excludeContainersWithNotAvailableModule(Space::class);
if (!$this->streamQuery->container instanceof Space) {
    $this->excludeContainersWithNotAvailableModule(User::class);
}
```

but you made this as:

```php
if ($this->streamQuery->container instanceof ContentContainerActiveRecord) {
    $containerClass = $this->streamQuery->container instanceof Space ? Space::class : User::class;
    $this->excludeContainersWithNotAvailableModule($containerClass);
}
```

I would like to explain why I suggessted the solution.
The stream can be called from 3 sides:

1. Dashboard, here `$this->streamQuery->container === null` because we display entries from all spaces and users,
2. Space wall, here the container is `Space::class`, and here we display entries only from the Space,
3. User profile wall, here the container is `User::class`, and here we display enrties of the User also some entries may go from Spaces where the User posted them.

Correct me if I am wrong.
If I am right then my original solutions should be used instead of current, because:

- the Space filter `$this->excludeContainersWithNotAvailableModule(Space::class)` must be applied for all 3 points.
- the User filter `$this->excludeContainersWithNotAvailableModule(User::class)` should be applied only for 1st and 3rd cases when the stream container is `null` or `User::class`, it is why I use there the condition `!$this->streamQuery->container instanceof Space`.

Could please test these 3 cases(Dashboard, Space, User wall stream) to be sure all excluded as expected?
Thank you!

protected/humhub/modules/stream/models/filters/ModuleStreamFilter.php
```diff
@@ -0,0 +1,61 @@
...
+
+        if (empty($moduleIds)) {
+            return;
+        }
+
+        $excludedContentClasses = [];
+        foreach ($moduleIds as $moduleId) {
+            $module = Yii::$app->getModule($moduleId);
+            if ($module instanceof ContentContainerModule) {
+                $excludedContentClasses = array_merge($excludedContentClasses, $module->getContentClasses());
```
yurabakhtin, line 48: @funkycram @luke- I think we should be sure that all modules extended from `ContentContainerModule` properly define the classes in the method `getContentClasses()`, because during tests I find at least one module is "Files" https://github.com/humhub/cfiles/blob/master/Module.php which doesn't there records like `humhub\modules\cfiles\models\File::class`.

I didn't investigate deeply why it was missed there, probably some reason exists.
But if we set the module "Files" to new option "Not available" then the `File` records will not be hidden from a wall stream as expected, because the records will not be excluded because of `humhub\modules\cfiles\Module->getContentClasses()` returns empty array now.

yurabakhtin (changes requested):

protected/humhub/modules/stream/models/filters/ModuleStreamFilter.php
funkycram: Done in commit https://github.com/humhub/humhub/pull/6499/commits/05cbb7257374231744c67b34804adc78e241bfae

protected/humhub/modules/stream/models/filters/ModuleStreamFilter.php
yurabakhtin: Discussed here https://github.com/humhub/humhub/issues/6486#issuecomment-1678532487

protected/humhub/modules/stream/models/StreamQuery.php
```diff
@@ -140,7 +141,8 @@ class StreamQuery extends Model
         BlockedUsersStreamFilter::class,
         DateStreamFilter::class,
         DraftContentStreamFilter::class,
-        ScheduledContentStreamFilter::class
+        ScheduledContentStreamFilter::class,
+        ModuleStreamFilter::class,
```
yurabakhtin, line 145: @funkycram I think init of this filter should be moved into `humhub\modules\stream\models\ContentContainerStreamQuery` because only this query class has the property `$container` which is required for the filter.
We need it at least for fixing tests here https://github.com/humhub/humhub/actions/runs/5866247719/job/15904674985#step:18:1437:

tests_stream_query

The `ContentContainerStreamQuery` can be updated with this code:

```php
/**
 * @inheritdoc
 */
public function init()
{
    $this->filterHandlers[] = ModuleStreamFilter::class;
    parent::init();
}
```

yurabakhtin (changes requested):

protected/humhub/modules/stream/models/StreamQuery.php
funkycram: Yes, good point @yurabakhtin done in commit https://github.com/humhub/humhub/pull/6499/commits/1813ae3b8fbf8e568aa6551a7e235988e729dd8a

funkycram: @yurabakhtin we still have the same error from `StreamQueryTest`:
Getting unknown property: humhub\modules\stream\models\StreamQuery::container  

I don't understand what's wrong as there are no change in `StreamQuery`. Any idea?

yurabakhtin: @funkycram I see after moved the filter here https://github.com/humhub/humhub/pull/6499/commits/1813ae3b8fbf8e568aa6551a7e235988e729dd8a the error is gone.

yurabakhtin (approved):

protected/humhub/modules/stream/models/StreamQuery.php
funkycram: Fine. So it's just that I hadn't waited long enough!

marc-farre:
@luke- I've updated the CHANGELOG so now they are no conflict anymore if you want to merge this PR in the develop branch.

marc-farre:
@yurabakhtin I've updated the CHANGELOG file to match with the current develop branch so we can merge them, but now I have failing tests on all PRs:
- This one
- https://github.com/humhub/humhub/pull/6365
- https://github.com/humhub/humhub/pull/6611
- https://github.com/humhub/humhub/pull/6616

Is there a global problem with tests?

luke- (approved):

yurabakhtin:
> Is there a global problem with tests?

@marc-farre yes, I think it is related to https://github.com/humhub/humhub/issues/6714.

luke- (approved):

PR #6503, opened 2023-08-12
Enh: Introduce FindInstanceInterface with methods to get (cached) instances - Part 1

martin-rueegg:
Since #6457 was only a partial solution for https://github.com/humhub/humhub/pull/6375 as it

1. uses a non-standard name for `findMembership` and
1. deletes the cache only when memberships are updated, and
1. does the latter in an insufficient way (ignores both [`deleteAll()` and `saveAll()`](https://github.com/humhub/humhub/pull/6457#issuecomment-1650977793) and [linked content](https://github.com/humhub/humhub/pull/6457#issuecomment-1655336797)),

there is an improved attempt in #6463 to address this in a broader way, and ultimately #6482, which addresses some more issues.

To break #6463 down, I've created this PR for review, and I've submitted towards the `statable` branch, as #6430 also bilds on top of #6463. Once, #6463 is merged in `statable` too, both can be merged int `develop`, which I recommend to happen before v1.15 is released.

- Bugfix
- Feature
- Refactor

- No. However, module developers SHOULD adapt to use the new implementation of `findInstance`, rather than `findOne` where the searched object is only identified by either `id` or `guid` (see additional examples in #6463).

- [ ] All tests are passing
- [x] Changelog was modified

### DIFF
--- a/CHANGELOG-DEV.md
+++ b/CHANGELOG-DEV.md
@@ -6,10 +6,11 @@ HumHub Changelog
 - Fix #6472: Initialization of account profile field type "Markdown"
 - Fix #6471: Wording "Default Homepage" in Space Default Settings
 - Fix #6468: Module Administration - Marketplace Links broken without Pretty URLs
-- Enh #6469: Added Info text for Marketplace page 
+- Enh #6469: Added Info text for Marketplace page
 - Fix #112: Reorder Table Rows
 - Fix #6476: Fix module disabling in queue
 - Enh #6469: Implement conditions for `fixed-settings` in config
+- Enh #68: Online Indicator Position
 - Fix #6492: Fix module form "Set as default"
 - Fix #6457: Regression with membership cache. Also move cache to `Membership::findMembership`.
 
--- a/CHANGELOG-DEV.md
+++ b/CHANGELOG-DEV.md
@@ -10,6 +10,7 @@ HumHub Changelog
 - Fix #112: Reorder Table Rows
 - Fix #6476: Fix module disabling in queue
 - Enh #6469: Implement conditions for `fixed-settings` in config
+- Fix #6492: Fix module form "Set as default"
 - Fix #6457: Regression with membership cache. Also move cache to `Membership::findMembership`.
 
 1.15.0-beta.1 (July 31, 2023)
--- a/protected/humhub/components/CacheableActiveQuery.php
+++ b/protected/humhub/components/CacheableActiveQuery.php
@@ -1,0 +1,115 @@
+<?php
+
+/*
+ * @link      https://www.humhub.org/
+ * @copyright Copyright (c) 2017-2023 HumHub GmbH & Co. KG
+ * @license   https://www.humhub.com/licences
+ */
+
+namespace humhub\components;
+
+use humhub\exceptions\InvalidArgumentTypeException;
+use Yii;
+use yii\db\ActiveQuery;
+use yii\db\ActiveRecord;
+
+/**
+ * ActiveQueryContent is an enhanced ActiveQuery with additional selectors for especially content.
+ *
+ * @inheritdoc
+ */
+class CacheableActiveQuery extends ActiveQuery
+{
+    public function findFor($name, $model)
+    {
+        $result = parent::findFor($name, $model);
+
+        if (!$result instanceof ActiveRecord) {
+            return $result;
+        }
+
+        static::cacheProcessVariants('set', $result);
+
+        return $result;
+    }
+
+    public static function normaliseObjectIdentifier($classOrObject, $id = null, ?string &$idWasUsed = null): string
+    {
+        if (is_object($classOrObject)) {
+            $classOrObject = get_class($classOrObject);
+        } elseif (!is_string($classOrObject)) {
+            throw new InvalidArgumentTypeException(
+                __METHOD__,
+                [1 => '$classOrObject'],
+                ['object', 'string'],
+                $classOrObject
+            );
+        }
+
+        $idWasUsed = '';
+        $id = $id === 0 || !empty($id) ? "__" . ($idWasUsed = implode('_', (array)$id)) : '';
+
+        return str_replace('\\', '_', ltrim($classOrObject, '\\')) . $id;
+    }
+
+    /**
+     * @param string $action
+     * @param ActiveRecord $record
+     * @param array $properties
+     *
+     * @return void
+     */
+    public static function cacheProcessVariants(string $action, ActiveRecord $record, array $properties = ['id', 'guid']): void
+    {
+        static $runtimeCache;
+
+        $runtimeCache ??= Yii::$app->runtimeCache;
+
+        $done = [];
+        $class = get_class($record);
+
+        $pk = $record->getPrimaryKey(true);
+
+        $runtimeCache->$action(self::normaliseObjectIdentifier($class, $pk, $identifier), $record);
+
+        $done[] = $identifier;
+
+        foreach ($properties as $identifier) {
+            if ($record->hasAttribute($identifier)) {
+                $identifier = (string)$record->$identifier;
+                if (!in_array($identifier, $done, true)) {
+                    $runtimeCache->$action(self::normaliseObjectIdentifier($class, $identifier, $identifier), $record);
+                    $done[] = $identifier;
+                }
+            }
+        }
+
+        /**
+         * Check if we have the related record cached in the polymorphic behavior, so we can delete the cache by ID.
+         * (This is not fully bullet-proof, as the object might still be saved in the cache, but only under the guid key.)
+         */
+        if ($action !== 'delete' || !$record->hasMethod('getPolymorphicRelation')) {
+            return;
+        }
+
+        if ($model = $record->getPolymorphicRelation(false) ?? $runtimeCache->get(self::normaliseObjectIdentifier($record->{$record->classAttribute}, $record->{$record->pkAttribute}))) {
+            static::cacheProcessVariants($action, $model, $properties);
+        }
+    }
+
+    public static function cacheDeleteByClass($class, $condition)
+    {
+        $cache = Yii::$app->runtimeCache;
+
+        if (is_array($condition) && $class::isPrimaryKey(array_keys($condition))) {
+            $key = self::normaliseObjectIdentifier($class, $condition);
+
+            if ($record = $cache->get($key)) {
+                self::cacheProcessVariants('delete', $record);
+                $cache->delete($key);
+            }
+        } else {
+            $cache->flush();
+        }
+    }
+}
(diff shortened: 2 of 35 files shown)


martin-rueegg:
@luke- could you please update `statable` with develop? Thank you!

@gevorgmansuryan this is the first step towards 6482, hence I'm including you here, too!

martin-rueegg:
_Originally posted by @gevorgmansuryan in https://github.com/humhub/humhub/issues/6457#issuecomment-1668305081:_
> @luke- Regarding `runtimeCache` potential issues we need to discuss. Adding cache delete in all model's updating lifecycle methods is not a good solution. Maybe we create some `behavior` that will handle it all automatically.

I've been looking a bit into the possibility of using a behavior here. It looks like an elegant solution at first, specially with the idea that [_behaviors can be attached and detached to a component dynamically without requiring modification of the component class_](https://www.yiiframework.com/doc/guide/2.0/en/concept-behaviors#pros-for-behaviors).

However, I have not found a way to actually do attach a behavior to let's say a ActiveRecord instance without modifying the thing. The reason being, that there is no event called when objects are instantiated so that I could listen to that event and then attach the behaviour. So we would have to
- find a reliable event to be able to do that
- create our own version of` \Yii` that does fire an event e.g `\Yii::EVENT_BEFORE_CREATE_OBJECT` and `\Yii::EVENT_AFTER_CREATE_OBJECT` (which would be a great thing to introduce anyway!)

But with that behavior we would have to listen to events again (after save and update) and we cannot add static functions (as `findInstance()`) AFAIK.

So we could do the cache cleaning using a behavior, but we would still need to change the classes in order to implement `findInstance()` method.

Or is there anything I've missed here?

luke-:
@martin-rueegg 

> So we could do the cache cleaning using a behavior, but we would still need to change the classes in order to implement `findInstance()` method.
> 
> Or is there anything I've missed here?

Yes, I don't think we can add e.g. a `SomeAR::findInstance(1)` method just using a behavior.

---

In general, I frankly don't know if the current solution, which is really very nicely worked out, doesn't introduce too much new complexity and non standard (Yii2) behavior for me. 

I basically like the idea of a `Model::findInstance($pkOrInstance)` apprach. 

But I would have liked to implement this with 2-3 new methods. 

This PR really covers a lot of cases: Relations, `::deleteAll()`, Conditions, Multi PKs, Options like (exceptionMessageSuffix, exception type) etc. 

When I look at the "Membership" model as an example, I really have trouble understanding right away what exactly is going on in the  `::findInstance()` method for example.   Also the introduction of new relations `hasOne()` to `hasOneCached()` really requires deeper understanding.

Currently, I think it's good that we rely very heavily on well-documented Yii 2 standard components. If we now start to change or introduce other approaches, we have to document it somehow but also maintain it in the long run. 

So I would prefer to introduce the simplest/clearest possible caching, which may only covers 80% of the cases, rather than something complex to achieve a bit higher coverage.   

For example, it is probably clear for most Yii 2 developers that `deleteAll()` or `updateAll()` do not execute events/before/after/validate and should therefore only be executed very selectively. So if no runtime cache is cleared here, I would be perfectly fine with that. 

I'm not sure if I'm being too critical here... maybe we should discuss this with the team...

martin-rueegg:
@luke- Thank you for your review and comments!

> I'm not sure if I'm being too critical here... maybe we should discuss this with the team...

I guess that's your job and I'm grateful you are a critical reviewer!

> This PR really covers a lot of cases: Relations, `::deleteAll()`, Conditions, Multi PKs, Options like (exceptionMessageSuffix, exception type) etc.

Yes, that's true. Some explanatory thoughts:

- **Relations:** Yep, a creative apprach. Motivation was that a lot of record lookups happen through this standard way of getting objects, particularly to get `Content` or a `ContentActiveRecord`. However, it could be probably reduced in the sense that only the result of the lookup is cached, not the `::getMyActiveRecord()` itself modified.
- **deleteAll/updateAll:** The thing is, that this is the same as `updateAll()` (in terms of what needs to be done cache-wise). And `updateAll()` is used by `\yii\db\BaseActiveRecord::updateAttributes()` which in turn is used in different places (e.g. (un)pinning content, but first and foremost in updating the status [currently `softDelete()`]). This would have to be discouraged.
- **Conditions:** came along since quite often, a very specific item is looked for, but only if a given status or visibility is given. Now, since these two criteria are looked up repeatedly it still does make sense to cache the object to avoid further lookups. Hence the specifically called `$simpleCondition`, as it only matches properties and does not support string conditions or ConditionExpressions as in eg `findByCondition()` would.
- Multi-PK: Yes, I don't see this as a big issue as it draws on standard Yii capabilities. The `Membership` is an exception here, as it 
  1. uses a unique key (not in an db-enforced sense) that is NOT the PK.
  2. the identifiers themselves vary accross the code and are sometimes User objects or IDs. This could be simplified by requiring the calling code to do something like
     `Membership::findInstance([Space::findInstanceAsId($space), User::findInstanceAsId($user]))` rather than just
     `Membership::findInstance([$space, $user])`.
     However, I hope to have made the code more understandable now by adding some comments
- **Option `exception` type:** The idea was to be able to throw a HTTP 404/403 exception, if a record is not available. But this is an edge case. Code has been removed.
- **Option `exceptionMessageSuffix`:** This is just to be kind with developers, showing them what they've missed. The suffix is alredy supported by the given `Exception` and so I added it. Can be removed obviously, or gently ignored. :-)

> When I look at the "Membership" model as an example, I really have trouble understanding right away what exactly is going on in the `::findInstance()` method for example.

I hope, my new comments in the code help. See also above.

> Also the introduction of new relations `hasOne()` to `hasOneCached()` really requires deeper understanding.

Agree. `hasOneCached()` can be discarded! See new commits.

> Currently, I think it's good that we rely very heavily on well-documented Yii 2 standard components. If we now start to change or introduce other approaches, we have to document it somehow but also maintain it in the long run.

Sure.

> So I would prefer to introduce the simplest/clearest possible caching, which may only covers 80% of the cases, rather than something complex to achieve a bit higher coverage.

> From: https://github.com/humhub/humhub/pull/6463#issuecomment-1682807932
> However, for me arises the question: why not using the default db-caching mechanism from Yii2 and just deal with it's clearing as we do in the runtimeCache?

Was that looked into? I mean, why not using the built-in db-query cache?

> For example, it is probably clear for most Yii 2 developers that `deleteAll()` or `updateAll()` do not execute events/before/after/validate and should therefore only be executed very selectively. So if no runtime cache is cleared here, I would be perfectly fine with that.

See my comments above. I don't think we can ignore the `updateAll` unless some code is re-manufactured to avoid it. But changing a state (soft-delete so far) can change the result that has to be returned from an ajax call for example. Having that one record in the cache, that has been updated by the `updateAll` via the `updateAttributes`, will completely ignore this update (that happened in the same request).

> maybe we should discuss this with the team...

Sure. Appreciate any feedback!

luke-:
@martin-rueegg Thanks for the response. I'm soon in a meeting and need some time to reply. In the meantime an untested/buggy but totally simple/stupid example of a `findInstance()` implementation which would I prefer.

https://github.com/humhub/humhub/commit/64f0fccb08149b04b3ec85fcfd4456827aa69ba6

Of course there is (or it need) much room for improvements. (Interface, Behavior, Trait, Multi PK,..........)

martin-rueegg:
@luke- Thanks! I had a look into the dependency stuff but did not get my head around. That's helpful in your example. Well, helpful in the sense of understanding the usage of dependencies. However, since we do [not currently serialize](https://github.com/humhub/humhub/blob/develop/protected/humhub/config/common.php#L115) the cached values, the dependency [would simply be ignored](https://www.yiiframework.com/doc/api/2.0/yii-caching-cache#getOrSet()-detail): `$dependency` _"This parameter is ignored if [$serializer](https://www.yiiframework.com/doc/api/2.0/yii-caching-cache#$serializer-detail) is false."_ We would either have to change that and serialize stuff (creating overhead in serialization and unserialization and potential sideeffects that would have to be addressed in `__sleep()` and `__wakeup()` methods), or create an implementation that still deals with dependencies. (My own implementation in #6482 does that in a different way.)

But, yes, I get your point. And it could be a reasonable solution to avoid invalid objects in the store (if we can solve this with the dependency).

Things I see unresolved:
1. Same object is retrieved by different keys:
   -  `id`
   - `guid`
   - `object_model` + `object_id`
   - `space_id` + `user_id`.
   
   One could easily argue, that
   1. we only store PKs. Which would mean disabling cache for `Content`, `ContentContainer` and `Membership` for sure as they are usually retrieved using a combination of fields _that do not constitute the PK._
      While `Content` is cached already in the `PolymorphicRelation`, the latter two record classes are widely used repeatedly for the same values (**hence caching makes most sense**).
   3. the record is then retrieved two or three times, at most. Yes, that would be ok. But again, the _clearing_ is the issue here. If we could get dependency work, we could always add two tags, one for the class name and one for the unique identifier using the PK. Hoever, here we could no longer use the `getOrSet()` method, as we would know the PK only once we have the record in hand. This would require coding the logic of `getOrSet` into our function, which is of course not difficult, but easily adding a few lines.
4. Related record retrieval is not cached. Although, caching it is actually [quite simple](https://github.com/metaworx/humhub/blob/enh/new-find-interface-1/protected/humhub/components/CacheableActiveQuery.php#L27).
5. Related records are not cleared (e.g. if an `Content` is soft deleted, then the `Post` record needs to be removed from the cache, as otherwise the `Post:getInstance()` would return the thing although it should not.
6. If you add a bit of type and other validity checks, the complexity of the function will soon reach the one of [`findInstanceHelper`](https://github.com/metaworx/humhub/blob/enh/new-find-interface-1/protected/humhub/components/FindInstanceTrait.php#L35).

I'm getting the impression, we're trying to reduce complexity so much but missing the targeting by doing so: If for the classes where the most repeated calls are expected, we do not have a cache, then for the others, we need to cache the objects often twice, since it's first accessed by one key, and then the other, then we create more overhead than we solve.

Also, [your draft](https://github.com/humhub/humhub/commit/64f0fccb08149b04b3ec85fcfd4456827aa69ba6) is 60 lines long, and it will also require some other 10 lines for the missing `findOne()` method and some other 20-30 lines for validation.. [My version](https://github.com/metaworx/humhub/blob/enh/new-find-interface-1/protected/humhub/components/FindInstanceTrait.php#L23) is roughly 150 Lines plus roughly another 110 lines in [`CacheableActiveQuery`](https://github.com/metaworx/humhub/blob/enh/new-find-interface-1/protected/humhub/components/CacheableActiveQuery.php#L25). Making the ratio being somewhere around 100:260. We're not talking about an exorbitant complexity here. :-) However, the coverage of cases where the cache becomes useful seems to be much better, given the limitations laid out above ...

Of course, we could gain some lines by removing some options, or the exception suffix, or some other small stuff. But while saving a few lines, we have to add back those checks in the originating code. In the end, that's making the change for the 181 lines we save with this PR. If you deduct those from my 260 lines above, I might still end up better with only about 80 net plus. :-)

martin-rueegg:
@luke- @gevorgmansuryan 

Before we can really continue with this, 
- #6527 needs to be addressed
- `develop` should be merged into `statable`

PR #6509, opened 2023-08-14
Fix #6507: Accepting an invite does not reload page

### LINKED ISSUE #6507: Accepting an invite does not reload page
Invite someone to a space. Log in as that user, accept the invite. 
When someone accepts an invite, the page should reload automatically so they see the content they should see as a member.
### What do you get instead?
User only sees content that they would see if they were not a member. The user must manually reload the page to see what they should see.
### Additional info
This should be a very simple fix for force a reload on accepting invite (or just send them to the space again, which will in-turn reload it)

| Q                | A
| ---------------- | ---
| HumHub version   | 1.x
| PHP version      | 7.x
| Operating system | Linux

yuriimaz:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] All tests are passing
- [x] Changelog was modified

**Other information:**


luke-:
@yurabakhtin Can you please review this?

yurabakhtin:
> @yurabakhtin Can you please review this?

@luke- I don't think we should reload page after accepting of invite in all cases, because we have the button in several places:

1. Space header - here we must fix it
2. Spaces directory - here we should keep the ajax approach, because it was implemented here https://github.com/humhub/team_tasks/issues/117#issuecomment-855660572 and it works like this:

space-directory

I.e. I don't agree the changes in the JS code of this PR #6509, we should keep the JS for the 2nd point, but for 1st point we should fix the widget `MembershipButton` which is used in both cases.
I have done this in PR https://github.com/humhub/humhub/pull/6513, so the button "Accept Invite" works now in `link` mode from the Space Header side and in `ajax` mode from the Space Directory.
As you can see here https://github.com/humhub/humhub/pull/6513/files#diff-b043931572814ed2968ae1351ae1a757be8897e216a860f316b2002429f8e677R27-R28 the buttons "Join"(auto joining without approving) and "Accept Invite" works in link mode only from the Space Header.
I have tested all other buttons like "Join"(with approving), "Decline Invite" and "Pending"(withdraw your request to join Space) and we don't need to have the buttons in `link` mode even from Space Header side because after their action the space view is not changed because no new permission appears after their actions.

---

If you look at my screenshot above then we have there one issue after "Accept Invite" is done, it is the members count is not updated to `2`. Do you think we should update it by ajax as well?

luke- (approved):

PR #6511, opened 2023-08-14
Enh: Add module manager test and add some improvements and bug fixes

martin-rueegg:

- Bugfix
- Feature
- Tests

**Please have a look at the commit titles to get an idea!**

- Unlikely. There are some return types added to `ModuleManager` though.

- a developer would have to adapt their inheritance of `ModuleManager` if they use a custom Manager in their setup or for the module itself.

    Would it make sense to create different issues for documentation/tracking purposes? 
- [x] Changelog was modified

**Other information:**


martin-rueegg:
@luke-,

While writing a test case to test the `ModuleManager` I came across a anomailty:

While registering modules, there is no test if the module is inheriting from `\humhub\components\Module`. However, in many other places, the module will fail if it does not do so, e.g. in
- `getModules()`,
- `filterModulesByKeyword` (class is not tested, but methods are accessed that do not exist in `\yii\base\Module`)
- `enable()` and `disable()` require this type in the parameter
- `enableModules()` and `disableModules()`

I would suggest to test for either that inheritance, or even better to create an interface that needs to be met (which is to be implemented by `\humhub\components\Module`, of course).

protected/humhub/components/ModuleManager.php
```diff
@@ -213,23 +216,91 @@ public function register($basePath, $config = null)
...
+                if (empty($var)) {
+                    $error("required property '$property' missing!", $throw);
+                    return false;
+                }
+
+                return true;
+            };
+
+            foreach ($events as $event) {
+                if (empty($event)) {
```
luke-, line 250: It might be good to have a private method e.g. `validateEventCallback($event)`?

protected/humhub/components/ModuleManager.php
```diff
@@ -213,23 +216,91 @@ public function register($basePath, $config = null)
...
         // Register Event Handlers
-        if (isset($config['events'])) {
-            foreach ($config['events'] as $event) {
-                $eventClass = $event['class'] ?? $event[0];
-                $eventName = $event['event'] ?? $event[1];
-                $eventHandler = $event['callback'] ?? $event[2];
+        $events = $config['events'] ?? null;
+        $strict = $config['strict'] ?? false;
+        while (!empty($events)) {
+            $error = static function (string $message, bool $throw = false) use (&$config, $basePath) {
```
luke-, line 222: Can we simply use a private method here? Or did I missed something?

protected/humhub/components/ModuleManager.php
```diff
@@ -407,6 +490,10 @@ public function getModule($id, $throwOnMissingModule = true)
             return Yii::createObject($class, [$id, Yii::$app]);
         }
 
+        if (is_dir($id) && is_file("$id/config.php")) {
```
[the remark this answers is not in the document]
martin-rueegg: @luke- Ok, I've updated the code (not yet pushed). Is there a specific reason for this?

Btw, you've missed the "not" on the code style page ... ;-)

protected/humhub/components/ModuleManager.php
martin-rueegg: rather than extracting `validateEventCallback($event)`, I've extracted the entire `registerEventHandlers()`. As such I could simplify the layout of the code a bit. Do you want to take another look?

protected/humhub/components/ModuleManager.php
martin-rueegg: Ah, yes, see previous comment.

protected/humhub/components/ModuleManager.php
```diff
@@ -213,23 +216,106 @@ public function register($basePath, $config = null)
...
+            $error('events must be traversable', $strict);
+            return;
+        }
+
+        $getProperty = static function ($event, &$var, string $property, int $index, bool $throw = false) use ($error): bool {
+
+            $var = $event[$property] ?? $event[$index] ?? null;
+
+            if (empty($var)) {
+                $error("required property '$property' missing!", $throw);
```
martin-rueegg, line 261: @luke-, I guess I should also update this line here, right?

protected/humhub/components/ModuleManager.php
```diff
@@ -213,23 +216,106 @@ public function register($basePath, $config = null)
...
+    {
+        $events = $config['events'] ?? null;
+        $strict = $config['strict'] ?? false;
+
+        if (empty($events)) {
+            return;
+        }
+
+        $error = static function (string $message, bool $throw = false) use (&$config, $basePath) {
+            $message = "Configuration at $basePath has an invalid event configuration: $message";
```
martin-rueegg, line 242: @luke-, I guess I should also update this line here, right?

protected/humhub/components/ModuleManager.php
luke-: Yes, would be great.

protected/humhub/components/ModuleManager.php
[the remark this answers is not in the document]
luke-: @martin-rueegg Hehe, thanks. No specifc reason, mainly to have it uniform and I do not like this mix.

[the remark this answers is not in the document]
martin-rueegg: @luke- got you. No argument against preferences....

martin-rueegg:
Requested changes applied!

@luke-, how would you like the CHANGELOG to be ammended? I guess it makes sense to have more than one line. Shall I create issues or add multiple lines with the same PR number, in order to document the separate the changes?

luke- (approved):

luke-:
@martin-rueegg As you like, but probably it is easier to add several lines with the same PR number.

Is it sufficient for you if we merge this PR into v1.16? I would like to minimize changes to the current beta as little as possible.

luke- (approved):

PR #6512, opened 2023-08-15
Enh: Attempt to show better error messages when DB-Connection is invalid

martin-rueegg:

- Bugfix

- No

- [x] Changelog was modified

**Other information:**

When the connection information is wrong, or the database server name resolution fails, there are misleading error messages as there is actually no error handling of that situation so far.

This PR is a poor man's attempt to give the user some sort of helpful information. Yes, it does not look pretty and styled, but that's partly because the theming tries to access the db itself...

Particularly un-informative is the error when you try to run migrations with an invalid connection configuration.

### DIFF
--- a/protected/yii
+++ b/protected/yii
@@ -8,6 +8,8 @@
  * @license http://www.yiiframework.com/license/
  */
 
+use humhub\helpers\DatabaseHelper;
+
 defined('YII_DEBUG') or define('YII_DEBUG', true);
 
 // fcgi doesn't have STDIN and STDOUT defined by default
@@ -25,6 +27,11 @@ $config = yii\helpers\ArrayHelper::merge(
     require(__DIR__ . '/config/console.php')
 );
 
-$application = new humhub\components\console\Application($config);
-$exitCode = $application->run();
-exit($exitCode);
+try {
+    $exitCode = (new humhub\components\console\Application($config))->run();
+    exit($exitCode);
+} catch (\Throwable $ex) {
+    if (null === DatabaseHelper::handleConnectionErrors($ex)) {
+        throw $ex;
+    }
+}
--- a/CHANGELOG-DEV.md
+++ b/CHANGELOG-DEV.md
@@ -1,6 +1,10 @@
 HumHub Changelog
 ================
 
+1.16.0 (Unreleased)
+--------------------------
+- Enh #6512: Show error messages when DB connection configuration is invalid
+
 1.15.0-beta.2 (Unreleased)
 --------------------------
 - Enh #6478: Add pseudo test class to allow population of DB with standard test data
--- a/index.php
+++ b/index.php
@@ -7,6 +7,8 @@
  */
 
 // comment out the following two lines when deployed to production
+use humhub\helpers\DatabaseHelper;
+
 defined('YII_DEBUG') or define('YII_DEBUG', true);
 defined('YII_ENV') or define('YII_ENV', 'dev');
 
@@ -22,4 +24,10 @@ $config = yii\helpers\ArrayHelper::merge(
     require(__DIR__ . '/protected/config/web.php')
 );
 
-(new humhub\components\Application($config))->run();
+try {
+    (new humhub\components\Application($config))->run();
+} catch (\Throwable $ex) {
+    if (null === DatabaseHelper::handleConnectionErrors($ex)) {
+        throw $ex;
+    }
+}
--- a/protected/humhub/commands/MigrateController.php
+++ b/protected/humhub/commands/MigrateController.php
@@ -9,6 +9,7 @@
 namespace humhub\commands;
 
 use humhub\components\Module;
+use humhub\helpers\DatabaseHelper;
 use Yii;
 use yii\console\Exception;
 use yii\web\Application;
@@ -78,9 +79,17 @@ class MigrateController extends \yii\console\controllers\MigrateController
      */
     public function beforeAction($action)
     {
-        // Make sure to define default table storage engine
-        if (in_array(Yii::$app->db->getDriverName(), ['mysql', 'mysqli'], true)) {
-            Yii::$app->db->pdo->exec('SET default_storage_engine=' . Yii::$app->params['databaseDefaultStorageEngine']);
+        // Make sure to define a default table storage engine
+        $db = Yii::$app->db;
+
+        try {
+            $db->open();
+        } catch (\Throwable $ex) {
+            DatabaseHelper::handleConnectionErrors($ex);
+        }
+
+        if (in_array($db->getDriverName(), ['mysql', 'mysqli'], true)) {
+            $db->pdo->exec('SET default_storage_engine=' . Yii::$app->params['databaseDefaultStorageEngine']);
         }
         return parent::beforeAction($action);
     }
--- a/protected/humhub/components/console/Application.php
+++ b/protected/humhub/components/console/Application.php
@@ -59,7 +59,7 @@ class Application extends \yii\console\Application implements \humhub\interfaces
             ));
         }
 
-        if (BaseSettingsManager::isDatabaseInstalled()) {
+        if (BaseSettingsManager::isDatabaseInstalled(Yii::$app->params['databaseInstalled'] ?? false)) {
             $baseUrl = Yii::$app->settings->get('baseUrl');
             if (!empty($baseUrl)) {
                 if (Yii::getAlias('@web', false) === false) {
--- a/protected/humhub/helpers/DatabaseHelper.php
+++ b/protected/humhub/helpers/DatabaseHelper.php
@@ -1,0 +1,147 @@
+<?php
+
+/*
+ * @link      https://www.humhub.org/
+ * @copyright Copyright (c) 2017-2023 HumHub GmbH & Co. KG
+ * @license   https://www.humhub.com/licences
+ */
+
+namespace humhub\helpers;
+
+use Throwable;
+use Yii;
+
+/**
+ * @since 1.15
+ */
+class DatabaseHelper
+{
+    public static function handleConnectionErrors(
+        Throwable $ex,
+        bool $print = true,
+        bool $die = true,
+        bool $forcePlainText = false
+    ): ?string {
+        static $last = false;
+
+        if (!$ex instanceof \yii\db\Exception) {
+            return null;
+        }
+
+        if ($last) {
+            return null;
+        }
+
+        $last = true;
+
+        $trace = debug_backtrace(0);
+        $trace = end($trace);
+        if ($trace && $trace['function'] === 'handleException' && $trace['args'][0] instanceof \yii\db\Exception) {
+            return null;
+        }
+
+        switch ($ex->getCode()) {
+            case 2002:
+                $error = 'Hostname not found.';
+                break;
+
+            case 1044:
+                $error = 'Database not found or not accessible.';
+                break;
+
+            case 1049:
+                $error = 'Database not found.';
+                break;
+
+            default:
+                $error = $ex->getMessage();
+        }
+
+        /**
+         * @see https://www.php.net/manual/en/ref.pdo-odbc.connection.php
+         * @see https://www.php.net/manual/en/ref.pdo-ibm.connection.php
+         * @see https://www.php.net/manual/en/ref.pdo-pgsql.connection.php
+         */
+        $dsn = preg_replace(
+            '@((?<=:|;)(?:user|uid|User ID|pwd|password)=)(.*?)(?=;(?:$|\w+=)|$)@i',
+            '$1****',
+            Yii::$app->db->dsn
+        );
+
+        try {
+            $additionalInfo = json_encode([get_class($ex), ...$ex->errorInfo], JSON_THROW_ON_ERROR);
+        } catch (\JsonException $e) {
+            $additionalInfo = 'N/A';
+        }
+
+        while ($ex->getPrevious()) {
+            $ex = $ex->getPrevious();
+        }
+
+        $htmlMessage = defined('YII_DEBUG') && YII_DEBUG
+            ? sprintf('
+<h1>Invalid database configuration</h1>
+<p><strong>%s</strong></p>
+<p>The following connection string was used:<br><code>%s</code></p>
+<br>
+<h2>Technical information</h2>
+<p><code>%s</code></p>
+<p><pre>%s</pre></p>
+', $error, $dsn, $additionalInfo, $ex)
+            : sprintf('
+<h1>Invalid database configuration</h1>
+<p><strong>%s</strong></p>
+', $error);
+
+        $txtMessage = defined('YII_DEBUG') && YII_DEBUG
+            ? sprintf('
+Invalid database configuration
+==============================
+
+%s
+
+The following connection string was used:
+%s
+
+
+Technical information
+---------------------
+%s
+
+%s
+
+', $error, $dsn, $additionalInfo, $ex)
+            : sprintf('
+Invalid database configuration
+==============================
+
+%s
+
+The following connection string was used:
+%s
+
+
+Technical information
+---------------------
+%s
+
+', $error, $dsn, $additionalInfo);
+
+        if ($print) {
+            if ($forcePlainText) {
+                echo $txtMessage;
+            } elseif (Yii::$app instanceof \yii\console\Application && Yii::$app->controller instanceof \yii\console\Controller) {
+                Yii::$app->controller->stderr($txtMessage);
+            } else {
+                header("HTTP/1.1 500 Internal Server Error");
+                echo $htmlMessage;
+            }
+        }
+
+        if (!$die) {
+            return $txtMessage;
+        }
+
+        die(1);
+    }
+}
--- a/protected/humhub/libs/BaseSettingsManager.php
+++ b/protected/humhub/libs/BaseSettingsManager.php
@@ -10,12 +10,12 @@ namespace humhub\libs;
 
 use humhub\components\SettingActiveRecord;
 use humhub\exceptions\InvalidArgumentTypeException;
+use humhub\helpers\DatabaseHelper;
 use Stringable;
 use Yii;
 use yii\base\Component;
 use yii\base\InvalidArgumentException;
 use yii\base\InvalidConfigException;
-use yii\db\conditions\LikeCondition;
 use yii\db\StaleObjectException;
 use yii\helpers\Json;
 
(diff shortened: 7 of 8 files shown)


luke- (approved):
@martin-rueegg Thanks, looks good for me. If this is not only used on console usage, we need to ensure that no database creds are leaked.

martin-rueegg:
@luke- 
> @martin-rueegg Thanks, looks good for me. If this is not only used on console usage, we need to ensure that no database creds are leaked.

I see, great point, thank you!

Because the password would be included in the connection string? I guess, we could leave the output of the connection string to the console and otherwise print an error to the log.

Would that do?

martin-rueegg:
@luke- nope, it is also shown in web context. But I have entirely removed the DSN in web context. So the only way where the DSN would be shown on the web is while running migrations, as this is run as a console task from the webserver, right?

But a migration is only run by humhub admins? It could still be a problem as the humhub admin and the web server admin aren't necessarily the same people.

Otherwise I completely remove the DSN and write the message to a temp-file and only display the path of that temp file to the user (as it makes no sense sending anything to the log which might try to access the db as well).

I mean it's a rare case where this scenario happens, eg. when moving the installation from one server to another, when changing the DB server or things like that.

luke-:
@martin-rueegg Ah sorry, I missed the `index.php` change.

In the web context, I would only output details when the software is running in debug mode. (Currently by default).
https://github.com/humhub/humhub/blob/master/index.php#L9-L11 

Otherwise, there should be no detailed output at all in web.

--- 

In the migration (web)/console, there is no problem if there are some more detailed information. Maybe we can search and replace the DB dassword with ***?

martin-rueegg:
@luke-

> Ah sorry, I missed the `index.php` change.
> 
> In the web context, I would only output details when the software is running in debug mode. (Currently by default). https://github.com/humhub/humhub/blob/master/index.php#L9-L11
> 
> Otherwise, there should be no detailed output at all in web.

Actually, the index.php change wasn't there, but it was in the bootstrapping phase when checking the db-connection.
However, currently no detailed output is shown when the web application is running. However, I can re-add it if `YII_DEBUG === true` and/or `YII_ENV === 'dev'`

martin-rueegg:
> In the migration (web)/console, there is no problem if there are some more detailed information. Maybe we can search and replace the DB dassword with ***?

@luke- I've did my best: https://regex101.com/r/IRO3rG/2

# Console with debug
```
20:32 $ php7.4 $PHP_XDEBUG ${DEVILBOX_PROJECT_DOCROOT}/protected/yii migrate/up --includeModuleMigrations=1 --interactive=0

Invalid database configuration
==============================

Database not found.

The following connection string was used:
mysql:host=127.0.0.1;dbname=humhub_testx;pwd=****;foo=bar;user=****;


Technical information
---------------------
["yii\\db\\Exception","HY000",1049,"Unknown database 'humhub_testx'"]

PDOException: SQLSTATE[HY000] [1049] Unknown database 'humhub_testx' in /home/mdr/projects/humhub/htdocs/protected/vendor/yiisoft/yii2/db/Connection.php:722
Stack trace:
#0 /home/mdr/projects/humhub/htdocs/protected/vendor/yiisoft/yii2/db/Connection.php(722): PDO->__construct()
#1 /home/mdr/projects/humhub/htdocs/protected/vendor/yiisoft/yii2/db/Connection.php(637): yii\db\Connection->createPdoInstance()
#2 /home/mdr/projects/humhub/htdocs/protected/humhub/commands/MigrateController.php(86): yii\db\Connection->open()
#3 /home/mdr/projects/humhub/htdocs/protected/vendor/yiisoft/yii2/base/Controller.php(176): humhub\commands\MigrateController->beforeAction()
#4 /home/mdr/projects/humhub/htdocs/protected/vendor/yiisoft/yii2/console/Controller.php(180): yii\base\Controller->runAction()
#5 /home/mdr/projects/humhub/htdocs/protected/vendor/yiisoft/yii2/base/Module.php(552): yii\console\Controller->runAction()
#6 /home/mdr/projects/humhub/htdocs/protected/vendor/yiisoft/yii2/console/Application.php(180): yii\base\Module->runAction()
#7 /home/mdr/projects/humhub/htdocs/protected/vendor/yiisoft/yii2/console/Application.php(147): yii\console\Application->runAction()
#8 /home/mdr/projects/humhub/htdocs/protected/vendor/yiisoft/yii2/base/Application.php(384): yii\console\Application->handleRequest()
#9 /home/mdr/projects/humhub/htdocs/protected/yii(32): yii\base\Application->run()
#10 {main}
```

# Console without debug
```
20:34 $ php7.4 $PHP_XDEBUG ${DEVILBOX_PROJECT_DOCROOT}/protected/yii migrate/up --includeModuleMigrations=1 --interactive=0

Invalid database configuration
==============================

Database not found.

The following connection string was used:
mysql:host=127.0.0.1;dbname=humhub_testx;pwd=****;foo=bar;user=****;


Technical information
---------------------
["yii\\db\\Exception","HY000",1049,"Unknown database 'humhub_testx'"]
```

# web with debug

## Invalid database configuration

**Database not found.**

The following connection string was used:
mysql:host=127.0.0.1;dbname=humhub_testx;pwd=****;foo=bar;user=****;

### Technical information

```
["yii\\db\\Exception","HY000",1049,"Unknown database 'humhub_testx'"]

PDOException: SQLSTATE[HY000] [1049] Unknown database 'humhub_testx' in /shared/httpd/humhub/htdocs/protected/vendor/yiisoft/yii2/db/Connection.php:722
Stack trace:
#0 /shared/httpd/humhub/htdocs/protected/vendor/yiisoft/yii2/db/Connection.php(722): PDO->__construct('mysql:host=127....', 'humhub', 'humhub', NULL)
#1 /shared/httpd/humhub/htdocs/protected/vendor/yiisoft/yii2/db/Connection.php(637): yii\db\Connection->createPdoInstance()
#2 /shared/httpd/humhub/htdocs/protected/vendor/yiisoft/yii2/db/Connection.php(1067): yii\db\Connection->open()
#3 /shared/httpd/humhub/htdocs/protected/vendor/yiisoft/yii2/db/Connection.php(1054): yii\db\Connection->getMasterPdo()
#4 /shared/httpd/humhub/htdocs/protected/vendor/yiisoft/yii2/db/mysql/QueryBuilder.php(413): yii\db\Connection->getSlavePdo(true)
#5 /shared/httpd/humhub/htdocs/protected/vendor/yiisoft/yii2/db/mysql/QueryBuilder.php(437): yii\db\mysql\QueryBuilder->supportsFractionalSeconds()
#6 /shared/httpd/humhub/htdocs/protected/vendor/yiisoft/yii2/db/mysql/QueryBuilder.php(59): yii\db\mysql\QueryBuilder->defaultTimeTypeMap()
#7 /shared/httpd/humhub/htdocs/protected/vendor/yiisoft/yii2/base/BaseObject.php(109): yii\db\mysql\QueryBuilder->init()
#8 /shared/httpd/humhub/htdocs/protected/vendor/yiisoft/yii2/db/QueryBuilder.php(116): yii\base\BaseObject->__construct(Array)
#9 [internal function]: yii\db\QueryBuilder->__construct(Object(yii\db\Connection), Array)
#10 /shared/httpd/humhub/htdocs/protected/vendor/yiisoft/yii2/di/Container.php(411): ReflectionClass->newInstanceArgs(Array)
#11 /shared/httpd/humhub/htdocs/protected/vendor/yiisoft/yii2/di/Container.php(170): yii\di\Container->build('yii\\db\\mysql\\Qu...', Array, Array)
#12 /shared/httpd/humhub/htdocs/protected/vendor/yiisoft/yii2/BaseYii.php(345): yii\di\Container->get('yii\\db\\mysql\\Qu...', Array)
#13 /shared/httpd/humhub/htdocs/protected/vendor/yiisoft/yii2/db/mysql/Schema.php(225): yii\BaseYii::createObject('yii\\db\\mysql\\Qu...', Array)
#14 /shared/httpd/humhub/htdocs/protected/vendor/yiisoft/yii2/db/Schema.php(247): yii\db\mysql\Schema->createQueryBuilder()
#15 /shared/httpd/humhub/htdocs/protected/vendor/yiisoft/yii2/db/Connection.php(884): yii\db\Schema->getQueryBuilder()
#16 /shared/httpd/humhub/htdocs/protected/vendor/yiisoft/yii2/db/Query.php(157): yii\db\Connection->getQueryBuilder()
#17 /shared/httpd/humhub/htdocs/protected/vendor/yiisoft/yii2/db/Query.php(304): yii\db\Query->createCommand(Object(yii\db\Connection))
#18 /shared/httpd/humhub/htdocs/protected/vendor/yiisoft/yii2/web/DbSession.php(185): yii\db\Query->scalar(Object(yii\db\Connection))
#19 [internal function]: yii\web\DbSession->readSession('3fbe391c5e2a686...')
#20 /shared/httpd/humhub/htdocs/protected/vendor/yiisoft/yii2/web/Session.php(149): session_start()
#21 /shared/httpd/humhub/htdocs/protected/vendor/yiisoft/yii2/web/Session.php(699): yii\web\Session->open()
#22 /shared/httpd/humhub/htdocs/protected/vendor/yiisoft/yii2/web/User.php(691): yii\web\Session->get('__id')
#23 /shared/httpd/humhub/htdocs/protected/vendor/yiisoft/yii2/web/User.php(199): yii\web\User->renewAuthStatus()
#24 /shared/httpd/humhub/htdocs/protected/vendor/yiisoft/yii2/web/User.php(367): yii\web\User->getIdentity()
#25 /shared/httpd/humhub/htdocs/protected/vendor/yiisoft/yii2/base/Component.php(139): yii\web\User->getIsGuest()
#26 /shared/httpd/humhub/htdocs/protected/humhub/components/i18n/I18N.php(53): yii\base\Component->__get('isGuest')
#27 /shared/httpd/humhub/htdocs/protected/humhub/components/bootstrap/LanguageSelector.php(29): humhub\components\i18n\I18N->autosetLocale()
#28 /shared/httpd/humhub/htdocs/protected/vendor/yiisoft/yii2/base/Application.php(325): humhub\components\bootstrap\LanguageSelector->bootstrap(Object(humhub\components\Application))
#29 /shared/httpd/humhub/htdocs/protected/vendor/yiisoft/yii2/web/Application.php(69): yii\base\Application->bootstrap()
#30 /shared/httpd/humhub/htdocs/protected/humhub/components/Application.php(84): yii\web\Application->bootstrap()
#31 /shared/httpd/humhub/htdocs/protected/vendor/yiisoft/yii2/base/Application.php(271): humhub\components\Application->bootstrap()
#32 /shared/httpd/humhub/htdocs/protected/humhub/components/Application.php(65): yii\base\Application->init()
#33 /shared/httpd/humhub/htdocs/protected/vendor/yiisoft/yii2/base/BaseObject.php(109): humhub\components\Application->init()
#34 /shared/httpd/humhub/htdocs/protected/vendor/yiisoft/yii2/base/Application.php(204): yii\base
\BaseObject->__construct(Array)
#35 /shared/httpd/humhub/htdocs/protected/humhub/components/Application.php(49): yii\base\Application->__construct(Array)
#36 /shared/httpd/humhub/htdocs/index.php(28): humhub\components\Application->__construct(Array)
#37 {main}
```

# Web without debug

## Invalid database configuration

**Database not found.**

luke- (approved):
@martin-rueegg Looks very good. Thank you!

I've been thinking about introducing a `humhub\helpers` namespace for a while and this could be worth to be added into  `DatabaseHelper` or `ErrorHelper` there for example.

Should we do this with this PR? 

I can also do this afterwards...

martin-rueegg:
In my point of view, `humhub\lib` and `humhub\helpers` are kinda same, same but different ... :-)

However, I'm happy start the thing. I would, however, not touch any other of the `lib\*Helper` classes. For two reasons:

- I'd urge to _move_ the current files/classes to the new namespace, commit that, than create a new class at the old place extending from the new one. This for backwards-compatibility. However, if that is done in a separate commit, then the move of the class is recoreded in git. If the move and the re-creation of the old are in the same commit, then the tracking is lost.
- Also, moving those classes will affect tons of files (billions and billions ;-) ) and therefore should not be in this one PR for reasons of oversight.

If we create a `humhub\helpers\DatabaseHelper` class, I'd suggest to not only put the error handling there, but also the `isDatabaseInstalled()` method from the `BasicSettingsManager` which is currently abused.

It could also make sense to have the `isDatabaseInstalled()` (and potentially `isDatabaseAvailable()`) in `DatabaseHelper`, while the actual handling of the error in `ErrorHelper` ...

How do you want it to be or shall I make a suggestion along these lines?

luke-:
@martin-rueegg 

> However, I'm happy start the thing. I would, however, not touch any other of the `lib\*Helper` classes. For two reasons:
> 
> * I'd urge to _move_ the current files/classes to the new namespace, commit that, than create a new class at the old place extending from the new one. This for backwards-compatibility. However, if that is done in a separate commit, then the move of the class is recoreded in git. If the move and the re-creation of the old are in the same commit, then the tracking is lost.

Good point + also the hint regarding the GIT tracking. 

> * Also, moving those classes will affect tons of files (billions and billions ;-) ) and therefore should not be in this one PR for reasons of oversight.

Migration from `humhub\libs` to `humhub\helpers` should be done gradually. This is not meant to be a point for now. This PR should only be the first file in the `helpers` namespace.

> If we create a `humhub\helpers\DatabaseHelper` class, I'd suggest to not only put the error handling there, but also the `isDatabaseInstalled()` method from the `BasicSettingsManager` which is currently abused.
> It could also make sense to have the `isDatabaseInstalled()` (and potentially `isDatabaseAvailable()`) in `DatabaseHelper`, while the actual handling of the error in `ErrorHelper` ...

I would prefer: we finish this PR. One more change should be to move the new helper method to its own file in the `humhub\helpers` namespace.

Further improvements e.g. the suggested `isDatabaseInstalled()` can be then done in a seperate PR. But I like the idea.

protected/yii
```diff
@@ -25,6 +27,11 @@ $config = yii\helpers\ArrayHelper::merge(
...
 );
 
-$application = new humhub\components\console\Application($config);
-$exitCode = $application->run();
-exit($exitCode);
+try {
+    $exitCode = (new humhub\components\console\Application($config))->run();
+    exit($exitCode);
+} catch (\Throwable $ex) {
+    if (null === DatabaseHelper::handleConnectionErrors($ex)) {
```
luke-, line 34: @martin-rueegg A small note, maybe it's also worth to include this at CodeStyle docs, but mostly we use the order `$x === null` and not `null === $x`. Maybe something for future PRs :-)

luke- (approved):

protected/yii
martin-rueegg: I see. Well, it was a slightly different code at some point, then I've simplified it. It was more like a ...

```php
if ($test === $resultToBeUsed = $deep->nested->path->with->magic->access->getValue())  {}
```

... kinda thing. So in this scenario, the `$test`/`null` must be on the left-hand side of the test operator.

Alternatives would be:

```php
$resultToBeUsed = $deep->nested->path->with->magic->access->getValue();
if ($resultToBeUsed === null)  {}
```

or


```php
if (($resultToBeUsed = $deep->nested->path->with->magic->access->getValue()) === null)  {}
```

Both of which are more cluttered (the former) or less readable (the latter), in my perception.

Unfortunately, in humhum I see a lot of this, where the `$resultToBeUsed` is not stored:

```php
if ($deep->nested->path->with->magic->access->value === null)  {
   // do something
} else {
    $deep->nested->path->with->magic->access->value->read();
    $deep->nested->path->with->magic->access->value->write();
    $deep->nested->path->with->magic->access->value->andSomeMore();
}
```
Which is super inefficient, unless you expect that one object in the chain would have been replaced due to the actions taken.

Maybe something to consider in the code style, too. Accessing array entries, particularly if the array has many keys or the value is an array on its own, is expensive.

Also, accessing magic methods is, and the getter function should be used instead (unless it is an un[cached relation](https://github.com/humhub/humhub/blob/9d984e443b486a412247a8986d892885d1344b85/protected/humhub/components/FindInstanceTrait.php#L163) definition).

And such result should be cached within the function or block.

IMO. :-)

luke-: @martin-rueegg 
> 
> ```
> $resultToBeUsed = $deep->nested->path->with->magic->access->getValue();
> if ($resultToBeUsed === null)  {}
> ```
> 
I personally prefer this one, because it's much more readable.

> Unfortunately, in humhum I see a lot of this, where the `$resultToBeUsed` is not stored:
> 
> ```
> if ($deep->nested->path->with->magic->access->value === null)  {
>    // do something
> } else {
>     $deep->nested->path->with->magic->access->value->read();
>     $deep->nested->path->with->magic->access->value->write();
>     $deep->nested->path->with->magic->access->value->andSomeMore();
> }
> ```

Yes, such deep nesting is really bad, especially in your example the `value` is often used. But depends... 
 
> Maybe something to consider in the code style, too. Accessing array entries, particularly if the array has many keys or the value is an array on its own, is expensive.
> 
> Also, accessing magic methods is, and the getter function should be used instead (unless it is an un[cached relation](https://github.com/humhub/humhub/blob/9d984e443b486a412247a8986d892885d1344b85/protected/humhub/components/FindInstanceTrait.php#L163) definition).
> 
> And such result should be cached within the function or block.
> 
> IMO. :-)

I need to think about the caching topic a bit. I understand that there is really room for improvement here, but I also want to stay as close as possible to the Yii2 framework concepts and also not add too much complexity.

martin-rueegg:
@luke- 

> > * I'd urge to _move_ the current files/classes to the new namespace, commit that, than create a new class at the old place extending from the new one. This for backwards-compatibility. However, if that is done in a separate commit, then the move of the class is recoreded in git. If the move and the re-creation of the old are in the same commit, then the tracking is lost.
> 
> Good point + also the hint regarding the GIT tracking.

Yes, it is a underestimated or unknown limitation of GIT, that it does not track actual file moves. All it does, when committing (at the end, be it a stashed commit) see, if any newly created file **resembles** any deleted file. That means, that if you move a file _and_ change it at the same time too much, GIT will commit a delete and create, rather than a move. An example of this is the move of `protected/humhub/config/__autocomplete.php` [here](https://github.com/humhub/humhub/commit/9fabc0e2d40b580527a57eb32f3a4a29903ca3f2), which was previously `protected/humhub/Yii.php` but the tracking broke in that (stashed) commit.

Hence, best thing is:
1. git mv file
2. git commit with message
3. edit file
4. potentially add new file at the old location (e.g. class for backwards-compatibility extending the class at the new location)
5. git commit with message
6. And do NOT stash these commits later! :-)

luke-:
@martin-rueegg Especially point 6 is important, because I usually always squash merge PRs.  :-)

PR #6516, opened 2023-08-16
Fix humhub test case would fail on skipped tests

martin-rueegg:
When a test is annotated with `@skip` (e.g. [`testCreatePublicContentOnPrivateSpace`](https://github.com/humhub/humhub/blob/e7c61781786a6184508681984f4f6e47bac99d9c/protected/humhub/modules/content/tests/codeception/unit/models/ContentVisibilityTest.php#L114) without adding a reason, or [`testConvertTableWithoutBodyAtEnd`](https://github.com/humhub/humhub/blob/e7c61781786a6184508681984f4f6e47bac99d9c/protected/humhub/modules/content/tests/codeception/unit/widgets/RichTextHtmlConverterTest.php#L629)) the Application is not started (`Yii::$app` will be empty), yet the test itself is not skipped (the `@skip` annotation is a [feature of Codeception](url), but [not of PhpUnit](https://github.com/sebastianbergmann/phpunit/issues/3220)).

Hence, tests which are intended to be skipped can (and most likely will) cause the test to [fail](https://github.com/humhub/humhub/actions/runs/5872676823/job/15924640805?pr=6430#step:18:1456).

This PR checks for the `@skip` annotation when a test fails and the application has not been instantiated. In that case, it will throw a propper `SkippedTestError`.

- Bugfix

- No

- [x] Changelog was modified

### DIFF
--- a/CHANGELOG-DEV.md
+++ b/CHANGELOG-DEV.md
@@ -3,6 +3,7 @@ HumHub Changelog
 
 1.15.0-beta.2 (Unreleased)
 --------------------------
+- Fix #6516: Humhub test case would fail on skipped tests
 - Enh #6478: Add pseudo test class to allow population of DB with standard test data
 - Enh #6480: Convert assert* and db* methods to static, in line with general usage pattern
 - Enh #6505: Introduce Application interface; now also fire the `onInit` event when the web application has initialized
--- a/protected/humhub/tests/codeception/_support/HumHubDbTestCase.php
+++ b/protected/humhub/tests/codeception/_support/HumHubDbTestCase.php
@@ -28,6 +28,7 @@ use humhub\modules\user\tests\codeception\fixtures\GroupPermissionFixture;
 use humhub\modules\user\tests\codeception\fixtures\UserFullFixture;
 use humhub\tests\codeception\fixtures\SettingFixture;
 use humhub\tests\codeception\fixtures\UrlOembedFixture;
+use PHPUnit\Framework\SkippedTestError;
 use TypeError;
 use Yii;
 use yii\db\ActiveRecord;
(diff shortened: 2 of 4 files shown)


martin-rueegg:
Number of skipped tests remained the same with this patch:
- 5 before: https://github.com/humhub/humhub/actions/runs/5859928331/job/15886940747#step:18:1432
- 5 after: https://github.com/humhub/humhub/actions/runs/5878465145/job/15940607104#step:18:1432

martin-rueegg:
I know, this should probably be fixed upstream, but I do not have the resources to go into this.

Also, I do not know why this only hit us while running tests in #6430 and not already.

martin-rueegg:
@luke-  Could I maybe draw your attention to this one, too?

The main magic is happening here: it is checking if the `Yii::$app` is empty (which happens when a test is tagged with a `@skip` annotation), then looks for that annotation and returns it's text in the `SkippedTestError`.

https://github.com/humhub/humhub/blob/6e2be1175fd851f9c16871eb7c0cc59e4169b617/protected/humhub/tests/codeception/_support/HumHubDbTestCase.php#L60-L70

The rest of the code change is a merger of common code of `HumHubDbTestCase.php` and `HumHubHelper.php` with the addition of some debugging note and making sure the main cache is also deleted.

martin-rueegg:
Ah, and calling `parent::setUp();` at the end, rather than the beginning of `HumHubDbTestCase::setUp();` helps avoiding some error with the mails, which is otherwise not correctly instantiated, given that the `Yii::$app` might be empty and the settings are still the one from the previous test ....

luke-:
@martin-rueegg I didn't look at it 100% in detail, but looks good for me, thanks!
Can we also merge this into v.16 `next`?

PR #6521, opened 2023-08-17
fixup! Enh: Add Application interface and make sure the `onInit` even…

martin-rueegg:
See
- https://github.com/humhub/humhub/pull/6519#issuecomment-1682536154
- https://github.com/humhub/humhub/pull/6505

### DIFF
--- a/protected/humhub/components/Application.php
+++ b/protected/humhub/components/Application.php
@@ -8,15 +8,18 @@
 
 namespace humhub\components;
 
+use humhub\interfaces\ApplicationInterface;
 use Yii;
-use yii\helpers\Url;
-use yii\base\Exception;
+use Exception;
 
 /**
+ * Description of Application
+ *
  * @inheritdoc
  */
-class Application extends \yii\web\Application implements \humhub\interfaces\Application
+class Application extends \yii\web\Application implements ApplicationInterface
 {
+    use ApplicationTrait;
 
     /**
      * @inheritdoc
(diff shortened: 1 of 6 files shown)


luke- (approved):

martin-rueegg:
@luke- I've double tested it locally and can't reproduce the error. Both the test suite as well as when I do it manually in the browser it works. Using PHP v8.2 ....

I guess it could be a timing issue, that the post has not yet been replaced with the non-draft post. I've added a wait of 2 seconds into the test ...

luke- (approved):

martin-rueegg:
Thanks. Would you mind updating `statable` branch at some point, too?

luke-:
Justed merged `develop` into `statable` or is it better to merge `next` into it?

martin-rueegg:
> Justed merged `develop` into `statable` 

Thank you. 

> or is it better to merge `next` into it?

Thanks, appreciated! - Not until it is clear if (any version of) `FindInstance` will make it into develop ...

martin-rueegg:
@luke- Another thing that came to my mind is if you like the idea of the new namespace `humhub\interfaces`, as currently most interfaces are alongside some of the classes.

Now for a`ApplicationInterface` that would then probably be `humhub\components\ApplicationInterface`.

I personally prefer having the interfaces in their own namespace, as the do not always concord with the implementing classes, as we can see with `humhub\components\console\Application`. However, I just wanted to (consciously) run it by you.

luke-:
@martin-rueegg I like to have now `humhub\interfaces` in addition to newly introduced `humhub\helpers`. It will take some time to move core interfaces into it. I hope we can sometimes use rector for such things.

luke-:
@martin-rueegg I don't know if realistic, but cool would be e.g. a
`php protected/yii rector/migrate-module /path/to/module`
which automatically applies some changes from the `MIGRATE-DEV.md`.

martin-rueegg:
Haven't used or even looked into rector yet. But certainly something to keep in mind and that would be a perfect approach. `migrations` for the code!

PR #6527, opened 2023-08-19
Enable RuntimeCache during tests

martin-rueegg:
The RuntimeCache (RTC) was introduced in #6375.

However, while the RuntimeCache
- is enabled by default in the deployed config,
  https://github.com/humhub/humhub/blob/419a7f1d016b9ebbf18c5ff2a09af637741c9f0c/protected/humhub/config/common.php#L113-L116
- it is disabled for test config:
  https://github.com/humhub/humhub/blob/9a28225efd9e1e6464c261d7685866e84c3200d9/protected/humhub/tests/config/common.php#L24-L26

This makes no sense, as
- enabling the RTC tests is closer to real-life setting
- the `Yii::$app` is destroyed and recreated between every single test. And with it, the RTC. Hence, no cache flushing is required between tests (it already happens out of the box by instantiating a new cache anyway).

Hence, RuntimeCache should also be enabled during test.

**The fact that [this test is failing](https://github.com/humhub/humhub/actions/runs/5910637937/job/16032424516?pr=6527) is showing that the current RTC setting and implementation might cause issues in production. This should be addressed.**

I do not understand why they fail. Doing them live the everything works perfectly.

- Bugfix (Fix #6375)

- No. The failing tests seem to stem from the introduction of the RTC in #6375. 

- [ ] [All tests are passing](#issuecomment-new)
- [ ] Changelog was modified

### DIFF
--- a/protected/humhub/tests/codeception/_support/HumHubDbTestCase.php
+++ b/protected/humhub/tests/codeception/_support/HumHubDbTestCase.php
@@ -4,6 +4,7 @@ namespace tests\codeception\_support;
 
 use Codeception\Test\Unit;
 use humhub\components\behaviors\PolymorphicRelation;
+use humhub\helpers\RuntimeCacheHelper;
 use humhub\libs\BasePermission;
 use Codeception\Configuration;
 use Codeception\Exception\ModuleException;
@@ -79,6 +80,7 @@ class HumHubDbTestCase extends Unit
 
     protected function flushCache()
     {
+        RuntimeCacheHelper::flush();
         RichTextToShortTextConverter::flushCache();
         RichTextToHtmlConverter::flushCache();
         RichTextToPlainTextConverter::flushCache();
--- a/protected/humhub/tests/config/common.php
+++ b/protected/humhub/tests/config/common.php
@@ -21,9 +21,6 @@ return [
         'queue' => [
             'class' => 'humhub\modules\queue\driver\Instant',
         ],
-        'runtimeCache' => [
-            'class' => \yii\caching\DummyCache::class
-        ],
     ],
     'params' => [
         'installed' => true,


martin-rueegg:
I do not understand why [those tests fail](https://github.com/humhub/humhub/actions/runs/5910637937/job/16032424516?pr=6527). Doing them live the function works perfectly.

luke-:
@martin-rueegg 
> * [Enh: Introduce FindInstanceInterface with methods to get (cached) instances - Part 1 #6503 (comment)](https://github.com/humhub/humhub/pull/6503#issuecomment-1683667779)
> 
> I've found a way do enable cache dependencies while _not_ using a serializer ...

Do I understand correctly that if the serializer is disabled, `TagDependencies` will also work with the `ArrayCache`?
That would be really cool.

I see no benefit to serialize cache data as PHP array on runtime.

> 
> Also, since runtimecache is gone with the `Yii::$app` (which is re-created between every test), it should also be enabled during test.
> 
> The serializer can be removed, obviously, if TagDependency will not be implemented. in that case, the following setting should be used instead:
> 
> ```
>     'runtimeCache' => [
>         'class' => \yii\caching\ArrayCache::class,
>         'serializer' => false,
>     ],
> ```
> 
> or the setting removed altogether, as it is already defined in the production `common.php`.

I would prefer to specify this once in `protected/humhub/config/common.php`. Did you meant that by production common?

martin-rueegg:
@luke- 
> > * [Enh: Introduce FindInstanceInterface with methods to get (cached) instances - Part 1 #6503 (comment)](https://github.com/humhub/humhub/pull/6503#issuecomment-1683667779)
> > 
> > I've found a way do enable cache dependencies while _not_ using a serializer ...
> 
> Do I understand correctly that if the serializer is disabled, `TagDependencies` will also work with the `ArrayCache`? That would be really cool.

Kind of. `TagDependencies` only work if a serializer is used. Either the default PHP-internal serializer, or a user-provided. If serialization is disabled, they will not work.

Now since it would be unnecessary effort to serialize and unserialize runtime-cached items (which we all agree), serializing was disabled in `common.php`. This makes sense. However, the side effect of setting `['serializer' => false]` is, that `TagDependencies` are disabled.

My suggested solution is, that we provide a user-defined serializer (an array of a serializing and unserializing function) which in both cases just return the value as-is. As such, the cache implementation thinks it does serializing, while actually, it doesn't. The only thing here is, that this trick only works for `ArrayCache`, not e.g. `FileCache`. So who ever decides to use a different Cache implementation for the runtime cache, needs to be aware of that.

The example on how to enable the `TagDependencies` see 
- #6528
- https://github.com/humhub/humhub/blob/419a7f1d016b9ebbf18c5ff2a09af637741c9f0c/protected/humhub/config/common.php#L113-L116
- https://github.com/humhub/humhub/blob/419a7f1d016b9ebbf18c5ff2a09af637741c9f0c/protected/humhub/helpers/RuntimeCacheHelper.php#L352-L362

> I see no benefit to serialize cache data as PHP array on runtime.

Exactly. Unless you'd use APCu or Memcache or something.

**Now, enabling `TagDependencies` is not the main issue of this PR. This PR shows, that tests fail if runtime cache is enabled, which it should be, as argued below.**

> > Also, since runtimecache is gone with the `Yii::$app` (which is re-created between every test), it should also be enabled during test.
> > The serializer can be removed, obviously, if TagDependency will not be implemented. in that case, the following setting should be used instead:
> > ```
> >     'runtimeCache' => [
> >         'class' => \yii\caching\ArrayCache::class,
> >         'serializer' => false,
> >     ],
> > ```
> > or the setting removed altogether, as it is already defined in the production `common.php`.
> 
> I would prefer to specify this once in `protected/humhub/config/common.php`. Did you meant that by production common?

Well, it _is_ specified in `protected/humhub/config/common.php` and enabled there; currently with `ArrayCache` and ` 'serializer' => false`. For the tests, this is currently disabled in the test config by overwriting it with `DummyCache`, as you can see in this PR. I just removed this overwriting and as such have enabled the runtime cache in tests as well.

It makes sense to use the runtime cache in tests, as it is closer to real-life setting, since it is enabled there by default too. And as laid out above, the `Yii::$app` is destroyed and recreated between every single test. And with it, the runtime cache. Hence, no cache flushing is required between tests (it already happens out of the box).

The fact that this test is failing is showing that the current runtime cache setting and implementation might cause issues in production. This should be addressed.

luke-:
@martin-rueegg Thanks for the explanation, I think it is clear to me now.

IMO it is justifiable when someone changes the Runtime Cache implementation to care about such side effects. Maybe we can comment this behavior a bit more in the config.

Its also good to enable the Runtime cache in the tests as well. I'll take a look here, hjowever, I am quite busy the next few weeks (with some days off) so it will take some time to deal with this issue...

luke-:
@martin-rueegg I've enabled the tests here: https://github.com/humhub/humhub/pull/6582 and already merged into `next`.

PR #6529, opened 2023-08-21
Enh: Add boolean return type to *safe* methods in migrations

martin-rueegg:
In migrations, it would be helpful to know if the underlying `*safe*` method did actually do the thing, or not, without having to re-code the function.

This PR simply adds a `bool` return type. This, however, is not enforced for backwards-compatibility, as previously the return type was `void`. By removing the type declaration, child classes are still valid if the use `void`.

Also, for reasons of consistency and readability, the code in the methods have all been re-arranged, so that first is always done the action that is supposed to happen, and second the logging if not.

- Feature
- Code style update
- Grammar corrections in comments

- No

- [x] All tests are passing
- [x] Changelog was modified

### DIFF
--- a/CHANGELOG-DEV.md
+++ b/CHANGELOG-DEV.md
@@ -3,6 +3,7 @@ HumHub Changelog
 
 1.15.0-beta.2 (Unreleased)
 --------------------------
+- Enh #6529: Add boolean return-type to `*safe*` methods in migrations
 - Enh #6478: Add pseudo test class to allow population of DB with standard test data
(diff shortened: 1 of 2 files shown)


luke- (approved):

martin-rueegg:
@luke- could this go into 1.15, please. There is no side-effect of this at all.

If not, could you merge it into `next`?

martin-rueegg:
@luke- Appreciated. Thank you!

Issue #6515, opened 2023-08-16
Scheduled Post Permalink Leads to error 


ArchBlood:
### What steps will reproduce the problem?
- Create a scheduled Post
- Open the drop-down on the post and select permalink and go to the post

![Screenshot_20230816-073211.jpg](https://github.com/humhub/humhub/assets/35392110/aa39d0f0-7f82-4a52-910b-5f8a30eec276)

![Screenshot_20230816-073153.png](https://github.com/humhub/humhub/assets/35392110/dc7d63de-020d-44ac-95e9-7072952dd2c3)

### What is the expected result?
Should show the scheduled Post or a mention of the post not being available till a scheduled date.

### What do you get instead?
Oops something went wrong

### Additional info

| Q                | A
| ---------------- | ---
| HumHub version   | Community site 
| PHP version      | Community site 
| Operating system | Community site 

> Note: Also provide the full error stacks from `Administration --> Information --> Logging` and errors from your browser console if possible.
> Note: Please use the [community](https://community.humhub.com) site for any questions on customizing HumHub.

yurabakhtin:
@luke- Fixed in PR https://github.com/humhub/humhub/pull/6540.

yurabakhtin:
We should use the approach from PR https://github.com/humhub/humhub/pull/6005 instead of https://github.com/humhub/humhub/pull/6540.

ArchBlood:
Seems to also throw this error for notifications if you're following someone that schedules a post, this affects all versions of `1.15`.

yurabakhtin:
@ArchBlood Do you mean user recieves a notification about scheduled post?

If yes, I am not sure it is possible because here https://github.com/humhub/humhub/blob/master/protected/humhub/modules/content/models/Content.php#L277-L280 we restrict this:

```php
if (!$this->getStateService()->isPublished()) {
    // Don't notify about not published Content
    return;
}
```

Could you show what error you have and how to reproduce this if you know? Thanks.

ArchBlood:
> @ArchBlood Do you mean user recieves a notification about scheduled post?
> 
> If yes, I am not sure it is possible because here https://github.com/humhub/humhub/blob/master/protected/humhub/modules/content/models/Content.php#L277-L280 we restrict this:
> 
> ```php
> if (!$this->getStateService()->isPublished()) {
>     // Don't notify about not published Content
>     return;
> }
> ```
> 
> Could you show what error you have and how to reproduce this if you know? Thanks.

[...]

PR #6540, opened 2023-09-07
Fix scheduled post visibility on permalink

yurabakhtin:

- [x] Bugfix

- [x] All tests are passing
- [x] Changelog was modified

**Other information:**
Issue: https://github.com/humhub/humhub/issues/6515

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ HumHub Changelog
 - Fix #6507: Reload page after accepting of invite in Space header
 - Fix #6531: Fix active style for `btn-lg`
 - Fix #6534: Cannot change Homepage of space in PHP 8.x
+- Fix #6515: Fix scheduled post visibility on permalink
 
 1.14.3 (July 27, 2023)
 ----------------------
--- a/protected/humhub/modules/content/components/ActiveQueryContent.php
+++ b/protected/humhub/modules/content/components/ActiveQueryContent.php
@@ -117,7 +117,7 @@ class ActiveQueryContent extends ActiveQuery
         $this->stateFilterCondition[] = $conditionUser;
         $this->andWhere($this->stateFilterCondition);
 
-        $this->andWhere(['OR', $conditionSpace, $globalCondition]);
+        $this->andWhere(['OR', $conditionSpace, $conditionUser, $globalCondition]);
 
         return $this;
     }
--- a/protected/humhub/modules/content/components/ActiveQueryContent.php
+++ b/protected/humhub/modules/content/components/ActiveQueryContent.php
@@ -11,6 +11,7 @@ namespace humhub\modules\content\components;
 use humhub\modules\content\models\Content;
 use humhub\modules\content\models\ContentTag;
 use humhub\modules\content\models\ContentTagRelation;
+use humhub\modules\space\models\Membership;
 use humhub\modules\space\models\Space;
 use humhub\modules\user\helpers\AuthHelper;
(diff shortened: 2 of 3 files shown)


luke- (changes requested):
@yurabakhtin Does this affect also other content types?

ArchBlood:
> @yurabakhtin Does this affect also other content types?

I believe it affected all content that can be scheduled.

yurabakhtin:
> @yurabakhtin Does this affect also other content types?

@luke- No, I have done this only for the Post view action.

If we need to apply this for all contents then I think we should modify the default `StreamQuery->stateFilterCondition` (https://github.com/humhub/humhub/blob/master/protected/humhub/modules/stream/models/StreamQuery.php#L152) from:

- `['OR', ['content.state' => Content::STATE_PUBLISHED]]`

to:

- `['OR', ['content.state' => Content::STATE_PUBLISHED], ['content.created_by' => Yii::$app->user->id]]`

But I am not sure it is correct solution, because probably we should not display for example all deleted contents in stream when the filter "Deleted" is not selected.

Ok, I will review again how to apply the rule `['content.created_by' => Yii::$app->user->id]` for all content types but only for action "View content".

yurabakhtin:
@luke- Ok, I have decided to allow for creators to see content with any state.
Commit https://github.com/humhub/humhub/pull/6540/commits/7ce1f98e0907a52666dda0fa7e688df4ae5dcacd.

Really it would be better for all such view actions remove the filter `->readable()` and replace it with code:

```
if (!$record->content->canView()) {
    throw new ForbiddenHttpException();
}
```

so after this all not published contents will be available for all users(admin, managers) who can view the record and not only a creator.
But the modification requires to find all such actions in all modules(wiki, news, etc.).
I think better to keep the `->readable()` because it seems the `->canView()` is already integrated in the `->readable()` with `$this->andWhere("{$conditionSpace} OR {$conditionUser} OR {$globalCondition}");` and I am thinking we just should append here the `$this->stateFilterCondition`, so a Content will be visible when `'content.state' => Content::STATE_PUBLISHED` or if user has at least one permission related to the `{$conditionSpace} OR {$conditionUser} OR {$globalCondition}`, I will test this case, so the PR is not ready for merging yet.

yurabakhtin:
@luke- Commit https://github.com/humhub/humhub/pull/6540/commits/16f9314cbaca836b717be6e7f9b30ceb2a6552de - I think this solution is better when we allow to view Content with any State if current User has a permission to view it.

The code:

```php
$this->stateFilterCondition[] = $conditionUser;
$this->andWhere($this->stateFilterCondition);
```

is equivalent of:

```
content.state = 1
OR content.created_by = 1
OR (cuser.id IS NOT NULL
    AND (content.visibility = 1
        OR (content.visibility = 0 AND content.contentcontainer_id = 1)
        OR (content.visibility = 0 AND cff.id IS NOT NULL)
    )
)
```

luke- (approved):

luke-:
@yurabakhtin Here it would be good to provide some unit tests for it.

Do I understand correctly that before your change `readable()` returned only `STATE_PUBLISHED` content and now it returns all states?

yurabakhtin:
> Do I understand correctly that before your change `readable()` returned only `STATE_PUBLISHED` content and now it returns all states?

@luke- Yes, all states if current user can view a Content.

> Here it would be good to provide some unit tests for it.

Ok, I will create it.

luke-:
https://github.com/humhub/humhub/pull/6005

PR #6548, opened 2023-09-10
Fix #6547 Invalid class name validated as valid and improve validation

### LINKED ISSUE #6547: Bug: Invalid class name validated as valid
[`\humhub\libs\Helpers::CheckClassType()`](https://github.com/humhub/humhub/blob/master/protected/humhub/libs/Helpers.php#L186) checks if a class name exists and is of a specific type and throws an Exception, if it doesn't.

In doing so, it removes invalid characters from the `$className` parameter:
https://github.com/humhub/humhub/blob/db3bf270d942b97c62b6a473600523c7f891b953/protected/humhub/libs/Helpers.php#L188

As a consequence, if a valid class name including invalid characters is provided, the function returns a positive result - as the name has been cleaned up. Yet the calling code doesn't know about the invalid characters in the class name and using the provided name to instantiate an object would result in an error.

The function should either
- return false/throw an Exception on a `$className` containing invalid characters, or
- return the cleaned `$className`

Also, if `$type` is an array, the existence of `$className` is checked multiple times.

https://github.com/humhub/humhub/blob/db3bf270d942b97c62b6a473600523c7f891b953/protected/humhub/libs/Helpers.php#L186-L203

martin-rueegg:
Besides fixing #6547, this version of `\humhub\libs\Helpers::checkClassType()` allows
- providing an object instance, rather than a class name, to be tested
- allows to return a negative result, rather than throwing an exception
- allows to distinguish between an invalid parameter `$className` and an invalid parameter `$type` (different code in Exception)

Depends and includes
- #6566

- Bugfix (Fix #6547)
- Feature
- Code style update

- Yes

The current version of the function throws a `\yii\base\Exception` if the type is not met.
The new version of the function throws a `\humhub\exceptions\InvalidArgumentException`, which derives from `\yii\base\InvalidArgumentException` and `\yii\base\InvalidParamException`, but not `\yii\base\Exception`.

As such, if the calling code is handling the exception (which is not the case anywhere in the core code) with the specific exception class `\yii\base\Exception` (and not the more generic `\Exception` or even `\Throwable`), then the `catch` clause needs to be updated/extended.

**Note:** the [change in the method name](https://github.com/humhub/humhub/compare/develop...metaworx:humhub:enh/improve-class-type-validation?expand=1#diff-764f074e8da6eac0b7a0597550ed7fa54638193a0f93d21181ed8ef5c3a18fc4R189) (case change) **does not constitute** a breaking change, as method names are case-insensitive. Nonetheless, the core code has been updated accordingly.

- [x] Changelog was modified

### DIFF
--- a/protected/humhub/libs/Helpers.php
+++ b/protected/humhub/libs/Helpers.php
@@ -8,9 +8,11 @@
 
 namespace humhub\libs;
 
+use humhub\exceptions\InvalidArgumentClassException;
+use humhub\exceptions\InvalidArgumentTypeException;
 use humhub\exceptions\InvalidArgumentValueException;
 use Yii;
-use yii\base\Exception;
+use yii\base\InvalidArgumentException;
 
 /**
  * This class contains a lot of html helpers for the views
@@ -19,6 +21,15 @@ use yii\base\Exception;
  */
 class Helpers
 {
+    public const CLASS_CHECK_INVALID_CLASSNAME_PARAMETER = 1;
+    public const CLASS_CHECK_INVALID_TYPE_PARAMETER = 2;
+    public const CLASS_CHECK_VALUE_IS_EMPTY = 4;
+    public const CLASS_CHECK_INVALID_TYPE = 8;
+    public const CLASS_CHECK_NON_EXISTING_CLASS = 16;
+    public const CLASS_CHECK_TYPE_NOT_IN_LIST = 32;
+    public const CLASS_CHECK_VALUE_IS_INSTANCE = 64;
+    public const CLASS_CHECK_VALUE_IS_NULL = 128;
+
     /**
      * Shorten a text string
      *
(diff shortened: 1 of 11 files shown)


protected/humhub/exceptions/InvalidArgumentException.php
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+namespace humhub\exceptions;
+
+use yii\base\InvalidArgumentException as BaseInvalidArgumentException;
+
+/**
+ * @since 1.15
+ */
+class InvalidArgumentException extends BaseInvalidArgumentException
```
luke-, line 16: @martin-rueegg I'm not sure, if I'm a big fan of the new exception classes. 

e.g. so far in the IDE class picker, I simply searched for a suitable exception class type and then `throw new InvalidArgumentException($message = 'Invalid user given!')`. When this was thrown, you see a stack trace.

Now there are own Humhub "InvalidArgumentException" classes, instead of the well known`InvalidArgumentException(message)` style,  with new parameters `InvalidArgumentException(__METHOD__, parameter, valid, given, ...)`. For me and other devs, now a documentation is probably required.

I understand there are some benefits here, but for me it's also a bit over engineered.

To sum up, big cons for me:

- Complexity (Parameters & co). 
- Different Constructor Signature between `yii\base\InvalidArgumentException`  and `humhub\exceptions\InvalidArgumentException` 
- When accepted, we should use this more...

protected/humhub/exceptions/InvalidTypeExceptionTrait.php
```diff
@@ -13,56 +13,23 @@
...
-        $nullable = false,
-        $code = 0,
-        $previous = null
-    ) {
-
-        $this->methodName = $method;
-        $this->parameter = $parameter;
-        $this->validType = (array)($validType ?? ['mixed']);
-        $this->givenValue = $givenValue;
+    use InvalidArgumentExceptionTrait;
```
luke-, line 16: @martin-rueegg I also see in your other PR that you use traits very intensively. So far, we have done this only in rare cases.

Is it really necessary to this extent? Where are the advantages here? I'm a bit worried that it complicates things again if it's spread over multiple class/traits.

protected/humhub/libs/Helpers.php
```diff
@@ -179,27 +180,61 @@ public static function GetUniqeId()
...
-     * @return boolean
+     * @param string|object $className Object or classname to be checked
+     * @param string|array $type
+     * @param bool $throw
+     * @param string|null $matchingClass
+     *
+     * @return string|null
      */
-    public static function CheckClassType($className, $type = '')
+    public static function checkClassType($className, $type, bool $throw = true, ?string &$matchingClass = null): ?string
```
luke-, line 190: Can you please add more details of the usage of the `$matchingClass` parameter?

luke- (changes requested):
@martin-rueegg Thanks for the PR. Please see my remarks...

protected/humhub/exceptions/InvalidArgumentException.php
martin-rueegg: Thank you for your review, @luke-, and your comments.

I appreciate it takes some time to craft the new `InvalidArgumentException(__METHOD__, parameter, valid, given, ...)`.

However, this is a one-time effort and once you're used to the parameters (which appart from `$parameter` and `$valid` seem to be quite self-explanatory to me), it's not costing much more time than writing `InvalidArgumentException('Invalid user given!')`. And for me the benefit are 
- When the error is thrown, I see what value was received and why it was the wrong $value or value type
- Without this, I have to dig through the code of the called function/method to understand which values/types are allowed (doc comments unfortuanately are not always up-to-date and are only visible in the source code).
- The source of the value is often many calls earlier. There can be easily 10-20 function calls in between and the stack trace usually shows the defining class, not the class of the respective instance, and hence I find it really time-consuming to track down the source code and where it came from. In my experience, giving a more verbose output in the error message helps with this.
- Also, the "invalid user" example may be too simple, and for such a use-case, it may be more suitable to use the `yii\base\InvalidArgumentException`. However, if it is a `state` (see new status interface and implementation) then it makes sense to see what states are available. In this case, this exception lists the received value along with the array of allowed values (which is automatically retrieved from the defining `StateService`). Or if one is validating a `$config` parameter, providing the list of allowed parameters in the Exception message can e.g. show a typo immediately, without me going to search the source code
- So the main advantage for me is, that I have a little more effort to write the Exception message, but often save time when debugging the cause. Which, as opposed to the former, the latter happens many times for a given method.
- `InvalidArgumentException` is extended by `InvalidArgumentTypeException`, which has previously been accepted.
- the `$method` parameter could be left away and be replaced by a trace inspection. However, sometimes it's helpful to pass the name of the method that has the restriction to a sub-method that actually does (part of) the verification. But again, it's a small habit to just type `_`, `_`, `M`, `tab`, `,` and go on ... (PhpStorm also provides facilites to store snippets. It may be a useful snippet for a shared PhpStorm-Config-Repository.)
- Happy to extend some documentation in the Exception class.

protected/humhub/exceptions/InvalidTypeExceptionTrait.php
martin-rueegg: Traits have the following advantages
- Reduce code redundancy
- Allow classes of different inheritance to behave exactly the same, e.g. when implementing an interface
- have been introduced to PHP v5.4.0 and as such should be known by a common PHP developer and are also widely supported by IDEs

In contrast, **Yii2 behaviors** seem much more complex to me and 
  - are only supported by some IDEs [with plugin]) and
  - require more configuration and
  - more memory, and
  - are only available to components
  - The advantage of behaviors is that
    -  they can be sneaked into common objects by altering the container's definition and as such
    - can be provided by extensions (traits could be provided by extensions, too, though).
    - Also, they allow to load code only if really required, which for me is the biggest advantage, unless the behavior is called on nearly every invocation and every object it implements it. For example, the polymorphic relation is used extensively and should, IMHO, better be used as a trait which would reduce memory and function calls.

The main disadvantage of traits for me are
- no support for constants before php v8.2
- defined variables must match type and initial value of any implementing or extending class
- you cannot just fallback on `parent::method()` but need to use either two different method names or manually rename the trait's method on import. However, for behaviors, there are also caveats in this scenario. Renaming on import is not a big deal. What annoys me most here, is, that if you ctrl-click on the alias, it takes you to the import statement and there you have to ctrl-click again on the original method name. And going back takes the two steps as well. 

Now, whether or not they are necessary? No, absolutely not. No trait can do what you could not do with copy-and-past (and potentially implementing interfaces to test the instance for a given capability).

Of course, the disadvantages of code-duplication is well known: more maintenance and eventually divergence.

What I find really cool with traits though, is when defining an Interface, e.g. `ArrayAccessInterface`, to also create a trait alongside it, i.e. `ArrayAccessTrait`. Now a class can implement the interface by adding 1.5 lines of code:

- `implements ArrayAccessInterface`
- `use ArrayAccessTrait`

Done.

The proposed StateService implementation uses this to implement Editable, Deletable, Draftable and other states. See https://github.com/humhub/humhub/pull/6430/files#diff-dd4d14a7d1bda6fa1dc4821d285fcbb1e336ee7dccd43951b8a6443b43569775R1

If this would have to be maintained in all the tables that will use the state field, this would be a lot of duplicate code to maintain. Also, if a new test for one state is required, this can be added to the trait and you're set.

By the way, even Yii2 itself is using exactly this design pattern in it's key components and on different levels (see `Query` and `ActiveQuery`):
- `\yii\base\Model`
  ```php 
  class Model extends Component implements StaticInstanceInterface, IteratorAggregate, ArrayAccess, Arrayable
  {
    use ArrayableTrait;
    use StaticInstanceTrait;
  ```
- `\yii\db\ActiveQuery`
  ```php 
  class ActiveQuery extends Query implements ActiveQueryInterface
  {
    use ActiveQueryTrait;
    use ActiveRelationTrait;
  ```
- `\yii\db\Query`
  ```php 
  class Query extends Component implements QueryInterface, ExpressionInterface
  {
    use QueryTrait;
  ```

Now for this current example of the Exception, I don't think an Interface is required.

Happy to discuss further. But eventually it's up to you to decide, if traits are eligible or in which particular cases.

protected/humhub/libs/Helpers.php
martin-rueegg: Happy to do so. Hoever, I'm not currently using this parameter anywhere. So I may as well remove it and we re-add it when reqired. It was more a spontaneous idea, since the variable already existed in the code of the function.

It's a bit like the `$matches` parameter of `preg_match`. I thought it might be useful to know straight away which class is implemented. So one could do somethign along the lines of (a fabricated example!)

```php
$className = checkClassType($className , [ContentAddonActiveRecord::class, ContentActiveRecord::class], true, $matchingClass);

switch($matchingClass) {
    case ContentAddonActiveRecord::class:
            /** @var ContentAddonActiveRecord $object */
            return $object->canEdit($userId) || $object->content->canEdit($userId);

    case ContentActiveRecord::class:
            /** @var ContentActiveRecord $object */
            return $object->content->canEdit($userId);
}
```

rather than

```php
$className = checkClassType($className , [ContentAddonActiveRecord::class, ContentActiveRecord::class]);

if (is_a($className, ContentAddonActiveRecord::class)) {
	/** @var ContentAddonActiveRecord $object */
	return (new $className())->canEdit($userId);
}

if (is_a($className, ContentActiveRecord::class)) {
	/** @var ContentActiveRecord $object */
	return (new $className())->content->canEdit($userId);
}
```

protected/humhub/exceptions/InvalidArgumentException.php
luke-: I understand that such exception are really handy too. 

Here is what I would prefer:

- Rename the `InvalidArgumentException` to `InvalidArgumentValueException`. To reflect it's difference between `yii\base\InvalidArgumentException`.

- Can we simplify them? For me, this is way too unintuitive: `throw new InvalidArgumentException(__METHOD__, [2 => '$type'], ['string', 'string[]'], $type, false, 1);`

What about just:
	- `InvalidArgumentValueException(string $argumentName, ?$validValues, $?givenValue);`
	- `InvalidArgumentTypeException($argumentName, ?$validTypes, ?$givenType)`
	
e.g. `throw new InvalidArgumentValueException('type', ['string', 'string[]'], $type);`

- Remove Traits when possible

protected/humhub/libs/Helpers.php
luke-: Ok, I see, that's an elegant solution.

If there is some example usage in the core, you are welcome to leave it in.

martin-rueegg: > ... However, I'm not currently using this parameter anywhere. (...) It was more a spontaneous idea, since the variable already existed in the code of the function.

No, as mentioned above, I'm not currently using it. So I'll remove the parameter for now and anyone who might use it can re-add the parameter. 👍

protected/humhub/exceptions/InvalidTypeExceptionTrait.php
martin-rueegg: You're welcome!

If I may add: I would assume that if Interfaces (and potentially Traits) would have been used more often in the code, the many different implementations of the `can*` methods would have been a bit more standardized. Some models use `canRead()`, others use `canView()`, although it's essentially the same. Some allow `User` as first parameter, others only `user_id`or vice versa.

The `FindInstanceInterface` (#6503) together with the `EditableInterface` and friends (#6451) are an attempt to streamline this.

martin-rueegg:
@luke-

I've 
- renamed [`InvalidArgumentException` to `InvalidArgumentValueException`](https://github.com/humhub/humhub/pull/6548/commits/7536be93a6d2d709474749e374cf8ed0159bbe00)
- removed the [unused parameter `$matchingClass` from `\humhub\libs\Helpers::checkClassType`](https://github.com/humhub/humhub/pull/6548/commits/ab80b607dbdbceecc2adea23df4d9709d9260fe1), as requested.

As mentioned in the [comment](https://github.com/humhub/humhub/pull/6548#discussion_r1325584702) above, I'd rather do the changes regarding the Exceptions and Traits in a separate PR. I'll do it straight away. Would that be ok? I'd build on top of this one, so that the entire code referring to those Exceptions is updated in one single PR ....

luke-:
@martin-rueegg Yes, thank you. Should I merge this?

protected/humhub/exceptions/InvalidArgumentException.php
martin-rueegg: @luke- I've moved the amendment of the Exceptions to
- #6566
And may have found a good compromise in regards to compatibility with `yii\base\InvalidArgumentException`.

martin-rueegg:
> Yes, thank you. Should I merge this?

@luke- No, in order to avoid invalid code, I have now rebased this on top of 
- #6566

luke-:
@martin-rueegg Ok, just let me know...

martin-rueegg:
@luke- 

Ok, this would be ready now, too. I've just rebased it on `develop`.

luke- (approved):

PR #6530, opened 2023-08-22
Enh: perfomance improvements

martin-rueegg:
### What kind of change does this PR introduce?

- Refactor

### Does this PR introduce a breaking change?

- No

**The PR fulfills these requirements:**

- [x] It's submitted to the `develop` branch, _not_ the `master` branch if no hotfix
- [ ] When resolving a specific issue, it's referenced in the PR's description (e.g. `Fix #xxx[,#xxx]`, where "xxx" is the Github issue number)
- [x] All tests are passing
- [ ] New/updated tests are included
- [x] Changelog was modified

Issue #4653, opened 2020-11-23
Module::update() is not called after module update with the marketplace


marc-farre:
### What steps will reproduce the problem?

In the `Module.php` file, create this class: `class Module extends \humhub\components\Module`
In this class, create this function:

    /**
     * @inheritdoc
     */
    public function update()
    {
        parent::update();
        
        // Code to execute here after marketplace update
    }

### What is the expected result?

After module update with the Marketplace, `// Code to execute here after marketplace update` should have been executed.

### What do you get instead?

Code is note exectuded.
Perhaps this code should be in `beforeUpdate()` ?
Or do we need to create an event to have a code executed after a module update ? The `EVENT_AFTER_UPDATE` of the marketplace module ?

### Additional info

No error in the log.
Thanks for your help !

| Q                | A
| ---------------- | ---
| HumHub version   | 1.7

luke-:
Probably related to: https://github.com/humhub/humhub/issues/3872

marc-farre:
Issue moved to https://github.com/humhub/humhub/issues/6744

Issue #4364, opened 2020-09-11
Module update throws error if event handler requires removed class


buddh4:
Similar to https://github.com/humhub/humhub/issues/3872 old event handler instances seem to be called once a module
is updated through the marketplace.

The following error was thrown after renaming the `Notifications` widget in an update.

Error:
```
Error: Class 'humhub\modules\mail\widgets\Notifications' not found in /srv/www/htdocs/xxx/modules/mail/Events.php:84
```

Info:
```
INFO  Fri, 11 Sep 2020 17:14:48 +0200 application$_GET = [] $_SERVER = [ 'REDIRECT_STATUS' => '200' 'HTTP_X_REAL_IP' => 'xxx' 'HTTP_X_FORWARDED_FOR' => 'xxx' 'HTTP_X_FORWARDED_PROTO' => 'https' 'HTTP_HOST' => 'xxx.com' 'HTTP_X_FORWARDED_HOST' => 'xxx.com' 'HTTP_CONNECTION' => 'close' 'HTTP_CACHE_CONTROL' => 'max-age=0' 'HTTP_ACCEPT' => 'text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.9' 'HTTP_UPGRADE_INSECURE_REQUESTS' => '1' 'HTTP_USER_AGENT' => 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/85.0.4183.102 Safari/537.36' 'HTTP_SEC_FETCH_SITE' => 'same-origin' 'HTTP_SEC_FETCH_MODE' => 'same-origin' 'HTTP_SEC_FETCH_DEST' => 'empty' 'HTTP_REFERER' => 'https://xxx.com/marketplace/update?betaSwitch=1
````

| Q                | A
| ---------------- | ---
| HumHub version   | 1.6
| PHP version      | 
| Operating system |

PR #6550, opened 2023-09-10
Enh improve module migrations

### LINKED ISSUE #6237: Modules: Migration errors during module activation are ignored

Activate a module that creates an error during migration, or `Migration::up()` returns `false`, hence the migration was _not_ successful.

In my opinion, the module should not be activated, all migrations should be rolled back, and the `Module::disable()` method should be called. Maybe also a `Module::enableError()` could be called, defaulting to `Module::disable()`?

At least some error or warnings should be showed, so that the log files are inspected.

### What do you get instead?

Module gets activated with no indication whatsoever that something didn't go, as it should have.

### Additional info

| Q                | A
| ---------------- | ---
| HumHub version   | 1.13.2
| PHP version      | 8.1
| Operating system | devilbox docker container

martin-rueegg:
Currently, any error that happens during the migration when enabling, updating, or disabling a module, go unnoticed. There is no exception thrown, nor a log entry added nor is the migration result nor its output available to any calling code.

With this PR does
- `enable,` `update()`, or `disable()` methods return
  - `true` on success
  - `false` on error
  - `null` if the `enable,` `update()`, or `disable()` method was cancelled due to the `before*` method.
- migration output is logged, if available
- `$module->migrationResult` holds the result of the migration action:
  - `ExitCode::OK = 0` for success, 
  - `ExitCode::UNSPECIFIED_ERROR = 1` for error, or
  - `null` if no migration was executed
- `$module->migrationOutput` holds the output of the migration. Can also be `null` on error or if no migration was executed.
- new methods `beforeEnable()`/`afterEnable()` and `beforeDisable()`/`afterDisable()` are added, corresponding to the `beforeUpdate()`/`afterUpdate()` methods
- new methods `beforeMigrate()`/`afterMigrate()` are also added
- return types of all the methods involved are now specified
- **A module is _not enabled_,** if the migration during _enabling_ fails
- **A module is _disabled_,** if the migration during an _update_ fails

As such, this PR does not only log errors and migration results by default, but also allows module developers more fine-grained control in the process of enabling, updating and disabling their module.

By disabling the module upon an error during enabling or updating ensures that the error is discovered immediately, rather than later down the road, e.g. when a missing column is accessed.

As such it improves the over-all stability of the software by avoiding a faulty state of database/modules.

It should be noted though, that the entire process of enabling and disabling should, in my humble opinion, be encapsulated within a database transaction. This PR, however, does not address this. Therefore, it is still possible that the migration results in a undefined state. But at least it will be noticed due to the disabling/logging.

- Bugfix
  - Fixes #6237 
- Feature
- Refactor (type restrictions)

- Yes

Return type restrictions on some methods of 

- [x] It's submitted to the `next` branch
- [x] Changelog was modified

### Other information

Depends on/includes
- #6530
- #6551
- #6553
- #6711

### DIFF
--- a/protected/humhub/commands/MigrateController.php
+++ b/protected/humhub/commands/MigrateController.php
@@ -10,8 +10,8 @@ namespace humhub\commands;
 
 use humhub\components\Module;
 use humhub\helpers\DatabaseHelper;
+use humhub\services\MigrationService;
 use Yii;
-use yii\base\InvalidRouteException;
 use yii\console\Exception;
 use yii\db\MigrationInterface;
 use yii\web\Application;
@@ -207,20 +207,14 @@ class MigrateController extends \yii\console\controllers\MigrateController
      * @param \yii\base\Module|null $module Module to get the migrations from, or Null for Application
      *
      * @return string output
-     * @throws Exception
-     * @throws InvalidRouteException
+     * @deprecated since 1.16; use MigrationService::migrateUp()
+     * @see MigrationService::migrateUp()
      */
     public static function webMigrateAll(string $action = 'up', ?\yii\base\Module $module = null): string
     {
-        ob_start();
-        $controller = new self('migrate', $module ?? Yii::$app);
-        $controller->db = Yii::$app->db;
-        $controller->interactive = false;
-        $controller->includeModuleMigrations = true;
-        $controller->color = false;
-        $controller->runAction($action);
-
-        return ob_get_clean() ?: '';
+        return $action === 'up'
+            ? MigrationService::create($module)->migrateUp()
+            : MigrationService::create($module)->migrateNew();
     }
 
     /**
@@ -229,8 +223,11 @@ class MigrateController extends \yii\console\controllers\MigrateController
      * @param string $migrationPath
      *
      * @return string output
+     * @deprecated since 1.16; use MigrationService::create($module)->migrateUp()
+     * @see MigrationService::create()
+     * @see MigrationService::migrateUp()
      */
-    public static function webMigrateUp(string $migrationPath): string
+    public static function webMigrateUp(string $migrationPath): ?string
     {
         ob_start();
         $controller = new self('migrate', Yii::$app);
@@ -240,7 +237,7 @@ class MigrateController extends \yii\console\controllers\MigrateController
         $controller->color = false;
         $controller->runAction('up');
 
-        return ob_get_clean() ?: '';
+        return ob_get_clean() ?: null;
     }
 
     /**
--- a/protected/humhub/components/Module.php
+++ b/protected/humhub/components/Module.php
@@ -1,6 +1,6 @@
 <?php
 
-/**
+/*
  * @link https://www.humhub.org/
  * @copyright Copyright (c) 2017 HumHub GmbH & Co. KG
  * @license https://www.humhub.com/licences
@@ -16,7 +16,11 @@ use humhub\modules\file\libs\FileHelper;
 use humhub\modules\marketplace\models\Module as OnlineModelModule;
 use humhub\modules\notification\components\BaseNotification;
 use humhub\modules\queue\helpers\QueueHelper;
+use humhub\services\MigrationService;
+use Throwable;
 use Yii;
+use yii\base\InvalidConfigException;
+use yii\db\StaleObjectException;
 use yii\helpers\Json;
 use yii\web\AssetBundle;
 
@@ -239,21 +243,29 @@ class Module extends \yii\base\Module
     /**
      * Enables this module
      *
-     * @return boolean
+     * @return bool|null Result of migration or null if beforeEnable() returned false (since v1.16)
+     * @throws InvalidConfigException
      */
     public function enable()
     {
-        Yii::$app->moduleManager->enable($this);
-        $this->migrate();
+        $result = $this->getMigrationService()->migrateUp();
 
-        return true;
+        if ($result === false) {
+            Yii::error('Could not enable module. Database Migration failed! See previous error for result.', $this->id);
+
+            return false;
+        }
+
+        Yii::$app->moduleManager->enable($this);
+
+        return $result;
     }
 
     /**
      * Disables a module
      *
      * This should delete all data created by this module.
-     * When override this method make sure to invoke call `parent::disable()` **AFTER** your implementation as
+     * When overriding this method, make sure to invoke call `parent::disable()` **AFTER** your implementation as
      *
      * ```php
      * public function disable()
@@ -262,61 +274,40 @@ class Module extends \yii\base\Module
      *     parent::disable();
      * }
      * ```
+     *
+     * @return bool|null Result uninstall-migration or null if beforeDisable() returned false (since v1.16)
+     * @throws InvalidConfigException
+     * @throws StaleObjectException
+     * @throws Throwable
      */
     public function disable()
     {
-        /**
-         * Remove database tables
-         */
-        $migrationPath = $this->getBasePath() . '/migrations';
-        $uninstallMigration = $migrationPath . '/uninstall.php';
-        if (file_exists($uninstallMigration)) {
-
-            /**
-             * Execute Uninstall Migration
-             */
-            ob_start();
-            require_once($uninstallMigration);
-            $migration = new \uninstall;
-            try {
-                $migration->up();
-            } catch (\yii\db\Exception $ex) {
-                Yii::error($ex);
-            }
-            ob_get_clean();
-
-            /**
-             * Delete all Migration Table Entries
-             */
-            $migrations = opendir($migrationPath);
-            $params = [];
-            while (false !== ($migration = readdir($migrations))) {
-                if ($migration == '.' || $migration == '..' || $migration == 'uninstall.php') {
-                    continue;
-                }
-
-                $command ??= Yii::$app->db->createCommand()->delete('migration', 'version = :version', $params);
-
-                $version = str_replace('.php', '', $migration);
-                $command->bindValue(':version', $version)->execute();
-            }
+        try {
+            $result = $this->getMigrationService()->uninstall();
+            ContentContainerSetting::deleteAll(['module_id' => $this->id]);
+            Setting::deleteAll(['module_id' => $this->id]);
+        } catch (Throwable $ex) {
+            Yii::error($ex, $this->id);
+            $result = false;
         }
 
-        ContentContainerSetting::deleteAll(['module_id' => $this->id]);
-        Setting::deleteAll(['module_id' => $this->id]);
-
         Yii::$app->moduleManager->disable($this);
+
+        return $result;
     }
 
     /**
      * Execute all not applied module migrations
+     * @deprecated since v1.16; use static::getMigrationService()->migrateUp()
      */
     public function migrate()
     {
-        $migrationPath = $this->basePath . '/migrations';
-        if (is_dir($migrationPath)) {
-            \humhub\commands\MigrateController::webMigrateUp($migrationPath);
-        }
+        return $this->getMigrationService()->migrateUp();
+    }
+
+    public function getMigrationService(): MigrationService
+    {
+        return new MigrationService($this);
     }
 
     /**
@@ -342,10 +333,15 @@ class Module extends \yii\base\Module
      */
     public function update()
     {
-        if($this->beforeUpdate() !== false) {
-            $this->migrate();
-            $this->afterUpdate();
+        if (!$this->beforeUpdate()) {
+            return null;
         }
+
+        $result = $this->getMigrationService()->migrateUp();
+
+        $this->afterUpdate();
+
+        return $result;
     }
 
     /**
@@ -360,13 +356,11 @@ class Module extends \yii\base\Module
         return true;
     }
 
-
     /**
      * Called right after the module update.
      */
     public function afterUpdate()
     {
-
     }
 
     /**
@@ -483,7 +477,7 @@ class Module extends \yii\base\Module
         $assetDirectory = $this->getBasePath() . DIRECTORY_SEPARATOR . 'assets';
         if (is_dir($assetDirectory)) {
             foreach (FileHelper::findFiles($assetDirectory, ['recursive' => false,]) as $file) {
-                $assetClass =  $assetNamespace . '\\' . basename($file, '.php');
+                $assetClass = $assetNamespace . '\\' . basename($file, '.php');
                 if (is_subclass_of($assetClass, AssetBundle::class)) {
                     $assets[] = $assetClass;
                 }
--- a/protected/humhub/events/MigrationEvent.php
+++ b/protected/humhub/events/MigrationEvent.php
@@ -1,0 +1,39 @@
+<?php
+
+/*
+ * @link      https://www.humhub.org/
+ * @copyright Copyright (c) 2023 HumHub GmbH & Co. KG
+ * @license   https://www.humhub.com/licences
+ */
+
+namespace humhub\events;
+
+use humhub\components\Event;
+use humhub\interfaces\ApplicationInterface;
+use yii\base\Module;
+
+/**
+ * @property int|null $result Result of the migration:
+ * ``
+ *  - `ExitCode::OK`: Success;
+ *  - `ExitCode::UNSPECIFIED_ERROR`: failure;
+ *  - `Null`: nothing done
+ * ``
+ */
+class MigrationEvent extends Event
+{
+    /**
+     * @var \humhub\components\Module|ApplicationInterface|null
+     */
+    public ?Module $module;
+
+    /**
+     * @var string Either `up` or `uninstall`
+     */
+    public string $migration;
+
+    /**
+     * @var string|null Output of the MigrationController's Action
+     */
+    public ?string $output = null;
+}
--- a/protected/humhub/modules/admin/controllers/InformationController.php
+++ b/protected/humhub/modules/admin/controllers/InformationController.php
@@ -8,7 +8,6 @@
 
 namespace humhub\modules\admin\controllers;
 
-use humhub\commands\MigrateController;
 use humhub\libs\SelfTest;
 use humhub\modules\admin\components\Controller;
 use humhub\modules\admin\components\DatabaseInfo;
@@ -18,6 +17,7 @@ use humhub\modules\queue\driver\MySQL;
 use humhub\modules\queue\helpers\QueueHelper;
 use humhub\modules\queue\interfaces\QueueInfoInterface;
 use humhub\modules\search\jobs\RebuildIndex;
+use humhub\services\MigrationService;
 use ReflectionClass;
 use ReflectionException;
 use Yii;
@@ -89,14 +89,18 @@ class InformationController extends Controller
 
     public function actionDatabase(int $migrate = self::DB_ACTION_CHECK)
     {
+        $migrationService = MigrationService::create();
+
         if ($migrate === self::DB_ACTION_RUN) {
+            $migrationService->migrateUp();
             $migrationOutput = sprintf(
                 "%s\n%s",
-                MigrateController::webMigrateAll(),
+                $migrationService->getLastMigrationOutput(),
                 SettingController::flushCache()
             );
         } else {
-            $migrationOutput = MigrateController::webMigrateAll(MigrateController::MIGRATION_ACTION_NEW);
+            $migrationService->migrateNew();
+            $migrationOutput = $migrationService->getLastMigrationOutput();
 
             $migrate = str_contains($migrationOutput, 'No new migrations found.')
                 ? self::DB_ACTION_CHECK
--- a/protected/humhub/modules/admin/controllers/InformationController.php
+++ b/protected/humhub/modules/admin/controllers/InformationController.php
@@ -99,12 +99,11 @@ class InformationController extends Controller
                 SettingController::flushCache()
             );
         } else {
-            $migrationService->migrateNew();
-            $migrationOutput = $migrationService->getLastMigrationOutput();
+            $migrate = $migrationService->hasMigrationsPending()
+                ? self::DB_ACTION_PENDING
+                : self::DB_ACTION_CHECK;
 
-            $migrate = str_contains($migrationOutput, 'No new migrations found.')
-                ? self::DB_ACTION_CHECK
-                : self::DB_ACTION_PENDING;
+            $migrationOutput = $migrationService->getLastMigrationOutput();
         }
 
         $databaseInfo = new DatabaseInfo(Yii::$app->db->dsn);
--- a/protected/humhub/modules/marketplace/components/OnlineModuleManager.php
+++ b/protected/humhub/modules/marketplace/components/OnlineModuleManager.php
@@ -173,7 +173,7 @@ class OnlineModuleManager extends Component
         $this->install($moduleId);
 
         $updatedModule = Yii::$app->moduleManager->getModule($moduleId);
-        $updatedModule->migrate();
+        $updatedModule->getMigrationService()->migrateUp();
 
         (new MarketplaceService())->refreshPendingModuleUpdateCount();
 
--- a/protected/humhub/services/MigrationService.php
+++ b/protected/humhub/services/MigrationService.php
@@ -109,9 +109,24 @@ class MigrationService extends Component
         return $this->lastMigrationResult;
     }
 
-    public function isValid(): bool
+    public function hasMigrations(): bool
     {
         return $this->path !== null;
+    }
+
+    public function hasMigrationsPending(): bool
+    {
+        if (!$this->hasMigrations()) {
+            return false;
+        }
+
+        if ($this->migrateNew() === false) {
+            return false;
+        }
+
+        $migrationOutput = $this->getLastMigrationOutput();
+
+        return !str_contains($migrationOutput, 'No new migrations found.');
     }
 
     /**
@@ -210,7 +225,7 @@ class MigrationService extends Component
         $this->lastMigrationOutput = null;
         $this->lastMigrationResult = null;
 
-        if (!$this->isValid()) {
+        if (!$this->hasMigrations()) {
             return null;
         }
 
--- a/protected/humhub/services/MigrationService.php
+++ b/protected/humhub/services/MigrationService.php
@@ -1,0 +1,324 @@
+<?php
+
+/*
+ * @link      https://www.humhub.org/
+ * @copyright Copyright (c) 2023 HumHub GmbH & Co. KG
+ * @license   https://www.humhub.com/licences
+ */
+
+namespace humhub\services;
+
+use humhub\commands\MigrateController;
+use humhub\components\Application;
+use humhub\components\Event;
+use humhub\components\Module;
+use humhub\events\MigrationEvent;
+use humhub\interfaces\ApplicationInterface;
+use humhub\libs\Helpers;
+use Throwable;
+use Yii;
+use yii\base\ActionEvent;
+use yii\base\Component;
+use yii\base\Controller;
+use yii\base\InvalidConfigException;
+use yii\base\Module as BaseModule;
+use yii\console\ExitCode;
+
+/**
+ * @since 1.16
+ */
+class MigrationService extends Component
+{
+    public const EVENT_AFTER_MIGRATION = 'afterMigration';
+    protected const MIGRATION_NEW = 'new';
+    protected const MIGRATION_UNINSTALL = 'uninstall';
+    protected const MIGRATION_UP = 'up';
+
+    protected BaseModule $module;
+    private ?string $path;
+    private ?int $lastMigrationResult = null;
+    private ?string $lastMigrationOutput = null;
+
+    /**
+     * @param Module|ApplicationInterface|Application|null $module
+     */
+    public function __construct(?BaseModule $module = null)
+    {
+        Helpers::checkClassType($module, [ApplicationInterface::class, Module::class, null]);
+
+        $this->module = $module ?? Yii::$app;
+
+        parent::__construct();
+    }
+
+    public function init()
+    {
+        parent::init();
+
+        /**
+         * Since for console application the id is set to 'humhub-console' and might be configured for the application too,
+         * we need to use the hard-coded 'humhub' string for non-modules.
+         *
+         * @see \humhub\components\console\Application::$id
+         * @see protected/humhub/config/console.php
+         */
+        $moduleId = $this->module instanceof Module
+            ? $this->module->id
+            : 'humhub';
+
+        $this->path = "@$moduleId/migrations";
+
+        $realpath = $this->getPath(true);
+
+        if ($realpath === false || !is_dir($realpath)) {
+            Yii::debug("Module has no migrations directory.", $this->module->id);
+            $this->path = null;
+        }
+    }
+
+    /**
+     * @return Module|ApplicationInterface|Application|null
+     */
+    public function getModule(): BaseModule
+    {
+        return $this->module;
+    }
+
+    private function getPath(bool $resolve = false): ?string
+    {
+        if (!$resolve || $this->path === null) {
+            return $this->path;
+        }
+
+        $path = realpath(Yii::getAlias($this->path));
+
+        if ($path === false) {
+            return null;
+        }
+
+        return $path;
+    }
+
+    public function getLastMigrationOutput(): ?string
+    {
+        return $this->lastMigrationOutput;
+    }
+
+    public function getLastMigrationResult(): ?int
+    {
+        return $this->lastMigrationResult;
+    }
+
+    public function isValid(): bool
+    {
+        return $this->path !== null;
+    }
+
+    /**
+     * Run migrations.
+     */
+    public function migrateNew(): ?bool
+    {
+        return $this->migrate(MigrationService::MIGRATION_NEW);
+    }
+
+    /**
+     * Check for pending migrations
+     */
+    public function migrateUp(): ?bool
+    {
+        return $this->migrate(MigrationService::MIGRATION_UP);
+    }
+
+    /**
+     * @param string $action Must be MigrationService::MIGRATION_ACTION_UP to run migrations,
+     * or MigrationService::MIGRATION_ACTION_NEW to check for pending migrations
+     *
+     * @return bool|null
+     */
+    private function migrate(string $action): ?bool
+    {
+        $result = $this->checkMigrationBefore($action);
+
+        if ($result === null) {
+            return null;
+        }
+
+        // this event is collecting the migration's result status and storing it in our event
+        Event::on(
+            MigrateController::class,
+            Controller::EVENT_AFTER_ACTION,
+            [
+                $this,
+                'onMigrationControllerAfterAction'
+            ],
+            $result
+        );
+
+        // Disable max execution time to avoid timeouts during migrations
+        @ini_set('max_execution_time', 0);
+
+        $module = $this->getModule();
+
+        ob_start();
+        $controller = new MigrateController('migrate', $module, [
+            'db' => Yii::$app->db,
+            'interactive' => false,
+            'color' => false,
+            'migrationPath' => $this->getPath(),
+            'includeModuleMigrations' => true,
+        ]);
+
+        /** @noinspection PhpUnhandledExceptionInspection */
+        $controller->runAction($action);
+
+        $result->output = ob_get_clean() ?: null;
+
+        // we no longer need to listen to this event
+        Event::off(
+            MigrateController::class,
+            Controller::EVENT_AFTER_ACTION,
+            [
+                $this,
+                'onMigrationControllerAfterAction'
+            ]
+        );
+
+        return $this->checkMigrationStatus($result);
+    }
+
+    /**
+     * Catches migration results.
+     *
+     * @internal
+     */
+    public function onMigrationControllerAfterAction(ActionEvent $event)
+    {
+        if (!$event->sender instanceof MigrateController) {
+            return;
+        }
+
+        if (!$event->data instanceof MigrationEvent || $event->data->sender !== $this) {
+            return;
+        }
+
+        $event->data->result = $event->result ?? ExitCode::UNSPECIFIED_ERROR;
+    }
+
+    private function checkMigrationBefore(string $migrationAction): ?MigrationEvent
+    {
+        $this->lastMigrationOutput = null;
+        $this->lastMigrationResult = null;
+
+        if (!$this->isValid()) {
+            return null;
+        }
+
+        return new MigrationEvent([
+            'sender' => $this,
+            'module' => $this->getModule(),
+            'migration' => $migrationAction,
+        ]);
+    }
+
+    /**
+     * @param MigrationEvent $result
+     *
+     * @return bool
+     * @throws InvalidConfigException
+     * @throws Throwable
+     */
+    private function checkMigrationStatus(MigrationEvent $result): bool
+    {
+        $this->lastMigrationOutput = $result->output ?: 'Migration output unavailable';
+        $this->lastMigrationResult = $result->result;
+
+        /** @see \yii\console\controllers\BaseMigrateController::actionUp() */
+        if ($result->result > ExitCode::OK) {
+            Yii::error($this->lastMigrationOutput, $this->module->id);
+        } else {
+            Yii::info($this->lastMigrationOutput, $this->module->id);
+        }
+
+        $this->trigger(self::EVENT_AFTER_MIGRATION, $result);
+
+        /** @see \yii\console\controllers\BaseMigrateController::actionUp() */
+        if ($result->result > ExitCode::OK) {
+            $errorMessage = "Migration failed!";
+
+            if (YII_DEBUG) {
+                throw new InvalidConfigException($errorMessage);
+            }
+
+            Yii::error($errorMessage, $this->module->id);
+
+            return false;
+        }
+
+        return true;
+    }
+
+    public function uninstall(): ?bool
+    {
+        $result = $this->checkMigrationBefore(self::MIGRATION_UNINSTALL);
+
+        if ($result === null) {
+            return null;
+        }
+
+        $path = $this->getPath(true);
+        $uninstallMigration = $path . '/uninstall.php';
+
+        if (!file_exists($uninstallMigration)) {
+            Yii::warning("Module has no uninstall migration!", $this->module->id);
+            return null;
+        }
+
+        /**
+         * Execute Uninstall Migration
+         */
+        ob_start();
+        require_once($uninstallMigration);
+
+        $migration = new \uninstall();
+        $migration->compact = false;
+
+        try {
+            $result->result = $migration->up() === false ? ExitCode::UNSPECIFIED_ERROR : ExitCode::OK;
+        } catch (\yii\db\Exception $ex) {
+            Yii::error($ex, $this->module->id);
+            $result->result = ExitCode::UNSPECIFIED_ERROR;
+        }
+        $result->output = ob_get_clean();
+
+        /**
+         * Delete all Migration Table Entries
+         */
+        $migrations = opendir($path);
+        $params = [];
+        while (false !== ($migration = readdir($migrations))) {
+            if ($migration === '.' || $migration === '..' || $migration === 'uninstall.php') {
+                continue;
+            }
+
+            $command ??= Yii::$app->db->createCommand()->delete('migration', 'version = :version', $params);
+
+            $version = str_replace('.php', '', $migration);
+            $command->bindValue(':version', $version)->execute();
+            $result->output .= "    > migration entry $version removed.\n";
+        }
+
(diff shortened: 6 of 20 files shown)


martin-rueegg:
@luke- 

Currently, when **_enabling_** a module, the existence of a `migrations/uninstall.php` is not considered and any existing new migration is applied. However, when _disabling_ a module, those migrations are nor rolled back (`$migration->donw()`) nor removed from the migration log. Hence, if the `uninstall.php` is missing, then migrations are permanent.

I am not sure this is intended behaviour.

Would it not be better to make `migrations/uninstall.php` mandatory for _any_ migration to be applied? And should a module, that has a `migrations` folder _without_ an `uninstall.php` not be denied to be enabled (maybe unless it's a core module, but those are never `enabled()` in the first place)?

martin-rueegg:
Ping @luke-

martin-rueegg:
While generally the terms `enable` and `disable` are used, there is the method `\humhub\components\Module::getIsActivated()` (and it's related virtual `@property-read bool $isActivated`).

The later is queried by the `\humhub\modules\admin\widgets\InstalledModuleList::initDefaultGroups()` (among other places) to see if the Module is enabled or not. Which is already a bit confusing, given that the state of enabling/disabling is (also) registered in the `module_enabled` table and the `ModuleManager`'s `$enabledModules` array...

It would be less confusing to use `enabled` here, too.

Would it be ok, to deprecate this method in favor of a new method `getIsEnabled()`? If so, it might make sense to include this in the current PR?

Or what exactly is the difference between a Module that is `enabled` but not `active`?

I personally see the need to be able to "suspend" a module. Meaning, it's not actively loaded and used in the code, while it is not disabled (with the consequence of deleting all related data. I've implemented this in our module (See also https://github.com/humhub/humhub/issues/3737#issuecomment-1488472544). Happy to contribute to a solution that this could be implemented in the core. But it would introduce more terms and hence to merge `active`/`enabled` (if they're the same) would be even more important to reduce confusion.

luke- (changes requested):
Thanks for the PR, I really like the improved  migration handling and the added tests :)

- I like the idea to remove the term `active` and always work with `enabled`. If you like feel free to create a PR and mark `activate*` as deprecated for now. @yurabakhtin ok from your side?  
- What do you think about outsourcing the migration logic to a `humhub\protected\services\MigrationService`? I would like the `Module` base class to be as compact as possible.  And I also hope we can get rid of the `migration*` class attributes. 
- In Yii2, the `afterSomething()` methods never have a return value, maybe we should handle it the same way.

martin-rueegg:
> Something like this?

Understood. Very nice.

How about moving the `enable()`, `disable()`, `update()`, and `before/afterUpdate()` to a separate service class, too? It's also code that is used very rarely but the module is loaded on most requests?

Maybe an additional `ModuleMaintenanceService`?

luke-:
> > * I like the idea to remove the term `active` and always work with `enabled`. If you like feel free to create a PR and mark `activate*` as deprecated for now. @yurabakhtin ok from your side?
> 
> Turns out, `\humhub\modules\friendship\Module::getIsEnabled()` is used for a different purpose. Shall we rename this or does that cause too much ripple?

Hmm, I think we can rename it. Any idea? isActivated? :-)

martin-rueegg:
> > > * I like the idea to remove the term `active` and always work with `enabled`. If you like feel free to create a PR and mark `activate*` as deprecated for now. @yurabakhtin ok from your side?
> > 
> > 
> > Turns out, `\humhub\modules\friendship\Module::getIsEnabled()` is used for a different purpose. Shall we rename this or does that cause too much ripple?
> 
> Hmm, I think we can rename it. Any idea? isActivated? :-)

For me the question would be: Does the friendship module do anything, if the system is not set to use friendship functionality. Otherwise, we could introduce the `isSuspended()` state, where a module is enabled, but not actively used. This state allows a "broken" Module to be in the system (data won't be deleted) but not actively usable. Once e.g. an update is provided and functionality restored, the "suspension" is removed and the module fully functional again.

The advantage of this is that the data does not get lost if an update compromises the module, and the system is still functional while just one module "suspended".

Otherwise, I'd use a more specific term, e.g. `isFriendshipEnabled()` (I'd not use "active" as the functionality may be "active" and still no one uses them, so there is no "active" friendship. "enabled" or "available" would reflect the potentiality.

marc-farre:
@martin-rueegg thanks for this great work!

FYI, we have an issue for the `Module::update()` method: https://github.com/humhub/humhub/issues/4653

luke-:
> > > > Something like this?
> > > 
> > > 
> > > Understood. Very nice.
> > > How about moving the `enable()`, `disable()`, `update()`, and `before/afterUpdate()` to a separate service class, too? It's also code that is used very rarely but the module is loaded on most requests?
> > > Maybe an additional `ModuleMaintenanceService`?
> > 
> > 
> > I really like the idea of refactoring this logic into a separate service too. But in many cases the `enable` or `disable` methods are extended. e.g. https://github.com/humhub/polls/blob/master/Module.php#L64 - there are also `disableContentContainer' .
> 
> Of course, the `enable()` `update()`, and `disable()` would have to be supported for now. But we could mark them deprecated. Then we'd have to check in the `ModuleMaintenanceService` if the methods are used or not. I think that should be doable and the little overhead is no problem as this happens rarely.
> 
> > Ideally, however, content would be cleaned up automatically.
> 
> That was envisioned here:
> 
> * [[WIP][RFC] Proof-of-concept for drop-in-replacement of class names in table fields #6217](https://github.com/humhub/humhub/pull/6217)
>   Which would allow the data to be there and cleaned up at a later point, if desired by the user/admin.

Ok. Do I understand correctly? We leave `enable()` and `disable()`, ... as they are. If we automate this later, we can deprecate these methods.

> > By the way, is a `beforeDisable` and `afterDisable` really necessary, are there so many cases here where simply overriding the `disable` and positioning the `parent::disable()` call is not enough?
> 
> And no, I would replace all `before/after` methods with events that can be hooked in as I did with the current proposed MigrationService. But then I'd do the same for the `beforeUpdate/afterUpdate` so that it is consistent.

Sounds good.

Something like this?

- `humhub\events\MigrationEvent` - triggered by `MigrationService`, as `EVENT_MIGRATE_UP` or `EVENT_UINSTALL`.

- `humhub\events\ModuleEvent` - triggerd by `Module` class, as `EVENT_BEFORE_ENABLE`, `EVENT_AFTER_ENABLE`, `EVENT_BEFORE_DISABLE`, `EVENT_AFTER_DISABLE`, `EVENT_BEFORE_UPDATE`, `EVENT_AFTER_UPDATE`, `EVENT_BEFORE_ENABLE_CONTENTCONTAINER` 

However, I would not integrate the migration result into the `EVENT_ENABLE`, but consider it as two separate events. For example, the following events are triggered when a module is enabled

1. `Module::EVENT_BEFORE_ENABLE`
2. `MigrationService::EVENT_MIGRATE_UP`
3. `Module::EVENT_AFTER_ENABLE`

martin-rueegg:
> @martin-rueegg thanks for this great work!
> 
> FYI, we have an issue for the `Module::update()` method: #4653

So there seems to be a series of issues related to the module upgrade:
- #3872
  As @luke- rightly [points out](https://github.com/humhub/humhub/issues/3872#issuecomment-884763893), this has to do with the class already loaded.
- #4364
  This has probably to do with the fact that the module config is cached and the cache is not cleared/updated at the right time. The current PR does already delete the cache a bit more aggressively, but I have not considered the update yet. Also, the events are not unregistered before update, which could be implemented quite easily.
- #4653
  Agree with @luke- that also this is only the case, if the code in `update()` has been added in the given update.

Now, php does not offer the possibility to unload a class - at least not without hack and generally available. It does offer to create aliases of classes. But it does not really help, as the original class would normally still have the same name.

A reliable solution would be to call the migration and update in a separate process. But ...

- In a web server environment, this would require a request to the same server (which may or may not work, depending on the server configuration - the FQDN may not be resolvable by the web server. Or do you already do self-referenced requests - without the client/browser involved?). 
- Calling a PHP script as a system call will also only work, if allowed by the configuration. This also applies to the CLI.

For CLI this could be tested before the update starts and thrown an error if calling a separate PHP process is not possible.
For web environment we could perform a secondary request to the server from the client. But this may fail, too.

There are some other approaches:

### using PHARs
Make modules distributed from the marketplace as PHAR, and use versioned PHARs like with https://phar.io/ and so every version of a module has it's own namespace and an updated version would not conflict with a previous version. However, this involves changes to the marketplace infrastructure to automatically build the PHARs for each version.


### Requiring the module class to include a versioned name

Module devs could be require to incorporate the module version into the name of the module class. As such, after an update, the module class would be called differently (reflected in the new `config.php`) and loaded.

But that is not really a solution as other classes that are called during update may have the same problem, just it has not hit anyone yet.

Also, you have to rename the class with every release and that will most likely cause a lot of files in your module to change because they're referencing that name. Still, it would be quite a safe way.

### Aliased class during update

We could [alias](https://www.php.net/manual/en/function.class-alias.php) the module class to a versioned name during update, unless such a class actually already exists. If it is provided by the module, it is loaded. Otherwise the existing class is aliased. Then this versioned name is used to instantiate the new module instance. Then, any code accessing the module through `Yii::$app->getModule('mymodule')` would get the new instance. So properties, methods, constants (even static ones) need to be called from there during update, and never on the static class directly.

If the module class introduces new fields or methods, the module dev would need to create a copy of the original class under the versioned name (or a name specified in `config.php` under the `update_class` name). Of course, this class may not extend from the regular class name, as this would kill the entire trick. But this exactly is also a problem, as any `instanceof` comparison with the real class name would now fail.

### Automatic copying

Alternatively to the above aliasing, this coping could be done automatically, renaming the class's name in the temporary file, load it and instantiate the module.

The update process would look something like this:

1. Fire BEFORE_UPDATE event
1. unregister all events
3. replace files from update
4. if the new module class has _the same FQN_ as the old one (which will be most of the cases)
    1. copy module class file (must be at the same location than the previous one)
    2. replace module class name with versioned or random name
    3. load altered class file
    5. instantiate module
4. If the module class name _has changed_, instantiate the new name instead.
6. The new instance replaces the `Yii::$app->getModule('mymodule')`
7. register events 
8. Fire BEFORE_MIGRATION event
9. run migration
10. Fire AFTER_MIGRATION event
11. Fire AFTER_UPDATE_EVENT

It's the dev's responsability to mitigate any other change of already loaded classes of the module. Also the `instanceof` issue mentioned above still applies.


So yes, the most reliable way in terms of the classes loaded, would be to run the update in two processes:
1. Initiate update (before events), replace files, call second process
2. Call migrations and after events

martin-rueegg:
> Ok. Do I understand correctly? We leave `enable()` and `disable()`, ... as they are. If we automate this later, we can deprecate these methods.

I would deprecate them already. But still support them, if they are there.

> Something like this?
> 
> * `humhub\events\MigrationEvent` - triggered by `MigrationService`, as `EVENT_MIGRATE_UP` or `EVENT_UINSTALL`.
> * `humhub\events\ModuleEvent` - triggerd by `Module` class, as `EVENT_BEFORE_ENABLE`, `EVENT_AFTER_ENABLE`, `EVENT_BEFORE_DISABLE`, `EVENT_AFTER_DISABLE`, `EVENT_BEFORE_UPDATE`, `EVENT_AFTER_UPDATE`, `EVENT_BEFORE_ENABLE_CONTENTCONTAINER`
> 
> 
> However, I would not integrate the migration result into the `EVENT_ENABLE`, but consider it as two separate events. For example, the following events are triggered when a module is enabled
> 
> 1. `Module::EVENT_BEFORE_ENABLE`
> 2. `MigrationService::EVENT_MIGRATE_UP` 
> 3. `Module::EVENT_AFTER_ENABLE`

Yes, something like this. Although there are also the ModuleManager events for [enable](https://docs.humhub.org/docs/develop/modules/#enabled-a-module) and [disable](https://docs.humhub.org/docs/develop/modules/#disable-module). Not sure we need all of them? I find that having 5 events is a bit overkill.

I'd suggest:

1. `ModuleManager::EVENT_BEFORE_MODULE_ENABLE` with `humhub\events\ModuleEvent`
2. `MigrationService::EVENT_MIGRATE_UP` with `humhub\events\MigrationEvent`
3. `ModuleManager::EVENT_AFTER_MODULE_ENABLE` with `humhub\events\ModuleEvent`

And maybe adapt `humhub\events\ModuleEvent` if needed. I have to look into the code.

luke-:
@marc-farre @martin-rueegg Regarding the Module Update, I've created a new Issue here:  https://github.com/humhub/humhub/issues/6687

martin-rueegg:
@luke- 

There are also a few places, where the term `enabled` or `isEnabled` is used
- `\humhub\modules\file\models\FileHistory::isEnabled()` refers to the functionality, not the module (should be fine)
- `\humhub\modules\marketplace\Module::isEnabled()` connected to
- `\humhub\modules\marketplace\Module::$enabled` which is set to false if `Yii::$app->params['humhub']['apiEnabled']` is not `true`, and could also be disable through configuration
- `\humhub\modules\content\components\ContentContainerModuleManager::isEnabled($id)` checks if module `$id` is enabled on the container

So the there is some potential for confusion between
-  `\humhub\modules\marketplace\Module::isEnabled()`, which refers to a configured state, and it's inherited
-  `\humhub\components\Module::getIsEnabled()`, which refers to the module enabled state (but which is always true for a core module like the marketplace.

Would it make sense to also rename `\humhub\modules\marketplace\Module::isEnabled()` into `\humhub\modules\marketplace\Module::getIsEnabled()`, overriding the `\humhub\components\Module::getIsEnabled()`? In a way it seems to have a similar meaning. Or does that increase the confusion?

Alternatively, we could obviously leave it, as it is, or rename it to `\humhub\components\Module::isMarketplaceEnabled()`, analogous to the friendship module.

luke-:
> @luke-
> 
> There are also a few places, where the term `enabled` or `isEnabled` is used
> 
> * `\humhub\modules\file\models\FileHistory::isEnabled()` refers to the functionality, not the module (should be fine)
:+1: 

> * `\humhub\modules\marketplace\Module::isEnabled()` connected to
> * `\humhub\modules\marketplace\Module::$enabled` which is set to false if `Yii::$app->params['humhub']['apiEnabled']` is not `true`, and could also be disable through configuration

TBD

> * `\humhub\modules\content\components\ContentContainerModuleManager::isEnabled($id)` checks if module `$id` is enabled on the container

Is ok or?

> So the there is some potential for confusion between
> 
> * `\humhub\modules\marketplace\Module::isEnabled()`, which refers to a configured state, and it's inherited
> * `\humhub\components\Module::getIsEnabled()`, which refers to the module enabled state (but which is always true for a core module like the marketplace.
> 
> Would it make sense to also rename `\humhub\modules\marketplace\Module::isEnabled()` into `\humhub\modules\marketplace\Module::getIsEnabled()`, overriding the `\humhub\components\Module::getIsEnabled()`? In a way it seems to have a similar meaning. Or does that increase the confusion?
> 
> Alternatively, we could obviously leave it, as it is, or rename it to `\humhub\components\Module::isMarketplaceEnabled()`, analogous to the friendship module.

Hmm, I would prefer to introduce `\humhub\modules\marketplace\Module::isMarketplaceEnabled()` analgour to the new  `\humhub\modules\friendship\Module::isFriendshipEnabled()`.
What do you think @yurabakhtin ?

For `Module::isFriendshipEnabled()` we need to do some refactorings in the `calendar`, `cfiles`, `content-bookmarks`, `mail` , `task`.

yurabakhtin:
> Hmm, I would prefer to introduce `\humhub\modules\marketplace\Module::isMarketplaceEnabled()` analgour to the new `\humhub\modules\friendship\Module::isFriendshipEnabled()`.
> What do you think @yurabakhtin ?

@luke- Yes, I prefer the name `isMarketplaceEnabled()` too.

martin-rueegg:
@luke- in case you have the chance to review this, that'd be awesome.

protected/humhub/services/MigrationService.php
```diff
@@ -0,0 +1,258 @@
...
+        return $this->path;
+    }
+
+    public function migrateUp(string $action): ?bool
+    {
+        $result = new MigrationEvent([
+            'path' => $this->path,
+            'action' => $action,
+            'migration' => 'up',
+            'deactivateOnError' => true
```
luke-, line 83: - Not sure the `MigrationEvent` should handle deactivation?
- Do we really want to delete? Currently deactive === data loses

luke- (changes requested):
Thanks, see my thoughts below.

protected/humhub/components/Module.php
```diff
@@ -37,6 +41,9 @@ class Module extends \yii\base\Module
      */
     private ?array $_moduleInfo = null;
 
+    public ?int $migrationResult = null;
```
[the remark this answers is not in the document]
martin-rueegg: No, this can be removed, of course. Forgot them. Thanks!

protected/humhub/services/MigrationService.php
```diff
@@ -0,0 +1,258 @@
...
+    }
+
+    /**
+     * @param MigrationEvent $result
+     *
+     * @return bool
+     * @throws InvalidConfigException
+     * @throws Throwable
+     */
+    public function checkMigrationStatus(MigrationEvent $result): bool
```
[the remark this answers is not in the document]
martin-rueegg: Doesn't need to be public. Was an automated code extraction in PhpStorm which defaults to `public`.

However, since the `MigrationService` could be extended from to incorporate additional functionality, `protected` might be an option? Or we make it `private` and if need arises, it can still be change to `protected`.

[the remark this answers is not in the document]
luke-: I would say `private` when required make `protected`.

protected/humhub/services/MigrationService.php
martin-rueegg: Are you happy to keep this like that for now? But we might address it again in the PR (which I have already ready) that refactors module update and migration

luke- (changes requested):
@martin-rueegg Sorry, had a pending review here. Not sure there is something open from me.

martin-rueegg:
> @martin-rueegg Sorry, had a pending review here. Not sure there is something open from me.

I see. Yes, the latest comment wasn't there before ... :-)

protected/humhub/services/MigrationService.php
martin-rueegg: Ok, sounds fair.

We use the events for the before/after hooks.

protected/humhub/services/MigrationService.php
```diff
@@ -0,0 +1,262 @@
...
+
+    /**
+     * @param Module $module
+     * @param array $config = ['path' => string|null]
+     */
+    public function __construct(Module $module, array $config = [])
+    {
+        $this->module = $module;
+
+        $path = $config['path'] ?? null;
```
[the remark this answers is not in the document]
martin-rueegg: I'm not 100% sure about the paths in relation to modules:

- the module's `config.php` has to reside in a subfolder of any of the module search path (app config parameter).
- however, the module can be _anywhere_, as long as one registered auto-loader can find it.

My questions:
1. Is now the `/migrations` folder always relative to the `config.php` file or the module's class file (usually, but not necessarily `module.php`)?
2. Are there real-world situations where the two paths (of `config.php` and the module class file) are different?

And yes, if we use the reflection to get the module's path, the class will be loaded. And yes, this might have been an issue.

But I think we should consider to revisit the `MigrationService` together with the `ModuleService` that I have already written to do the install, update, remove tasks of the module. I find it a bit difficult to discus details that do depend on code that is not yet in this PR (to keep it in an manageable size). If you prefer, we could apply this PR to a new branch and then apply the other two on top of it: a) for renaming `active` => `enabled`, and b) the `ModuleService`.

> Maybe we could make $module optional, when not given, the migration service covers alls migrations?

I like that idea. I think we could also move some functionality from the `MigrateController` to the `MigrateService`, keeping the "routing and managing" in the controller, but the `executing` in the service. And then either pass the application as the `$module` parameter or leave it optional which will default to `Yii::$app` ...

protected/humhub/services/MigrationService.php
```diff
@@ -0,0 +1,262 @@
...
+    {
+        return $this->module;
+    }
+
+    public function getPath(): ?string
+    {
+        return $this->path;
+    }
+
+    public function migrateUp(string $action): ?bool
```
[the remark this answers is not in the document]
martin-rueegg: The current Module class has four methods that involve migrations:
1. `enable()`
2. `update()`
3. `migrate()`
4. `disable()`

To mirror those four "actions" I've added the `$action` parameter that is only used in the event (`MigrationEvent::$action`; so that registering callbacks may know what is the current action).

The "MigrationEvent::$migration" on the other hand shows the `up`, `down` and `uninstall`. And there's `new` to to check for new migrations.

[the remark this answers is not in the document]
luke-: Hmm, I would actually like to remove this.

Can you give me UseCases for the `MigrateEvent`? And if there are any, that a differentiation between `enable`, `update`, `migrate` action  is necessary here?

If there is logic, could this not be simply implemented in the `Module::update()` method?

protected/humhub/services/MigrationService.php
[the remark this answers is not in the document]
luke-: > I'm not 100% sure about the paths in relation to modules:
> 
> * the module's `config.php` has to reside in a subfolder of any of the module search path (app config parameter).
> * however, the module can be _anywhere_, as long as one registered auto-loader can find it.
> 
> My questions:
> 
> 1. Is now the `/migrations` folder always relative to the `config.php` file or the module's class file (usually, but not necessarily `module.php`)?

Following structure should be expected: https://docs.humhub.org/docs/develop/modules#module-structure

A HumHub module can have a `migrations` folder in its base folder.
The same folder also contains the module class `Module.php` and `config.php`.    

I would not assume special cases here.

> 2. Are there real-world situations where the two paths (of `config.php` and the module class file) are different?

Not sure, what you mean. 

But `config.php` and `Module.php` file names are fixed. Theoretically, `Module.php` could have a different class name. However, I am not aware of any cases.
> 
> And yes, if we use the reflection to get the module's path, the class will be loaded. And yes, this might have been an issue.



> 
> But I think we should consider to revisit the `MigrationService` together with the `ModuleService` that I have already written to do the install, update, remove tasks of the module. I find it a bit difficult to discus details that do depend on code that is not yet in this PR (to keep it in an manageable size). If you prefer, we could apply this PR to a new branch and then apply the other two on top of it: a) for renaming `active` => `enabled`, and b) the `ModuleService`.
> 

Unfortunately, a tricky situation. I struggle a lot with reviewing large PR changes and I'm concerned that potentially small changes will be not discussed in detail. 

> > Maybe we could make $module optional, when not given, the migration service covers alls migrations?
> 
> I like that idea. I think we could also move some functionality from the `MigrateController` to the `MigrateService`, keeping the "routing and managing" in the controller, but the `executing` in the service. And then either pass the application as the `$module` parameter or leave it optional which will default to `Yii::$app` ...

luke- (changes requested):

protected/humhub/services/MigrationService.php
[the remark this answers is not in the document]
martin-rueegg: I've tried to implement this along your suggestion:
https://github.com/humhub/humhub/pull/6550#issuecomment-1831587153

I have removed `MigrationEvent::$action`

Simple use case would be to clear a cache after a migration. Or to suspend the module upon migration error. Or to enable a feature in the settings upon migration success.

These events are basically a replacement for the current module's `migrate()` method.

I thought the long-term idea was to reduce as much code from the Module class as possible. Hence the migration code should be going into the `MigrationService` and later also the `enable()`, `update()` and `disable()` should be going into a `ModuleService`, which also adresses the issue that the updated code is not available after `update` if the class has already been loaded.

Having the functionality in the service, would no longer "allow" to add that code to methods in the module. Then module devs have two options:
- either register to the events called by those services
  - `MigrationService`: before and after migration
  - `ModuleService`: before and after install/enable/disable/uninstall plus after update (so the class is not loaded before the file is replaced)
- alternatively, they may overwrite the `getMigrationService()` and create their own implementation of the service.

Hence, the events are meant to avoid the need to create a custom service class, but allow to just register to the events.

protected/humhub/services/MigrationService.php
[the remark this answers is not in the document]
martin-rueegg: > Is it really worth implementing a `$config['path']` variable in addition to the module?

I have removed the explicit evaluation of the path variable. But given it's a component, a calling method could still set the path upon initialization. Which, in fact. is used for backward-compatibility of the `MigrateController::webMigration()`. However, by default the path is now evaluated by the module's path alias "@$module-id/migrations" (or "@humhub/migrations" for the application itself).

martin-rueegg:
> > But I think we should consider to revisit the `MigrationService` together with the `ModuleService` that I have already written to do the install, update, remove tasks of the module. I find it a bit difficult to discus details that do depend on code that is not yet in this PR (to keep it in an manageable size). If you prefer, we could apply this PR to a new branch and then apply the other two on top of it: a) for renaming `active` => `enabled`, and b) the `ModuleService`.
> 
> Unfortunately, a tricky situation. I struggle a lot with reviewing large PR changes and I'm concerned that potentially small changes will be not discussed in detail.

I understand that. And I appreciate your effort. I also try to make meaningful commits in order to chunk the changes into logical entities.


> > > Maybe we could make $module optional, when not given, the migration service covers alls migrations?
> > 
> > 
> > I like that idea. I think we could also move some functionality from the `MigrateController` to the `MigrateService`, keeping the "routing and managing" in the controller, but the `executing` in the service. And then either pass the application as the `$module` parameter or leave it optional which will default to `Yii::$app` ...

I have now implemented this idea. The static functions of the `MigrateController` have been deprecated and replaced with `MigrationService` calls.

protected/humhub/components/Module.php
```diff
@@ -239,21 +243,28 @@ public function getIsActivated(): bool
...
     public function enable()
     {
         Yii::$app->moduleManager->enable($this);
-        $this->migrate();
 
-        return true;
+        $result = $this->getMigrationService()->migrateUp();
+
+        if ($result === false) {
+            Yii::$app->moduleManager->disable($this);
```
luke-, line 256: Since `disable` deletes data, I would like to drop this. Until we have a suspend or similar.

Not sure: But maybe in some reinstall situations the enable is executed with existing data.

protected/humhub/events/MigrationEvent.php
```diff
@@ -0,0 +1,49 @@
...
+{
+    /**
+     * @var \humhub\components\Module|ApplicationInterface|null
+     */
+    public ?Module $module;
+
+    /**
+     * @var string|null path to the migrations folder for modules, empty for application
+     */
+    public ?string $path;
```
luke-, line 33: Can be removed?

protected/humhub/events/MigrationEvent.php
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * @var string Either `up` or `uninstall`
+     */
+    public string $migration;
+
+    /**
+     * @var string|null Error message to be used in the log
+     */
+    public ?string $errorMessage = null;
```
luke-, line 43: When `errorMessage` null, means no error?

protected/humhub/events/MigrationEvent.php
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+namespace humhub\events;
+
+use humhub\components\Event;
+use humhub\interfaces\ApplicationInterface;
+use yii\base\Module;
+
+/**
+ * @property int|null $result Result of the migration:
```
luke-, line 16: Can be removed?

I would prefer a status code instead the fixed `errorMessage`.

protected/humhub/modules/admin/controllers/InformationController.php
```diff
@@ -8,24 +8,30 @@
...
 use ReflectionException;
 use Yii;
 
 /**
- * Informations
+ * Information
  *
  * @since 0.5
  */
 class InformationController extends Controller
```
luke-, line 30: Related to another PR?

protected/humhub/services/MigrationService.php
```diff
@@ -0,0 +1,328 @@
...
+
+    /**
+     * @return Module|ApplicationInterface|Application|null
+     */
+    public function getModule(): BaseModule
+    {
+        return $this->module;
+    }
+
+    public function getPath(bool $resolve = false): ?string
```
luke-, line 80: private?

protected/humhub/services/MigrationService.php
```diff
@@ -0,0 +1,328 @@
...
+    {
+        return $this->lastMigrationOutput;
+    }
+
+    public function getLastMigrationResult(): ?int
+    {
+        return $this->lastMigrationResult;
+    }
+
+    public function isValid(): bool
```
luke-, line 105: What is this needed for?

protected/humhub/services/MigrationService.php
```diff
@@ -0,0 +1,328 @@
...
+    {
+        return $this->migrate(MigrationService::MIGRATION_NEW, $output);
+    }
+
+    /**
+     * Check for pending migrations
+     *
+     * @param string|null $output Output of the migration, if available
+     */
+    public function migrateUp(?string &$output = null): ?bool
```
luke-, line 125: Why should the output of a migration be passed to the migrate command? Same for `migrateNew` and following methods.

protected/humhub/services/MigrationService.php
```diff
@@ -0,0 +1,328 @@
...
+use yii\base\Module as BaseModule;
+use yii\console\ExitCode;
+
+/**
+ * @since 1.16
+ */
+class MigrationService extends Component
+{
+    public const EVENT_AFTER_MIGRATION = 'afterMigration';
+    public const EVENT_BEFORE_MIGRATION = 'beforeMigration';
```
luke-, line 32: I know I suggested having two events here earlier. But isn't it enough to only implement `AFTER_MIGRATE`?

luke- (changes requested):

protected/humhub/services/MigrationService.php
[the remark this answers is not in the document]
luke-: @martin-rueegg Thank you.

Maybe we misunderstood a bit.

I think we should have implementable (enable/disable/update) methods in the module base class for the module developer. So that the workaround via events etc. is not necessary.

We should only move the core code that is necessary for migrating/enabling/updating into services.

protected/humhub/modules/marketplace/components/OnlineModuleManager.php
```diff
@@ -173,7 +173,7 @@ public function update($moduleId)
         $this->install($moduleId);
 
         $updatedModule = Yii::$app->moduleManager->getModule($moduleId);
-        $updatedModule->migrate();
+        $updatedModule->getMigrationService()->migrateUp();
```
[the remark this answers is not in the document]
martin-rueegg: I don't know. I just replaced the call to `migrate()` with it's respective call of the `MigrationService`. 

I'd suggest to keep it simple for this PR which is only about migration. This code will change again as soon as we have the `ModuleService` which addresses `enable`, `update`, `disable`.

protected/humhub/components/Module.php
martin-rueegg: No, as explained [before](https://github.com/humhub/humhub/pull/6550#discussion_r1419434181), the `$moduleManager->disable` does _not_ delete any data, it only removes the module from the the `module_enable` table and the `Yii::$app`'s modules again. It's the `Module::disable()` that deletes data.

Why should the module be enabled in the `module_enable` table, if the migrations fail?

Maybe we should only "enable" the module in the module manager only _after_ the migration and only if they are successful?

protected/humhub/events/MigrationEvent.php
martin-rueegg: What do you mean by a fixed `errorMessage`?

Firstly, the $result variable is already part of the `humhub\components\Event`. So I can't remove it.
Secondly, the `$result` _is_ filled with a status code:
- `ExitCode::OK`: which is integer `0` and stands for Success
- `ExitCode::UNSPECIFIED_ERROR`: which is integer `1` which stands for failure;
- `Null`: nothing done (e.g. no migrations directory or no migrations pending)

protected/humhub/events/MigrationEvent.php
martin-rueegg: No, when set to `null` the [standard message](https://github.com/humhub/humhub/pull/6550/files/1b50ea92d6a9f8f534abe34389372de1220b4710#diff-381c812c100b93c52f71ee2cd25c98821a7436b5d1223f2f2d96059fcc05950dR237) will be used. This allows a module developper to add custom error message. But this can be removed or renamed, if preferred.

protected/humhub/modules/admin/controllers/InformationController.php
martin-rueegg: I leave this up to you. I honestly don't have the time to create an extra pr for this. I'll remove the change here.

protected/humhub/services/MigrationService.php
martin-rueegg: As you wish.

Currently the developer has the option to do both in `Module`:

```php

class Module {

public function migrate() {

  // code that is run before the migration, replaced with new EVENT_BEFORE_MIGRATION
  // would currently allow to change the migration path
 
  parent::migrate();

  // code that is run after the migration, replaced with new EVENT_AFTER_MIGRATION
}

}
```

But I have no use case for the EVENT_BEFORE_MIGRATION. happy to remove.

protected/humhub/services/MigrationService.php
martin-rueegg: The idea was, that when getting the `MigrationService` from `$module->getMigrationService`, I can check if there is a valid migration path (path `/module/migrations` exist). Hence, I can decide whether or not to run the migration.

protected/humhub/services/MigrationService.php
martin-rueegg: For convenience. Since it's just a reference, it does not consume any (considerable amount of) memory.

But to have the output of the migration makes the call in the `InformationController` cheaper:
```php
MigrationService::create()->migrateUp($migrationOutput);
```
rather than:
```php
$service = MigrationService::create();
$service->migrateUp();
$migrationOutput = $service->getLastMigrationOutput()
```

protected/humhub/services/MigrationService.php
martin-rueegg: done.

protected/humhub/events/MigrationEvent.php
martin-rueegg: nope. But is private now. With private getter only. And `$config` has also been removed. So this should do no harm.

protected/humhub/commands/MigrateController.php
```diff
@@ -223,18 +223,12 @@ public static function webMigrateAll(): string
...
-        ob_start();
-        $controller = new self('migrate', Yii::$app);
-        $controller->db = Yii::$app->db;
-        $controller->interactive = false;
-        $controller->migrationPath = $migrationPath;
-        $controller->color = false;
-        $controller->runAction('up');
-
-        return ob_get_clean() ?: '';
+        return MigrationService::create(null, ['path' => $migrationPath])->migrateUp();
```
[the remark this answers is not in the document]
martin-rueegg: done.

protected/humhub/events/MigrationEvent.php
luke-: I would prefer to remove. Should be no need to maniplulate the error message.

protected/humhub/events/MigrationEvent.php
luke-: Aah ok, missed that.

So the default `result` attribute in the Event class, indicates the Migration success status?

protected/humhub/components/Module.php
martin-rueegg: Ok, no worries.

What do you think of the following?

> Maybe we should only "enable" the module in the module manager only after the migration and only if they are successful?

protected/humhub/services/MigrationService.php
luke-: Ok got it, Maybe we could mark it as `private`?

martin-rueegg: Well, if it's private, it makes no sense anymore. Because e.g. from the Module class or the MigrateController or InstallController I could no longer check if there is a valid path (since the `getPath` is now also private).

Internally, I only use it in one single place, so a private method would be overkill.

So it boils down to the question, if we want to be discoverable, if there is a valid path.

Alternatively, we could create a method `pendingMigrations(): ?bool` which would return
- `true` if path is valid and pending migrations where found
- `false` if path is valid but no pending migrations found
- `null` if path is invalid

protected/humhub/services/MigrationService.php
luke-: Ok, I see. Thank you.

I still prefer the second option. This makes the code easier to understand and we don't need more in-depth optimizations here.

protected/humhub/events/MigrationEvent.php
martin-rueegg: That's correct.

Currently, the MigrateController always returns `ExitCode::OK` or `ExitCode::UNSPECIFIED_ERROR`. So in theory we could translate them into a bool (`true`: success, `false`: error, `null`: nothing migrated)

protected/humhub/services/MigrationService.php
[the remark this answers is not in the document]
martin-rueegg: I see.

I thought the main goal was to reduce the code that needs to be loaded on most requests (as the module classes are usually loaded). Specifically code that is hardly ever used (as enable/migrate/update/disable) should be outsourced, so it is only loaded if required.

If we keep the enable/update/disable methods in the class, than we do not gain much, as the current methods in the base class were very light-weight and the heavy-lifting for a module's disable procedure, for example, is in the dev's concrete implementation. Getting this away would serve us much more than the few lines we've now saved with the `MigrationService`.

protected/humhub/services/MigrationService.php
martin-rueegg: EVENT_BEFORE_MIGRATION has been removed.

protected/humhub/components/Module.php
luke-: @martin-rueegg Yeah, I really like this idea. Maybe with error message: 

```php

    public function enable()
    {
        $migrationService = $this->getMigrationService();

        if (!$migrationService->migrateUp()) {
            Yii::error('Could not enable module. Database Migration failed! Result: ' .
                $migrationService->getLastMigrationOutput(), $this->id);
            return false;
        }

        Yii::$app->moduleManager->enable($this);

        return true;
    }

```

martin-rueegg: Sure, I can add an error message.

The migration's output is sent to the log in any way. As warning, if it's successful. As error if it failed:
https://github.com/humhub/humhub/pull/6550/files#diff-381c812c100b93c52f71ee2cd25c98821a7436b5d1223f2f2d96059fcc05950dR237-R241

But it may be worth adding an other error.

luke-: Ok, then the migration output does not have to be included in the error message again. But a hint as to why the module was not enabled would be good.

martin-rueegg: Already there:
https://github.com/humhub/humhub/blob/374501b9d30fe5c5df634870a2c7e690e9a41f16/protected/humhub/components/Module.php#L251-L259

protected/humhub/services/MigrationService.php
luke-: What about two methods?

- hasMigrations(): bool     /* or `isValid();` migrations available */  
- hasPendingMigrations(): bool

protected/humhub/modules/marketplace/components/OnlineModuleManager.php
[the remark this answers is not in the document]
luke-: I'm currently not sure, if we really need a `ModuleService`...

protected/humhub/components/Module.php
martin-rueegg: Done: [Add hasMigrations(Pending)()](https://github.com/humhub/humhub/pull/6550/commits/bfe92a47fa159d5cc7d95c707074217085ba4ba1);

protected/humhub/services/MigrationService.php
martin-rueegg: Done: [Add hasMigrations(Pending)()](https://github.com/humhub/humhub/pull/6550/commits/bfe92a47fa159d5cc7d95c707074217085ba4ba1);

protected/humhub/modules/marketplace/components/OnlineModuleManager.php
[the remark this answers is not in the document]
martin-rueegg: Ok, I guess that can be seen in the light of #6687.

Are you happy to leave it the way "it was" before, for now?

protected/humhub/services/MigrationService.php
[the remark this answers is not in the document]
luke-: My goal was more improved logical code separation for better maintainability.
I really like it, that the Migration code is now in a separate service.

I currently tend to leave enable/disable/update methods in the module class. I assume that all classes are already loaded and in the PHP cache on application servers.

protected/humhub/modules/marketplace/components/OnlineModuleManager.php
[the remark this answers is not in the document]
luke-: The specific change here is fine with me.

Regarding ModuleService etc. we should discuss in #6687.

luke- (approved):
@martin-rueegg Thanks for the work here. Should I click the merge? :-)

martin-rueegg:
> @martin-rueegg Thanks for the work here. Should I click the merge? :-)

Yes. Unless you want me to squash the commits first to avoid a merge.

And then get ready for the next PR :-D

marc-farre:
@martin-rueegg @luke- thanks for this great work.

Can I ask you to be document ALL changes on method declarations so that we can keep class inheriting them working?

E.g. on https://github.com/humhub/humhub/commit/4d142b002b4e681da4885b75569077bf663bceb9#diff-3634c090cf95aea01f8986dca83fa4721c2f44685893c63d683de1550ffc02f2R31 a return `:bool` has been added to `humhub\modules\content\components\ContentContainerModule::disable()`

And now we have e.g. this error (but in fact it will occur on many modules): `Declaration of humhub\modules\linklist\Module::disable() must be compatible with humhub\modules\content\components\ContentContainerModule::disable(): ?bool`

And I can see in the commit that there are changes on other methods too.

So it is very important for each change on method to document it here (I know most of them have already been added): https://github.com/humhub/humhub/blob/develop/MIGRATE-DEV.md#type-restrictions

Thanks very much!

martin-rueegg:
@marc-farre 

Thanks for your feedback.

I do my best to document every change in MIGRATE-DEV, as you have noticed. Unfortunately, I have missed to document this one (and maybe others). For which I'm very sorry.

As you can see, the one you mentioned is already in the queue to be fixed:
- @6755

For other's you're welcome to create a PR and help correcting.

martin-rueegg:
I have now added the following previously undocumented new type restrictions:

- `\humhub\components\bootstrap\ModuleAutoLoader::findModules()` is enforcing types on method parameters and return value
- `\humhub\components\bootstrap\ModuleAutoLoader::locateModules()` is enforcing return type
- `\humhub\components\ModuleManager::register()` is enforcing types on method parameters

See
- #6755

marc-farre:
@martin-rueegg  Thanks very much for your quick fix.

I think one is missing in the doc: `protected/humhub/components/bootstrap/ModuleAutoLoader.php::findModulesByPath()`
https://github.com/humhub/humhub/commit/4d142b002b4e681da4885b75569077bf663bceb9#diff-a201fcb66f328a30a64fa095e7d69ed8c2fd94bdda7bd06a6b84fcbdc32691ceR129

I don't have the time to check all previous PR, so I hope it's OK for them also.

Let's continue the discussion here: https://github.com/humhub/humhub/issues/6759

martin-rueegg:
> I think one is missing in the doc: `protected/humhub/components/bootstrap/ModuleAutoLoader.php::findModulesByPath()` [4d142b0#diff-a201fcb66f328a30a64fa095e7d69ed8c2fd94bdda7bd06a6b84fcbdc32691ceR129](https://github.com/humhub/humhub/commit/4d142b002b4e681da4885b75569077bf663bceb9#diff-a201fcb66f328a30a64fa095e7d69ed8c2fd94bdda7bd06a6b84fcbdc32691ceR129)
> 
> I don't have the time to check all previous PR, so I hope it's OK for them also.
> 
> Let's continue the discussion here: #6759

Thanks, missed that one, too. I guess I was too tired last night. Apologies.

fixed in
- #6762

marc-farre:
@martin-rueegg No problem, thanks!

luke-:
@martin-rueegg The REST API tests seems to be broken due to this PR: https://github.com/humhub/rest/issues/148 Any idea?

PR #6566, opened 2023-09-14
Fix: invalid argument exceptions

martin-rueegg:
## Update `humhub\exceptions`
- simplify constructor arguments
- make it compatible with `yii\base\InvalidArgumentException` or `\Exception` for that matter
- Reduce usage of Traits in `\humhub\exceptions\*`

See https://github.com/humhub/humhub/pull/6548#discussion_r1325425500 and following

- Refactor

###Does this PR introduce a breaking change?

- Yes, but only for v1.15 code, which is addressed in this PR.

### The PR requirements

- [ ] Changelog was modified

### DIFF
--- a/protected/humhub/exceptions/InvalidArgumentExceptionTrait.php
+++ b/protected/humhub/exceptions/InvalidArgumentExceptionTrait.php
@@ -1,6 +1,6 @@
 <?php
 
-/**
+/*
  * @link      https://www.humhub.org/
  * @copyright Copyright (c) 2023 HumHub GmbH & Co. KG
  * @license   https://www.humhub.com/licences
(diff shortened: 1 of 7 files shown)


martin-rueegg:
@luke- 

Maybe this could be a good compromise?

I have not renamed `InvalidArgumentTypeException`, however, it is now compatible with `yii\base\InvalidArgumentException` or `\Exception` for that matter. As such, it can be used as a drop-in replacement for `yii\base\InvalidArgumentException` with the additional possibility to specify valid values:
- if the first parameter contains a ` ` (space), it is treated as the message string
- if it does not, it is considered a parameter name and `$valid` and `$given` are considered

luke-:
> @luke-
> 
> Maybe this could be a good compromise?
> 
> I have not renamed `InvalidArgumentTypeException`, however, it is now compatible with `yii\base\InvalidArgumentException` or `\Exception` for that matter. As such, it can be used as a drop-in replacement for `yii\base\InvalidArgumentException` with the additional possibility to specify valid values:
> 
> * if the first parameter contains a ` ` (space), it is treated as the message string
> * if it does not, it is considered a parameter name and `$valid` and `$given` are considered

Yes, would also be a solution, although I would like `InvalidArgumentType` and `InvalidArgumentValue` a bit better. That shows more clearly for me that it is about the value or type, while `InvalidArgument` is simply generic and just receives a text.  

Just  to ensure when I throw `throw new InvalidArgumentException('parameter');` it the exception message ` Argument $valid passed to ' . __METHOD__ . ' must be one of the following types: string, string[] - NULL given.`. ?

https://github.com/humhub/humhub/pull/6566/files#diff-27efec3584982da65aafd18b341c7dd4a45962f9e2a5cf10c645376cc222db4eR78

martin-rueegg:
@luke-

> Yes, would also be a solution, although I would like `InvalidArgumentType` and `InvalidArgumentValue` a bit better. That shows more clearly for me that it is about the value or type, while `InvalidArgument` is simply generic and just receives a text.

Ok, implemented with `InvalidArgumentValueException` and `InvalidArgumentTypeException`. The latter does now no longer inherit from the former, but both inherit from `yii\base\InvalidArgumentException`.

> Just to ensure when I throw `throw new InvalidArgumentException('parameter');` it the exception message ` Argument $valid passed to ' . __METHOD__ . ' must be one of the following types: string, string[] - NULL given.`. ?

Not quite sure, what you mean.

In order to be upwards compatible, non of the parameters may be enforced. Hence, when `$parameterOrMessage` does not contain a space, it is considered a parameter name. In this case, parameter `$valid` must also be provided. The test [you referenced](https://github.com/humhub/humhub/pull/6566/files#diff-27efec3584982da65aafd18b341c7dd4a45962f9e2a5cf10c645376cc222db4eR78) does test if this is enforced.

Does that answer your question?

martin-rueegg:
@luke- Thanks. Yes, please do! 👍

PR #6568, opened 2023-09-15
Enh: improve UUID validator

martin-rueegg:
The current `humhub\libs\UUID::is_valid()` is not really suitable to check if an input is a valid UUID, as the `$uuid` parameter _MUST_ be a string, resulting in an Exception otherwise.

If I want to do a validation, I'd like to know, if the input is valid or not. There is no point in having to test both for a possible Exception _and_ the result to be true or to do a `is_string()` first.

The new `UUID::validate()` addresses these issues and
- returns 
  - a validated (and normalized) UUID on success
  - `null` on error
- allows to configure the normalization. Default format is `xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx` (lowercase, no brackets, but dash delimiters). This format is used in the database for `guid` columns. Though all three characteristics can be configured:
  - case
  - delimiter
  - curly brackets

- Feature

- No
- ~~However, using `UUIDValidater` in models could render some `guid` values that currently have not been validated as invalid. Maybe this commit should be removed and then added to `next` instead.~~ (those commits have been removed)

- [x] Changelog was modified

**Other information:**


protected/humhub/modules/content/models/ContentContainer.php
```diff
@@ -42,10 +42,11 @@ public function rules()
     {
         return [
             [['pk', 'owner_user_id'], 'integer'],
-            [['class', 'pk', 'guid'], 'required'],
```
ArchBlood, line 45: Is there a reason why `guid` shouldn't be required?

martin-rueegg: @ArchBlood thanks for your review. If you look at Line 48 of the new version, you'll notice the new `UUIDValidator` to be applied to `guid`. That Validator ensures that there is a valid `guid`.

Though please note, that this is still a draft PR. And as mentioned in the introductory comment, the second commit might not be due for the currently selected branch as it introduces some changes that need to be discussed. One of which is, that `UUIDValidator` automatically fills an empty `guid` with a valid `guid`.

ArchBlood: Could this then be randomized in cases where an empty `guid` is required? Namely in places where guest posting is implemented?

martin-rueegg: I'm not quite sure what you mean by "randomized." The generation of a UUID _is_ a randomization. But I guess I misunderstood.

By default, `UUIDValidator` would generate a new (hence randomized and valid) UUID for `null` values and save that new UUID in the attribute. However, by setting [`UUIDValidator::$autofillWith`](https://github.com/humhub/humhub/pull/6568/files#diff-03de26eae2159bb01a861e012850fe98526752187d82ffb4046185d708a2bf4bR48) to `false`, then `empty()` values would be left alone (as long as [`UUIDValidator::$allowNull = true`](https://github.com/humhub/humhub/pull/6568/files#diff-03de26eae2159bb01a861e012850fe98526752187d82ffb4046185d708a2bf4bR29).

So maybe this could be tweaked a bit to achieve the following (or some variation of it):

| case | `$allowNull` | `$strict` | `$autofillWith` | Input | Result |
|---|---|---|---|---|---|
| 1* | `true`* | `false`* | `true`* | `null` | UUID |
| 2 | `true`* | `false`* | `null` | `null` | `null` |
| 3* | `true`* | `false`* | `true`* | `''`, `0`, `'0'`, `false` `[]` | UUID |
| 4 | `true`* | `false`* | `null` | `''`, `0`, `'0'`, `false` `[]` | `null` |
| | --- | --- | --- | --- | --- |
| 5 | `true`* | `true` | `true`* | `null` | UUID |
| 6 | `true`* | `true` | `null` | `null` | `null` |
| 7 | `true`* | `true` | not relevant | `''`, `0`, `'0'`, `false` `[]` | value error |
| | --- | --- | --- | --- | --- |
| 8 | `false` | not relevant | `null` | not relevant | config error |
| | --- | --- | --- | --- | --- |
| 9 | `false` | `true` | `true`* | `null` | UUID |
| 10 | `false` | `true` | `false` | `null` | value error |
| 11 | `false` | `true` | not relevant | `''`, `0`, `'0'`, `false` `[]` | value error |
| 12 | `false` | `false`* | `true`* | `''`, `0`, `'0'`, `false` `[]` | UUID |
| 13 | `false` | `false`* | `false` | `''`, `0`, `'0'`, `false` `[]` | value error |

*) denotes the default setting

These settings can be controlled in `rules()` when a `$config` array is added.

This would require a small change of the code, including
- setting `$strict = false` as the default
- more elaborate [evaluation of empty values](https://github.com/humhub/humhub/pull/6568/files#diff-03de26eae2159bb01a861e012850fe98526752187d82ffb4046185d708a2bf4bR71-R72)
- store the `$strict` value [before validating the string](https://github.com/humhub/humhub/pull/6568/files#diff-03de26eae2159bb01a861e012850fe98526752187d82ffb4046185d708a2bf4bR92) and resetting it afterwards

Would that help?

martin-rueegg:
@luke- This one? With usage implementation here:
- #6587

luke- (approved):

PR #6577, opened 2023-09-19
Fix using of user scenarios

yurabakhtin:

- [x] Bugfix

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**
Fix small wrong using of user scenarios during review the PR https://github.com/humhub/rest/pull/140


luke- (approved):

yurabakhtin:
@marc-farre Ok, thanks for the info, it can be fixed in next PR because it was already merged.
I have used the `static::` because it was used before here https://github.com/humhub/humhub/blob/d0c2de9f46933fcf6aae6857877a3e2fc6f1fcf9/protected/humhub/modules/user/models/User.php#L278-L282.

PR #6587, opened 2023-09-23
Enh: Apply UUID validator

martin-rueegg:

- Refactor

- Yes

UUID (e.g. used in `guid` columns) must actually be valid UUIDs, meaning
- valid format and characters
- unique at least within any given table

Applied to
- `contentcontainer.guid`
- `content.guid`
- `file.guid`
- `space.guid`
- `user.guid` 

- [x] It's submitted to the `next` branch, _not_ the `master` branch if no hotfix
- [x] Changelog was modified

**Other information:**

### DIFF
--- a/protected/humhub/components/behaviors/PolymorphicRelation.php
+++ b/protected/humhub/components/behaviors/PolymorphicRelation.php
@@ -9,6 +9,7 @@
 namespace humhub\components\behaviors;
 
 use Exception;
+use humhub\components\ActiveRecord;
 use humhub\libs\Helpers;
 use humhub\modules\content\components\ContentActiveRecord;
 use humhub\modules\content\components\ContentAddonActiveRecord;
@@ -16,9 +17,7 @@ use ReflectionClass;
 use ReflectionException;
 use Yii;
 use yii\base\Behavior;
-use yii\base\Model;
-use yii\db\ActiveRecord;
-use yii\db\BaseActiveRecord;
+use yii\db\ActiveRecordInterface;
 use yii\db\IntegrityException;
 
 /**
@@ -33,53 +32,57 @@ class PolymorphicRelation extends Behavior
     /**
      * @var string the class name attribute
      */
-    public $classAttribute = 'object_model';
+    public string $classAttribute = 'object_model';
 
     /**
      * @var string the primary key attribute
      */
-    public $pkAttribute = 'object_id';
+    public string $pkAttribute = 'object_id';
 
     /**
      * @var boolean if set to true, an exception is thrown if `object_model` and `object_id` is set but does not exist
      */
-    public $strict = false;
+    public bool $strict = false;
 
     /**
      * @var array the related object needs to be an "instanceof" at least one of these given classnames
      */
-    public $mustBeInstanceOf = [];
+    public array $mustBeInstanceOf = [];
 
     /**
      * @var ActiveRecord|object|null the cached object
      */
-    private $cached = null;
+    private ?object $cached = null;
 
     /**
      * Returns the Underlying Object
      *
-     * @return ActiveRecord|object|null
+     * @return ActiveRecordInterface|ActiveRecord|object|null
      * @throws IntegrityException
      */
-    public function getPolymorphicRelation()
+    public function getPolymorphicRelation(): ?object
     {
         if ($this->cached !== null) {
             return $this->cached;
         }
 
-        $object = static::loadActiveRecord(
+        $record = static::loadActiveRecord(
             $this->owner->getAttribute($this->classAttribute),
             $this->owner->getAttribute($this->pkAttribute)
         );
 
-        if ($this->strict && !$object && !empty($this->classAttribute) && !empty($this->pkAttribute)) {
-            throw new IntegrityException('Call to an inconsistent polymorphic relation detected on ' . get_class($this->owner) . ' (' . $this->owner->getAttribute($this->classAttribute) . ':' . $this->owner->getAttribute($this->pkAttribute) . ')');
+        if ($this->strict && !$record && !empty($this->classAttribute) && !empty($this->pkAttribute)) {
+            throw new IntegrityException(
+                'Call to an inconsistent polymorphic relation detected on '
+                . ($this->owner === null ? 'NULL' : get_class($this->owner))
+                . ' (' . $this->owner->getAttribute($this->classAttribute) . ':' . $this->owner->getAttribute($this->pkAttribute) . ')'
+            );
         }
 
-        if ($object !== null && $this->validateUnderlyingObjectType($object)) {
-            $this->cached = $object;
+        if ($record !== null && $this->validateUnderlyingObjectType($record)) {
+            $this->cached = $record;
 
-            return $object;
+            return $record;
         }
 
         return null;
@@ -88,9 +91,9 @@ class PolymorphicRelation extends Behavior
     /**
      * Sets the related object
      *
-     * @param mixed $object
+     * @param object|null $object
      */
-    public function setPolymorphicRelation($object)
+    public function setPolymorphicRelation(?object $object)
     {
         if ($this->cached === $object) {
             return;
@@ -100,7 +103,7 @@ class PolymorphicRelation extends Behavior
             $cached = $this->cached;
             $this->cached = $object;
 
-            if ($object instanceof ActiveRecord) {
+            if ($object instanceof ActiveRecordInterface) {
                 $class = get_class($object);
                 if ($cached === null || get_class($cached) !== $class) {
                     $this->owner->setAttribute($this->classAttribute, $class);
@@ -114,7 +117,7 @@ class PolymorphicRelation extends Behavior
         }
     }
 
-    public static function getObjectModel(Model $object): string
+    public static function getObjectModel(ActiveRecordInterface $object): string
     {
         return $object instanceof ContentActiveRecord || $object instanceof ContentAddonActiveRecord
             ? $object::getObjectModel()
@@ -130,13 +133,24 @@ class PolymorphicRelation extends Behavior
     }
 
     /**
+     * Returns if the polymorphic relation is established
+     *
+     * @since 1.16
+     * @noinspection PhpUnused
+     */
+    public function isPolymorphicRelationLoaded(): bool
+    {
+        return $this->cached !== null;
+    }
+
+    /**
      * Validates if given object is of an allowed type
      *
      * @param mixed $object
      *
      * @return boolean
      */
-    private function validateUnderlyingObjectType($object)
+    private function validateUnderlyingObjectType(?object $object)
     {
         if (empty($this->mustBeInstanceOf)) {
             return true;
@@ -155,12 +169,12 @@ class PolymorphicRelation extends Behavior
     /**
      * Loads an active record based on classname and primary key.
      *
-     * @param $className
-     * @param $primaryKey
+     * @param string $className
+     * @param string|int $primaryKey
      *
-     * @return null|ActiveRecord
+     * @return null|ActiveRecord|ActiveRecordInterface
      */
-    public static function loadActiveRecord($className, $primaryKey)
+    public static function loadActiveRecord(string $className, $primaryKey): ?ActiveRecordInterface
     {
         if (empty($className) || empty($primaryKey)) {
             return null;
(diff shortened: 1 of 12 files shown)


protected/humhub/components/behaviors/PolymorphicRelation.php
```diff
@@ -27,16 +27,21 @@
...
      * @var string the primary key attribute
      */
-    public $pkAttribute = 'object_id';
+    public string $pkAttribute = 'object_id';
+
+    /**
+     * @var callable|null
+     * @since 1.16
+     */
+    public $onSet;
```
luke-, line 44: Is this used?

protected/humhub/modules/content/components/ContentContainerActiveRecord.php
```diff
@@ -248,6 +249,40 @@ public function afterSave($insert, $changedAttributes)
...
 
+    /**
+     * @since 1.16
+     */
+    public function onSetPolymorphicRelationOwner(?Component $owner)
+    {
+        if (!$owner instanceof ContentContainer) {
+            return;
+        }
+
```
luke-, line 260: Hmm, it is a special case and actually not an ideal implementation that in the `ContentContainer` case two records have the same GUID. 

So, I am unsure whether we should explicitly check the ContentContainer case here.

protected/humhub/components/behaviors/PolymorphicRelation.php
```diff
@@ -92,28 +101,53 @@ public function setPolymorphicRelation($object)
...
-            if ($object instanceof ActiveRecord) {
+            if ($object instanceof ActiveRecordInterface) {
                 $this->owner->setAttribute($this->classAttribute, self::getObjectModel($object));
                 $this->owner->setAttribute($this->pkAttribute, $object->getPrimaryKey());
             }
+
+            /**
+             * @since 1.16
+             */
+            if ($object->hasMethod('onSetPolymorphicRelationOwner')) {
```
luke-, line 112: - Couldn't we use `PolymorphicRelation::onSet` instead of these two `hasMethod()` checks?

- As alternative, we may could add an option `PolymorphicRelation::harmonizeGuid` which makes sure the guid of both are identical and set.

luke- (changes requested):
@martin-rueegg Thanks for the contribution. Please see my remarks below.

protected/humhub/components/behaviors/PolymorphicRelation.php
martin-rueegg: You're right. This is left-over from the idea to register callbacks here. See further comments where we discuss the `on*` Methods ...

protected/humhub/components/behaviors/PolymorphicRelation.php
martin-rueegg: Yes, that would be an option and I think that was my original idea.

Maybe it would make sense to define the `$onSet` field as an array of callbacks:

```php
/**
 * @var callable[]
 * @since 1.16
 */
public array $onSet = [];
```

and the call back would have a uniform signature:

```php
public function(PolymorphicRelation $relation, ActiveRecordInterface $owner, object $object): void;
// or more restricted
public function(PolymorphicRelation $relation, ActiveRecordInterface $owner, ActiveRecordInterface $object): void;
```

When calling the callback, the latter is given in array notation, and the first element is a class name that corresponds to the class of the $object, then the method from the second element will be called on this particular `$object`, and not statically.

protected/humhub/modules/content/components/ContentContainerActiveRecord.php
luke-: Yes, the global GUID table would be the better apporach. But currently all the `guid` attributes are already harmonized between `ContentContainer` and `ContentContainerActiveRecord` models. Why do we need such a new handling here?

luke- (changes requested):

protected/humhub/modules/content/components/ContentContainerActiveRecord.php
luke-: `ContentContainerActiveRecord::afterSave($insert)` does this.


I would prefer to keep it that way. Maybe we could add a UUID option that `ContentContainerAR` has no GUID auto generation? 

Especially if it is planned to replace the code later, I would keep it as simple as possible.

martin-rueegg: > `ContentContainerActiveRecord::afterSave($insert)` does this.

I see. Sorry, has been a while since I created the PR and have go get my head around it again.

So that's the line then:
https://github.com/humhub/humhub/pull/6587/files#diff-a2232aa8f7a9a3c12bcbc0443d6e59ec96f9a4d5e95e62b8498a0f9ad81a40d1L225
Which has been replaced in this PR with the functionality in `onSetPolymorphicRelationOwner()`

> I would prefer to keep it that way.

Ok, it can be reverted. It is still happening in `ContentContainerActiveRecord::afterSave()`, tough,. Just called in the following way:

- `ContentContainerActiveRecord::afterSave()`
  - `PolymorphicRelation::setPolymorphicRelation()`
    - `ContentContainerActiveRecord::onSetPolymorphicRelationOwner()`

> Maybe we could add a UUID option that `ContentContainerAR` has no GUID auto generation?

Yes, we could probably do that. IIRC it is also already possible. But the default is auto-assign. I'll look up what config we need here. But in both scenarios, the current as well as the new code, the GUID should be set already during validation. I realize that only now. Sorry.

> Especially if it is planned to replace the code later, I would keep it as simple as possible.

BTW,  the special case of ...
```php
if (!$owner instanceof ContentContainer) {
    return;
}
```
... will in practice never happen. It is _always_ a `ContentContainer`. I just added it to avoid a fatal error, since the `$cached` property of the PolymorphicRelationship instance could theoretically be any object and might not have a `guid` property...

luke-: Ok, I understand. For me, a solution that is as simple and easy to understand as possible is important. Then comes elegance and flexibility. :-)

Hmm, since there is only one case with the identical GUID for `ContentContainer` & `ContentContainerAR`, I would prefer the simplest possible solution here.

e.g. 
- `ContentContainerAR and the UUID validator has an `allowEmpty` option
- `ContentContainerAR::afterSave` sets the GUID.

martin-rueegg: I'm not sure this is a real-world example, but this pre-existing test leaves the $contentContainer without GUID. My patch would have made sure that also in this case, the correct GUID is assigned.

```php
  public function testGuid()
    {
        $user = User::findOne(['id' => 1]);

        // make sure we have a fresh ID and GUID
        $user->id = 9;
        $user->guid = UUID::v4();

        $contentContainer = new ContentContainer();
        $contentContainer->setPolymorphicRelation($user);

        $this->assertTrue($contentContainer->save());
        $this->assertEmpty($contentContainer->getErrors('guid'));
    }
```

Allowing the ContentContainer to have an empty guid does prevent the error from happening, but will result in a saved record without guid. Because even if you'd save the $user later, it would no longer update the guid in the $contentContainer, given it's not a new record to Yii.

But I guess this situation never happens in real.

luke-: Ok, I understand. The situations (CC creation) are very limited.  Ideally, we refactor/fix the problem of duplicate GUID storage and thus also solve the cases.

luke- (approved):

protected/humhub/modules/content/components/ContentContainerActiveRecord.php
martin-rueegg: > Ok, I understand. The situations (CC creation) are very limited. Ideally, we refactor/fix the problem of duplicate GUID storage and thus also solve the cases.

As you have seen, I've created a PR that would address this:
- #6718 

However, in the meantime we need to solve this issue as the above mentioned test fails (also if it is in it's original) version, as it does test something that in real-world never happens.

We have two options:
- change the test to save the `$user` instead of the `$contentContainer`. Then the user's `afterSave()` would create a CC and add the guid. That would reflect what the real code is doing
- add the `GUID` behavior to `ContentContainer` (or add a `beforeSave` method) that fills in the `guid` upon save if there is none (i.e. if CCAR has not set it). But that could lead to a case where the CCAR and CC records have different `guid`s
- add a `setPolymorphicRelation()` method to `ContentContainer` that calls the behavior's `setPolymorphicRelation()` first and then, upon success, copies the `guid` from the CCAR. This would be a similar approach to what I had with the `onSetPolymorphicRelationOwner`-idea, but more straight forward. The function calls would be as follows
  - `ContentContainerActiveRecord::afterSave()`
    1. calling `ContentContainer::setPolymorphicRelation()`
       1. calling`PolymorphicRelation::setPolymorphicRelation()` (similar to a `parent::method()` call)
          - checking type and copying `id` and `class`
       2. copying guid if CC is new
    2. setting the other stuff it currently does  

The latest approach would allow to do stuff the test is doing, ie. 
1. create CC separately from CCAR, 
2. set the relation, 
3. save either CC or CCAR first

Would that work for you?

martin-rueegg: It would look something like this: https://github.com/humhub/humhub/pull/6587/commits/e6e7c9bda8c5d5629cee525ef055f132df633afe

luke-: Hmm, so if we always save the CCAR `$spaceOrUser->save()` in all places (Core, Rest API, User Importer), it would probably be easiest to just modify the test.  

Maybe it's also worth to add following method in CC

```php
    /**
     * @inheritDoc
     */
    public function save($runValidation = true, $attributeNames = null)
    {
        if ($this->isNewRecord && (empty($this->guid) || empty($this->class))) {
            throw new InvalidCallException(
                'Could not create standalone ContentContainer instance. ' .
                'Instance will be automatically created on ContentContainerActiveRecord::afterSave()');
        }

        return parent::save($runValidation, $attributeNames);
    }
```

martin-rueegg: > Hmm, so if we always save the CCAR `$spaceOrUser->save()` in all places (Core, Rest API, User Importer), it would probably be easiest to just modify the test.

I don't know, if the test was the only place where CC was saved independently. However, since the `CCAR::afterSave()` creates a new instance of CC in any case for a new record, it would not make much sense and lead to an error, if CC was already created for this record.

Hence, yes I think it's safe to assume, that we always do that.

I have now updated the test.

> Maybe it's also worth to add following method in CC (if it's possible to catch `CC->save()`
> 
> ```
>     /**
>      * @inheritDoc
>      */
>     public function save($runValidation = true, $attributeNames = null)
>     {
>         if ($this->isNewRecord && (empty($this->pk) || empty($this->class))) {
>             throw new InvalidCallException(
>                 'Could not create standalone ContentContainer instance. ' .
>                 'Instance will be automatically created on ContentContainerActiveRecord::afterSave()');
>         }
> 
>         return parent::save($runValidation, $attributeNames);
>     }
> ```

Yes, that is a possiblity. Right now I've opted for validation instead. `class` was already required anyway, and `guid` is now, too.

If you prefer to have an error thrown, rather than validation fail, I'm happy to add your method. An exception is stronger than validation, since validation can be disabled and the result of `save()` ignored.

Whatever you prefer.

protected/humhub/components/behaviors/PolymorphicRelation.php
```diff
@@ -129,14 +132,23 @@ public function resetPolymorphicRelation()
         $this->cached = null;
     }
 
+    /**
+     * Returns if the polymorphic relation is established
+     * @since 1.16
+     */
+    public function isPolymorphicRelation(): bool
```
luke-, line 139:
```suggestion
    public function isPolymorphicRelationLoaded(): bool
```

For me the name is not quite clear.

luke- (changes requested):

protected/humhub/modules/content/components/ContentContainerActiveRecord.php
luke-: Thanks I'm fine with the current solution.

If we save ourselves code/checks, we could also omit the new UUID validator from the `ContentContainer` model. In the future there should be a different solution here anyway?

martin-rueegg: Cool.

Yes, with the [`global_id` table](https://github.com/humhub/humhub/pull/6718), the UUID validator would be used there, but not on those records implementing the proposed `GlobalActiveRecord` interface (indicating its a AR related to global_id)

protected/humhub/components/behaviors/PolymorphicRelation.php
martin-rueegg: Cool.

luke- (approved):

PR #6553, opened 2023-09-11
Enh: Support log assertions

martin-rueegg:
So far, there is no possibility to easily check in a test if a specific log entry has been issued.

This PR allows a test to assert that
- a specific message by providing
  - log level
  - category (includes and excludes)
  - message text or message regex
- a given message has _not_ been issued

### What kind of change does this PR introduce?

- Feature

### Does this PR introduce a breaking change?

- No

### The PR fulfills these requirements

- [x] It's submitted to the `next` branch, _not_ the `master` branch if no hotfix
- [x] When resolving a specific issue, it's referenced in the PR's description (e.g. `Fix #xxx[,#xxx]`, where "xxx" is the Github issue number)
- [x] All [tests](#issuecomment-new) are passing
- [x] New/updated tests are included
- [x] Changelog was modified

If adding a **new feature**, the PR's description includes:
- [x] A convincing reason for adding this feature (to avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it)

### Depends on and includes
-  ~~#6564~~
- #6562

PR #6588, opened 2023-09-23
Enh Testing: Backport #6553

martin-rueegg:
Back-porting from next to develop:
- #6516 
- #6553

- Bugfix
  - #6516 
- Feature
  - `assert*ContainsRegex`
  - `assert*Log*`
- Refactor
  - moves all db-related static functions to `DbHelper`
  - moves all assert* functions to trait so it can be used in both `HumHubDbTestCase` and `HumHubHelper`
  - make methods static where possible, so they can be used in class setup and tear-down

- No

Only affects tests. And is fully backwards-compatible.

- [x] Changelog was modified

**Other information:**


protected/humhub/tests/codeception/_support/HumHubDbTestCase.php
```diff
@@ -437,113 +183,15 @@ public function setContentContainerPermission(
...
+    public static function logout()
     {
         Yii::$app->user->logout();
     }
-
-    /**
-     * @see \yii\db\Connection::createCommand()
-     * @since 1.15
-     */
-    public static function dbCommand($sql = null, $params = []): Command
```
luke-, line 456: These new methods have not realized before. Please see `DbHelper` comment.

protected/humhub/tests/codeception/_support/HumHubHelperTrait.php
```diff
@@ -0,0 +1,666 @@
...
+use yii\db\ExpressionInterface;
+use yii\helpers\FileHelper;
+use yii\log\Dispatcher;
+
+/**
+ * Humhub Test Helper Functions
+ *
+ * @since 1.15
+ */
+trait HumHubHelperTrait
```
luke-, line 39: Why use a trait here?

luke- (changes requested):
Thanks, Looks good to me. Just not quite happy with the `DbHelper`. Had overlooked that in the earlier PR - not sure we should need this abstraction.

protected/humhub/helpers/DbHelper.php
```diff
@@ -0,0 +1,126 @@
...
+namespace humhub\helpers;
+
+use Yii;
+use yii\db\Command;
+use yii\db\Query;
+
+/**
+ * @since 1.15
+ */
+class DbHelper
```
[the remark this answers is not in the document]
martin-rueegg: I understand that it makes not so much of a difference. I personally prefer not having to write the same thing over and over again. But it's a different story if you have to maintain an API. I guess it's mainly useful in tests and migrations. Otherwise I've not been using these types of queries much, yet. Maybe that's why I find them cumbersome for the basic insert, update, and delete actions.

Now I have "reverted" the move of the methods into a separate class. That's probably not exactly what you wanted, but right now I do not have the capacity to re-write the places where it has been used. I also need to adapt my module.

"Reverted" because I've placed them now into the trait as some methods there rely on this.

protected/humhub/tests/codeception/_support/HumHubHelperTrait.php
martin-rueegg: Because the classes 
- `\tests\codeception\_support\HumHubHelper` and
- `\tests\codeception\_support\HumHubDbTestCase`
do share already a common set of functions but do not derive from the same parent classes.

Particularly, when it comes to cleaning the cache, it is useful to have only one place where things need to be amended. That's why the trait has been introduced in 
- #6516

Also, it is good to have all the assert* functions available in both cases. So that's why I have now added/moved all those functions into the trait. This reduces the codebase to maintain.

protected/humhub/helpers/DbHelper.php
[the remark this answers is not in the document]
luke-: @martin-rueegg Ok, understand. As a workaround, can we mark these methods as deprecated. Then it is at least clear that we do not want to maintain them in the long term.

luke- (approved):

PR #6590, opened 2023-09-23
Enh: `file` fields: code cleanup and optimization

martin-rueegg:
This PR contains code cleanup and code optimizations that are common to the new file column PRs:
- #6591
- #6592
- #6593
- #6594

### What kind of change does this PR introduce?

- Code style update

### Does this PR introduce a breaking change?

- No

If yes, please describe the impact and migration path for existing applications:

**The PR fulfills these requirements:**

- [x] It's submitted to the `develop` branch, _not_ the `master` branch if no hotfix
- [ ] When resolving a specific issue, it's referenced in the PR's description (e.g. `Fix #xxx[,#xxx]`, where "xxx" is the Github issue number)
- [x] All [tests](#issuecomment-new) are passing
- [ ] New/updated tests are included
- [ ] Changelog was modified

PR #6591, opened 2023-09-23
Enh: `file` fields: `category`

martin-rueegg:
Adds `category` field to `File` table/model.

- Feature

- No

- [x] Changelog was modified

**Other information:**

Depends on/includes:
- #6590

### DIFF
--- a/protected/humhub/modules/file/migrations/m230618_135509_file_add_category_column.php
+++ b/protected/humhub/modules/file/migrations/m230618_135509_file_add_category_column.php
@@ -56,5 +56,20 @@ class m230618_135509_file_add_category_column extends Migration
         )
                 ->execute()
         ;
+
+        if ($schema->getColumn('onlyoffice_key')) {
+            $command->update(
+                $this->table,
+                ['category' => File::CATEGORY_ONLY_OFFICE],
+                [
+                'and',
+                ['not', ['object_model' => null]],
+                ['not', ['object_id' => null]],
+                ['not', ['onlyoffice_key' => null]],
+                ]
+            )
+                ->execute()
+            ;
+        }
     }
 }
--- a/protected/humhub/modules/file/migrations/m230618_135509_file_add_category_column.php
+++ b/protected/humhub/modules/file/migrations/m230618_135509_file_add_category_column.php
@@ -28,10 +28,6 @@ class m230618_135509_file_add_category_column extends Migration
      */
     public function safeUp(): void
     {
-        $command = Yii::$app->getDb()
-                            ->createCommand()
-        ;
-
         $schema = Yii::$app->getDb()->getTableSchema($this->table, true);
         $after = $schema->getColumn('state') ? 'state' : 'guid';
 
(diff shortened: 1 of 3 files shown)


protected/humhub/modules/file/models/File.php
```diff
@@ -126,9 +179,30 @@ public function getHistoryFiles()
...
     {
+        /**
+         * Used for file history
+         *
+         * @see FileHistory::createEntryForFile()
+         */
         $this->old_updated_by = $this->getOldAttribute('updated_by');
         $this->old_updated_at = $this->getOldAttribute('updated_at');
 
+        if ($this->category === null) {
```
luke-, line 190: I would leave that out for the beta for now.
But I like the idea, to use the category to store the file type.

protected/humhub/modules/file/models/File.php
```diff
@@ -126,9 +179,30 @@ public function getHistoryFiles()
...
+            /** @noinspection NestedPositiveIfStatementsInspection */
+            if ($this->object_model && $this->object_id !== null) {
+                $this->category = self::CATEGORY_ATTACHED_FILE;
+
+                $mime_type = $this->mime_type;
+                if ($mime_type && str_starts_with($mime_type, 'image/')) {
+                    $this->category = self::CATEGORY_ATTACHED_IMAGE;
+                }
+
+                if ($this->hasProperty('onlyoffice_key') && $this->onlyoffice_key) {
```
luke-, line 200: Should be done in the oo module.

protected/humhub/modules/file/models/File.php
```diff
@@ -8,56 +8,86 @@
...
+ * Following properties are optional and for module-dependent use:
  * - title
  *
  * @since 0.5
+ * @noinspection PropertiesInspection
  */
 class File extends FileCompat
 {
+    /**
+     * Categories are still experimental. Expect changes in v1.16
```
luke-, line 68: This comment  is very helpful in understanding where this should be going. But I would leave out the constants in the beta for now. e.g. VARIANT_1 - 8 is quite unclear for me.

protected/humhub/modules/file/migrations/m230618_135509_file_add_category_column.php
```diff
@@ -0,0 +1,75 @@
...
+            $this->table,
+            'category',
+            $this->integer(11)
+                 ->unsigned()
+                 ->after($after)
+        );
+
+        $this->safeCreateIndex("ix-$this->table-category", $this->table, ['category', 'object_model', 'object_id']);
+
+        $command->update(
```
luke-, line 48: I would leave this updates for now. Category should be `0` for the beginning.

luke- (changes requested):

protected/humhub/modules/file/models/File.php
martin-rueegg: Yes, the variants would need some explanation/documentation.

The idea was, that if we want to do something along the lines of bit-mask, then it was the idea to reserve the first 4 bits for variants of a given category.

E.e. if you have
- `00010000` = 16: Category A, you can have the following variants:
  - `10010000` = 17: Category A.x
  - `01010000` = 18: Category A.y
  - `11010000` = 19: Category A.z
  - ...
- `00001000` = 32: Category B, you can have the following variants:
  - `10001000` = 32: Category A.x
  - `01001000` = 34: Category A.y
  - `11001000` = 35: Category A.z
  - ...

Like so, the categories are close to each other in the decimal notation, and still usable with bit-mask.

PS: removed.

protected/humhub/modules/file/migrations/m230618_135509_file_add_category_column.php
martin-rueegg: Since the default value shall be `0`, I've made the field non-nullable.

luke- (approved):

PR #6592, opened 2023-09-23
Enh: `file` fields: `state`

martin-rueegg:
Adds `state` field to `File` table/model.

- Feature

- No

- [x] Changelog was modified

**Other information:**

Depends on/includes:
- #6590

### DIFF
--- a/protected/humhub/modules/file/models/File.php
+++ b/protected/humhub/modules/file/models/File.php
@@ -36,15 +36,15 @@ use yii\web\UploadedFile;
  * @property string $title
  * @property string $mime_type
  * @property string $size
- * @property string $object_model
- * @property integer $object_id
- * @property integer $content_id
- * @property string $created_at
- * @property integer $created_by
- * @property string $updated_at
- * @property integer $updated_by
- * @property integer $show_in_stream
- * @property string $hash_sha1
+ * @property string|null $object_model
+ * @property integer|null $object_id
+ * @property integer|null $content_id
+ * @property string|null $created_at
+ * @property integer|null $created_by
+ * @property string|null $updated_at
+ * @property integer|null $updated_by
+ * @property integer|null $show_in_stream
+ * @property string|null $hash_sha1
  *
  * @property StorageManager $store
  * @property FileHistory[] $historyFiles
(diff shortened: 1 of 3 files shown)


protected/humhub/modules/file/models/File.php
```diff
@@ -8,56 +8,73 @@
...
 class File extends FileCompat
 {
     /**
      * @event Event that is triggered after a new file content has been stored.
      */
-    const EVENT_AFTER_NEW_STORED_FILE = 'afterNewStoredFile';
+    public const EVENT_AFTER_NEW_STORED_FILE = 'afterNewStoredFile';
+
+    /**
+     * The numeric value of the published state is not yet finalized. Use with caution and expect a chenge of value in later versions.
```
luke-, line 72:
```suggestion
     * The numeric value of the published state is not yet finalized. Use with caution and expect a change of value in later versions.
```

protected/humhub/modules/file/models/File.php
```diff
@@ -89,9 +109,17 @@ public function rules()
...
+                ['mime_type'],
+                'match',
+                'not' => true,
+                'pattern' => '/[^a-zA-Z0-9\.ä\/\-\+]/',
+                'message' => Yii::t('FileModule.base', 'Invalid Mime-Type')
+            ],
             [['file_name', 'title'], 'string', 'max' => 255],
             [['size'], 'integer'],
+            [['state'], 'integer'], // ToDo: merge with size after all fields have been added
+            [['state'], 'unique', 'targetAttribute' => ['state', 'guid']],
```
luke-, line 122: Should it be possible to have multiple entries in the file table with the same GUID but different states?

luke- (changes requested):

protected/humhub/modules/file/models/File.php
martin-rueegg: I have not really looked at the code again, just divided it to the different fields. 😊

I guess this was not really thought through. The idea came from the possibility that a file might be in draft-state, before overwriting the published one. But then I guess later I solved that with having a draft variant instead...

luke- (approved):

PR #6593, opened 2023-09-23
Enh: `file` fields: `sort_order`

martin-rueegg:
Adds `sort_order` field to `File` table/model.

- Feature

- No

- [x] Changelog was modified

**Other information:**

Depends on/includes:
- #6590

### DIFF
--- a/protected/humhub/modules/file/models/File.php
+++ b/protected/humhub/modules/file/models/File.php
@@ -43,6 +43,7 @@ use yii\web\UploadedFile;
  * @property string|null $object_model
  * @property integer|null $object_id
  * @property integer|null $content_id
+ * @property integer $sort_order
  * @property string|null $created_at
  * @property integer|null $created_by
  * @property string|null $updated_at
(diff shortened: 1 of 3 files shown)


luke- (changes requested):
So far we always use `sort_order` as a free/non unqiue numeric field for individual sorting. If a value is used multiple times, it will be sorted by a second field like ID or filename.

I would stay on this approach for files as well for now.

I am aware that it also has disadvantages. But I want to stay consistent here for now. Especially so shortly before the release.

protected/humhub/modules/file/models/File.php
```diff
@@ -327,48 +389,108 @@ public function setStoredFile($file, $skipHistoryEntry = false)
...
+        /**
+         * Make sure the unique index ux-file-object is not violated. Hence, if sort_order is zero (0) set it to the PK.
+         *
+         * @param array|ArrayAccess $attributes
+         *
+         * @return void
+         * @throws ErrorException
+         * @see self::afterSave()
+         */
+    public function updateAttributes($attributes): void
```
luke-, line 476: I would remove this method. At least for now.

Currently in HumHub and Yii2  this method is quite low level and never executes any validations or events.

I'm unsure if we should start doing that or leave it as it is.

luke- (changes requested):
Thanks for the change. Please see my remark regarding `updateAll()`.

luke- (approved):

PR #6594, opened 2023-09-23
Enh: `file` fields: `metadata`

martin-rueegg:
Adds `metadata` field to `File` table/model.

- Feature

- No

- [x] Changelog was modified

**Other information:**

Depends on/includes:
- #6590

### DIFF
--- a/protected/humhub/libs/StdClass.php
+++ b/protected/humhub/libs/StdClass.php
@@ -1,0 +1,902 @@
+<?php
+
+/*
+ * @link      https://www.humhub.org/
+ * @copyright Copyright (c) 2023 HumHub GmbH & Co. KG
+ * @license   https://www.humhub.com/licences
+ */
+
+namespace humhub\libs;
+
+use __PHP_Incomplete_Class;
+use ArrayAccess;
+use ArrayIterator;
+use Countable;
+use humhub\exceptions\InvalidArgumentTypeException;
+use humhub\exceptions\InvalidArgumentValueException;
+use humhub\exceptions\InvalidConfigTypeException;
+use OutOfBoundsException;
+use ReflectionClass;
+use ReflectionException;
+use ReflectionObject;
+use ReflectionProperty;
+use SeekableIterator;
+use Serializable;
+use Stringable;
+use Traversable;
+use Yii;
+use yii\base\Arrayable;
+use yii\base\ArrayableTrait;
+use yii\helpers\ArrayHelper;
+
+/**
+ * This class provides an object that can have dynamic properties (like \stdClass) but with some additional features:
+ * - the properties can also be access in the array-like manner, \
+ *      i.e. `$object['property']` for `$object->property`
+ * - `count()` can be used to get the number of properties
+ * - the object can be cleared \
+ *      with `static:clear()` or `static::addValue(null)`
+ * - add multiple values in as an array \
+ *      with `static::addValue($property1 => $value1, $property2 => $value2, ...)`
+ * - add multiple array of value sets which will be merged first \
+ *      with `static::addValue([$property1 => $value1, ...], [$property2 => $value2, ...], ...)`
+ * - allows to verify if the entire object or a particular property has been modified \
+ *      with `static::isModified()` \
+ *      and `static::isFieldModified($field)`
+ * - allows to retrieve the (un)modified fields \
+ *      with `static::fieldsModified(true|false)` (`null` will return all fields)
+ * - the internal array pointer can be moved \
+ *      with `static::seek($pos)`
+ * - the object is stringable (which by default uses serialize)
+ * - provides a factory
+ *      with `\humhub\libs\StdClass::create()`
+ * - the serialization is optimized to
+ *      - reduce data by only exporting modified versions of self, and using pre-defined short property keys
+ *      - provide format versioning
+ *      - only allow deserialization of self and subclasses of self
+ *
+ * @see static::addValues()
+ * @see static::__serialize()
+ * @see static::unserialize()
+ */
+class StdClass extends \stdClass implements ArrayAccess, Stringable, SeekableIterator, Countable, Serializable, Arrayable
+{
+    use ArrayableTrait;
+
+    public const SERIALIZE_FORMAT = 1;
+
+    protected const SERIALIZE_VALUE__VERSION = 'v';
+    protected const SERIALIZE_VALUE__DATA = '_0';
+    /**
+     * List of extracted properties from the object to be unserialized.
+     * - For optional properties, just add the property name to the list
+     * - For *required* properties, use the following syntax: `[$propertyName => true]`
+     */
+    protected const UNSERIALIZE_REQUIRED_VALUES = [
+        self::SERIALIZE_VALUE__VERSION => true,
+        self::SERIALIZE_VALUE__DATA,
+    ];
+
+    /**
+     * @var \stdClass|null
+     */
+    protected static ?\stdClass $validatedObject = null;
+
+    /**
+     * @param array|traversable|string|null ...$args see class description
+     *
+     * @return static
+     * @throws InvalidConfigTypeException
+     * @see          \humhub\libs\StdClass
+     * @noinspection MagicMethodsValidityInspection
+     * @noinspection PhpUnnecessaryFullyQualifiedNameInspection
+     */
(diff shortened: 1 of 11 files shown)


protected/humhub/libs/StdClass.php
```diff
@@ -0,0 +1,412 @@
...
+use humhub\exceptions\InvalidArgumentValueException;
+use humhub\exceptions\InvalidArgumentTypeException;
+use OutOfBoundsException;
+use SeekableIterator;
+use Serializable;
+use Stringable;
+use Traversable;
+use yii\helpers\ArrayHelper;
+
+class StdClass extends \stdClass implements ArrayAccess, Stringable, SeekableIterator, Countable, Serializable
```
luke-, line 23: This class seems very powerful. What are the advantages to the regular `\stdClass`. Do we really need  these additional features as well? Are they used?

protected/humhub/libs/StdClassConfig.php
```diff
@@ -0,0 +1,26 @@
...
+ * @link      https://www.humhub.org/
+ * @copyright Copyright (c) 2023 HumHub GmbH & Co. KG
+ * @license   https://www.humhub.com/licences
+ */
+
+namespace humhub\libs;
+
+use ReflectionObject;
+
+class StdClassConfig
```
luke-, line 13: Is this used? How does it work?

protected/humhub/modules/file/models/File.php
```diff
@@ -8,56 +8,74 @@
...
- * Following properties are optional and for module depended use:
+ * Following properties are optional and for module-dependent use:
  * - title
  *
  * @since 0.5
+ * @noinspection PropertiesInspection
  */
 class File extends FileCompat
 {
+    public const WELL_KNOWN_METADATA_UPLOAD_HASH = 'file._upload.hash';
```
luke-, line 66: We should hold these `const` back for now.
Maybe it is useful to create an own `FileMetadata` class later, which then contains such constants.

protected/humhub/modules/file/models/File.php
```diff
@@ -112,12 +139,37 @@ public function behaviors()
     }
 
     /**
-     * Gets query for [[FileHistory]].
+     * @return array
+     * @noinspection PhpMissingReturnTypeInspection
+     * @noinspection ReturnTypeCanBeDeclaredInspection
+     */
+    public function transactions()
```
luke-, line 146: We are close to the last beta release and stable release. I don't want to enable transactions in this stage.

protected/humhub/modules/file/models/File.php
```diff
@@ -112,12 +139,37 @@ public function behaviors()
...
+        return [
+            'default' => self::OP_INSERT + self::OP_DELETE,
+        ];
+    }
+
+    public function __get($name)
+    {
+        $value = parent::__get($name);
+
+        if ($name === 'metadata' && !$value instanceof Metadata) {
```
luke-, line 157: What about using this here?
```php
if ($name === 'metadata') {
   return $this->getMetadata();
}
```

protected/humhub/modules/file/models/File.php
```diff
@@ -8,56 +8,74 @@
...
+ * @property integer|null $created_by
+ * @property string|null $updated_at
+ * @property integer|null $updated_by
+ * @property integer|null $show_in_stream
+ * @property string|null $hash_sha1
  *
- * @property \humhub\modules\user\models\User $createdBy
- * @property \humhub\modules\file\components\StorageManager $store
+ * @property StorageManager $store
  * @property FileHistory[] $historyFiles
```
luke-, line 53: We should also add the `metadata` property here.

protected/humhub/modules/file/models/File.php
```diff
@@ -240,24 +313,60 @@ public function canDelete($userId = null)
...
+
+        return $this->metadata;
+    }
+
+    /**
+     * @param string|array $metadata
+     *
+     * @return File
+     */
+    public function setMetadata($metadata): File
```
luke-, line 355: TBD

luke- (changes requested):
Some first thoughts

protected/humhub/modules/file/models/File.php
martin-rueegg: Absolutely. Got lost in all the cherry-picking, rebasing, and merging! 🤣

protected/humhub/modules/file/models/File.php
martin-rueegg: Yes, that works, too!

Your solution is certainly easier to understand. My solution was maybe minimally more efficient on successive calls ...

protected/humhub/modules/file/models/File.php
martin-rueegg: disabled in https://github.com/humhub/humhub/pull/6590/commits/3480969ae6a0410ae5aa5ac78f3eed6a0d64b98b. See https://github.com/humhub/humhub/pull/6590#discussion_r1335063469 for more info.

protected/humhub/modules/file/models/File.php
martin-rueegg: Its funny that I was thinking the same, but I had to dash ...

`Metadata` _is_ actually in the `File` model-specific. As such, I've moved the over there, but only in the follow-up PR that actually fills the metadata during upload:
- #6554

martin-rueegg:
I've added some documentation as why this is better than `\stdClass` (see below).

For me the main advantages are
- security during unserialization
- structured content

My suggestion is to require the top-level property of the metadata to be a module ID. The given model 
- needs to be enabled
- has to register the namespace
- can then define the structure of it's content, by providing a class that is used to instantiate it's property

E.g., as I further suggest, the `File` module would use the top-level properties to correspond with the given variant of a file. Such as
- `_org` [^1]
- `_upload`
- `_preview`
- etc ...
Each of these properties would result in an object that stores file-information, e.g.,
- hash
- mime-type (for variants)
- size
- maybe dimensions

[^1]: I'd like to change that to `_original` at a later point. I find `_org` very confusing as it is not a common abbreviation for "original", but for "organization".

This would result in a structure like

- `$metadata`
  - `file`
    - `_upload`
      - `hash`
      - `size`
      - `mime-type`
      - `original_filename`
    - `_draft` 
      - `hash`
      - `size`
      - `mime-type`
  - `onlyoffice`
    - `key` 


You can find a sample implementation [here](https://github.com/humhub/humhub/compare/develop...metaworx:humhub:enh/file-preliminary-fields-metadata-1?expand=1), a branch that builds on this PR.

--

Here I've added some documentation:

- https://github.com/humhub/humhub/blob/9a15bcc60833c6e98947a00c35c39e9d38d500f9/protected/humhub/libs/StdClass.php#L36-L59
- https://github.com/humhub/humhub/blob/9a15bcc60833c6e98947a00c35c39e9d38d500f9/protected/humhub/libs/StdClassConfigurable.php#L20-L27

protected/humhub/libs/StdClassConfig.php
martin-rueegg: In order to not occupy any property names, some additional data can be stored in an attached "config" object. This is accessible through the `static::config()` function.

It is now only available in the `StdClassConfigurable` extension of `StdClass`

It stores the default value (that will be returned if a non-existing property was read. It also allows to lock the property set (so no new properties can be added).

protected/humhub/modules/file/models/File.php
martin-rueegg: Alternatively, this could be removed entirely and make the property read-only.

- To set a value, one would need to do:
   ``` php
   $model->getMetadata()->property = 'value';
   $model->metadata->property = 'value';
   ```
   You find more examples here:
   https://github.com/humhub/humhub/blob/78df7d89e17ac95106fd7afa41ed326f6dd61df8/protected/humhub/modules/file/models/File.php#L51-L62
- With the current implementation of `setMetadata()`, the above could be like so:
   ``` php
   $model->setMetadata('property' => 'value');
   
   // this would delete all the data within the metadata object
   $model->setMetadata(null);
   $model->metadata = null;
   ```

luke- (approved):
@martin-rueegg Thanks for the PR. I merge it as discussed in the Beta.2 However, I'm not sure if the `StdClass*` is my favorite solution here, but lets discuss this later.

PR #6602, opened 2023-09-27
Enh: add image helper tests

martin-rueegg:
Adds a new test for the ImageHelper class


luke- (approved):

luke-:
@martin-rueegg Thanks, good point. From my side we can improve this helper in the next version.

PR #6626, opened 2023-10-24
Content state flag "was published"

yurabakhtin:

- [x] Feature

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
Issue: https://github.com/humhub/news/issues/67#issuecomment-1775259496

### DIFF
--- a/protected/humhub/modules/content/migrations/m231024_062218_content_was_published.php
+++ b/protected/humhub/modules/content/migrations/m231024_062218_content_was_published.php
@@ -1,0 +1,28 @@
+<?php
+
+use humhub\components\Migration;
+
+
+/**
+ * Class m231024_062218_content_was_published
+ */
+class m231024_062218_content_was_published extends Migration
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function safeUp()
+    {
+        $this->safeAddColumn('content', 'was_published',
+            $this->boolean()->defaultValue(false)->notNull()->after('state')
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function safeDown()
+    {
+        $this->safeDropColumn('content', 'was_published');
+    }
+}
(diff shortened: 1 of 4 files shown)


protected/humhub/modules/content/migrations/m231024_062218_content_was_published.php
```diff
@@ -0,0 +1,28 @@
...
+class m231024_062218_content_was_published extends Migration
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function safeUp()
+    {
+        $this->safeAddColumn('content', 'was_published',
+            $this->boolean()->defaultValue(false)->notNull()->after('state')
+        );
```
luke-, line 18: @yurabakhtin Don't we need a data migration here?

e.g. set all `what_published` to `1` where the current state is not `Draft` or `Schedule`?

luke- (changes requested):

protected/humhub/modules/content/migrations/m231024_062218_content_was_published.php
yurabakhtin: @luke- Firstly I too thought to do this, but then I decided to update the column `what_published` at the moment when state is updated from `Published` to another state, i.e. the `what_published=1` says us a Content had the `state=Published` in past time.

It works this way:

- Create Draft content, `state=draft`, `was_published=0`
- Update to Published, `state=published`, `was_published=0`
- Update to Draft, `state=draft`, `was_published=1` <= only at this step we should set `was_published` to `1`,

We should not to set `was_published` to `1` right at the time when we set `state=published` because some delayed/queued/cron code may works incorrectly when it will use a condition like:

```php
if (Content->isPublished() && Content->wasPublished()) { ... }
```

By other words with such way we can have one more additional variant (see 2nd variant below):

- `state!=published` and `was_published=0` => it is not published now and was not published before
- `state=published` and `was_published=0` => it is published now and was not published before (First publishing)
- `state=published` and `was_published=1` => it is published now and was published before (all Next publications)
- `state!=published` and `was_published=1` => it is not published now and was published before

yurabakhtin:
@luke- I see error here https://github.com/humhub/humhub/actions/runs/6623955002/job/17994190925?pr=6626#step:18:1489 but I don't know how to fix this, such error appears last days, only the "re-run" helps to make test greens:

test_error

luke- (approved):

protected/humhub/modules/content/migrations/m231024_062218_content_was_published.php
luke-: @yurabakhtin Thanks for the explaination & sounds good!

Issue #6678, opened 2023-11-25
Tour: Auto start test issue


luke-:
Seems the `AutoTourCest` sometimes fails, maybe because the first login is not always detected.  

https://github.com/humhub/humhub/blob/master/protected/humhub/modules/tour/Module.php#L40

For the "first login" check, I do not want any additional DB/Settings queries from the `tour` module.
Maybe we could do a check via the AFTER LOGIN event and then set a session variable, which is detected by the tour widget ?

yurabakhtin:
@luke- PR https://github.com/humhub/humhub/pull/6685 - I have done to use Session from dashboard event side as you suggested, but after activating the test `AutoTourCest` here https://github.com/humhub/humhub/pull/6685/commits/ac3059f18600620505213928fa87293fa9eb35ca it is broken again.
I will review deeply how to fix it.

yurabakhtin:
@luke- I have fixed the auto tour test in the commit https://github.com/humhub/humhub/pull/6685/commits/97710e05b738f1d3cc19e28bc0279287c28fa2bf: as you can see I just set `last_login` to `null`, because it is used now to detect new user after creating/registration.
Before I strarted to use the flag `last_login == null` the condition `updated_by` was used there, it is why the auto test was broken as well because as you can see the user fixtures data has `'updated_by' => '1'` for admin and for several other users.

PR #6249, opened 2023-04-20
Enh #29: AutoStart Tour for new Users

yuriimaz:
**What kind of change does this PR introduce?** (check at least one)

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

**Does this PR introduce a breaking change?** (check one)

- [ ] Yes
- [x] No

If yes, please describe the impact and migration path for existing applications:

**The PR fulfills these requirements:**

- [x] It's submitted to the `develop` branch, _not_ the `master` branch if no hotfix
- [ ] When resolving a specific issue, it's referenced in the PR's description (e.g. `Fix #xxx[,#xxx]`, where "xxx" is the Github issue number)
- [x] All tests are passing
- [ ] New/updated tests are included
- [ ] Changelog was modified

If adding a **new feature**, the PR's description includes:
- [ ] A convincing reason for adding this feature (to avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it)

**Other information:**
https://github.com/humhub/humhub-internal/issues/29

PR #6685, opened 2023-11-28
Refactor auto start tour for new user

yurabakhtin:

- [ ] Refactor

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
Issue: https://github.com/humhub/humhub/issues/6678


protected/humhub/modules/tour/Events.php
```diff
@@ -0,0 +1,78 @@
...
+
+    public static function onDashboardSidebarInit($event)
+    {
+        if (Yii::$app->user->isGuest) {
+            return;
+        }
+
+        $settings = self::getModule()->settings;
+
+        if ($settings->get('enable') == 1 && $settings->user()->get('hideTourPanel') != 1) {
```
luke-, line 31: Hmm, since we load the module user settings `$settings->user()->get('hideTourPanel')` here anyway, the optimization with the session is probably obsolete. :(

luke- (changes requested):

protected/humhub/modules/tour/Events.php
yurabakhtin: @luke- If I understand you correctly you want a solution what I have implemented in new commit https://github.com/humhub/humhub/pull/6685/commits/23f4e22d8dd5e6e6ebba9b3abd97d82d4205717d, so new Session var(to display the dashboard tour widget) is initialized on the event `onAfterLogin`.

But please note the new Session flag will be updated only after re-login user, so I think here we may have 2 issues:
- for all logged in user the dashboard sidebar tour widget will be hidden after apply this update,
- the widget visibility will not be refreshed after change the settings `settings->get('enable')` and `settings->user()->get('hideTourPanel')`

Do you think it is normal for the tour widget? Or we should find a solution how to refresh it for these 2 cases?

luke-: @yurabakhtin Not exactly, since we read out the user tour settings anyway, we can also replace the Login Event & Session solutionwith `$settings->user()->get('firstLoginPopupDisplayed')` here as well. 

However, we also need a solution so that the dialog does not appear for existing users.

yurabakhtin: @luke- Sorry, not clear what to change.

Just for info: before this update https://github.com/humhub/humhub/pull/6249/files#diff-7151ee0d1890e8e7a3812a19e6b4f0c4bb287b1f485bdbd103d50f13b13c1dacL36-L38 the event `onDashboardSidebarInit` is used only for displaying of the tour widget on the dashboard sidebar.

I want to understand when we should use `$settings->user()->get('firstLoginPopupDisplayed')`:
- probably you mean to set this user setting if the tour window was forced to be displayed after login, because currently there code `$user->updated_by === null && $user->created_at === $user->updated_at` is used for this.
- or please describe your idea if what I wrote above is wrong, thanks.

luke-: @yurabakhtin @luke- Sorry, I didn't fully thought fully of that myself.

How about the following:

1.) EVENT_AFTER_REGISTRATION sets `$tourSettings->user()->set('autoStartTour', 1)`
2.) onDashboardSidebarInit checks for `$tourSettings->user()->get('autoStartTour')` and starts tour and removes this setting.

yurabakhtin: @luke- Ok, I have replaced the Session var to `$tourSettings->user()->get('autoStartTour')` in commit https://github.com/humhub/humhub/pull/6685/commits/a389af0dde8a8d60d313e00446c0f39a548f0afd:

- I have found the column `user.last_login` is NULL for new created/registered user,
- so instead of the updating `user.updated_by` as it was used as flag before, I have decided to compare `$user->last_login === null` on event `onUserBeforeLogin`,
- but we still use the Session var to display the tour widget in dashboard sidebar, this session var is refreshed only after each login action, I mean the same issue described here https://github.com/humhub/humhub/pull/6685#discussion_r1408926312, i.e. currently we have 2 flags:
  - Auto start the tour (initialized before login only for new registered users)
  - Display dashboard sidebar widget (initialized after login by condition `$settings->get('enable') == 1 && $settings->user($user)->get('hideTourPanel') != 1`, as I understood we don't need to call this on each dashboard sidebar init and it is why we cached it in the Session var).

yurabakhtin: @luke- 

> Hmm, isn't `last_login` set on the first login? Sounds like a bug for me.

After registration or creating of new user the column `last_login` == `NULL`.
After each login it is filled with current date, it is why I have decided to use the column as flag of first logging even after user creating by admin and not only after registration (`Registration::EVENT_AFTER_REGISTRATION`)

> Could an intermediate Term&Condition page possibly interfere here?

I think no, I have tested this case right now, so current implementaion works this way:

- New user has `last_login = NULL`
- User is logged in first time
- Event `User::EVENT_BEFORE_LOGIN` is called, there we set `user->settings->set('autoStartTour', true)`
- Redirect to page "Terms and Conditions", two case:
  - Logout => the user setting `autoStartTour` is stored for next login action.
  - Accept => go do bashboard, if the dashboard sidebar widget is visible so event `onDashboardSidebarInit` is called:
    - if `autoStartTour === true` => run the auto tour
    - else display the Tour Widget in the Dashboard Sidebar, but I don't like that we use here the condition `Yii::$app->session->get(self::DASHBOARD_WIDGET)` to display it, this Session var is set on the event `AuthController::EVENT_AFTER_LOGIN`, i.e. the Tour Widget visibility will be updated per User only after each relogin, if you think it is ok, then we can keep this solution, otherwise we should improve it to refresh the Session somehow after update the settings which are used to set the Session var `self::DASHBOARD_WIDGET`.

yurabakhtin: > I'm not sure if I'm missing something here, but wouldn't following approach the easiest solution:
> 
> **Event Handler `Registration::EVENT_AFTER_REGISTRATION`:**
> 
> ```
>         /** @var \humhub\modules\tour\Module $tour */
>         $tour = Yii::$app->getModule('tour');
>         $tour->settings->user($event->identity)->set('doAutostart', 1);
> ```

@luke- As I wrote above I prefer to use event `User::EVENT_BEFORE_LOGIN` together with flag `last_login === NULL`, in order to don't skip case when user is creating from Administration or API.

> **Event Tour Widget:**
> 
> ```
>     public static function onDashboardSidebarInit($event)
>     {
>           /** @var Module $module */
>         $module = Yii::$app->getModule('tour');
> 
>         $doAutoStart = false;
>         if (!empty($module->settings->user()->get('doAutostart'))) {
>             $module->settings->user()->delete('doAutostart');
>             $doAutostart = true;
>         }
>         
>         // ... pass $doAutostart to the widget
>     }
> ```

Yes, it works this way.

> Regarding Session Cache: I don't think this is useful, because `$settings->user($user)->get('hide')` will load all user settings (incl. autoStart) for the tour module anyway.

Currently we store in Session the condition to display the Tour Widget on the Dashboard Sidebar:

tour-widget

I have done to store it in Session because as I understood here https://github.com/humhub/humhub/pull/6685#discussion_r1408444095 you don't like to use there data from database.
But I don't think it is good solution to use Session var there, so I have reverted that and prefer to use it as I have done now in the commit https://github.com/humhub/humhub/pull/6685/commits/4ecaad9b7e5a269518cf20552391d77afe0df0ac, do you agree this or please tell why we cannnot this solution?

protected/humhub/modules/tour/Events.php
```diff
@@ -0,0 +1,70 @@
...
+    {
+        return $user->last_login === null && // Force auto start only for new created user who is logged in first time after registration
+            self::shouldDisplayDashboardWidget($user) && // Start it only when the dashboard sidebar widget is visible for the user
+            !self::getModule()->showWelcomeWindow($user); // No need auto start because it will be done by dashboard widget
+    }
+
+    private static function runAutoStartWelcomeTour(): void
+    {
+        self::getModule()->settings->user()->delete(self::AUTO_START);
+        Yii::$app->response->redirect(['/dashboard/dashboard', 'tour' => true]);
```
luke-, line 68: @yurabakhtin Do we really need this extra redirect here?

luke- (changes requested):
@yurabakhtin Thanks, your current solution looks so good to me.

One last  point: Is it really necessary to redirect or reload the dashboard to start the tour?

protected/humhub/modules/tour/Events.php
yurabakhtin: @luke- I find the GET flag `tour` is used to start it - https://github.com/humhub/humhub/blob/fix/6678-tour-auto-start-test-issue/protected/humhub/modules/tour/widgets/Tour.php#L26-L29:

```php
        // Active tour flag not set
        if (!isset($_GET['tour'])) {
            return;
        }
```

If you look at the Tour Dashboard widget:

tour-widget

there all links have URL with param `?tour=1`.

Do you want to find a solution to keep all these 4 links without the flag `tour=1`?
Or do you want to keep the redirect e.g. on the event `onUserAfterLogin` instead of current `onDashboardSidebarInit`?

luke-: I would just like to get rid of the unnecessary redirect. 

1. login (first)
2. dashboard
3. dashboard?tour=1

--- 

Why not something like
https://github.com/humhub/humhub/blob/fix/6678-tour-auto-start-test-issue/protected/humhub/modules/tour/widgets/Tour.php#L26-L29

```
        // Active tour flag not set
        if (!isset($_GET['tour']) || !$this->doAutostart()) {
            return;
        }
```

yurabakhtin: @luke- Ok, I got it, thanks for the explanation. Done in commit https://github.com/humhub/humhub/pull/6685/commits/bef0799a4ae410c7b2b95a9bb2cbcabe6adf1f7a.

luke- (approved):

luke-:
@yurabakhtin Looks good, can you please take a look into tests?

yurabakhtin:
> @yurabakhtin Looks good, can you please take a look into tests?

@luke- Really I don't understand why the tests were broken again after the merging of `master`, because a fix was done in the previous commit where I set `last_login` to `null` for new users because this column is used as flag of new created/registered user.

I just find the warning `The following actions uses node12 which is deprecated and will be forced to run on node16: actions/checkout@v2, actions/setup-node@v1, actions/cache@v1. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/`, so I have updated the actions to `v3` in commit https://github.com/humhub/humhub/pull/6685/commits/dd3c56ad5a455bfd139fa8f561ec64512170a791 and now all tests are green but I don't know how it is related.

luke- (approved):

luke-:
@yurabakhtin Hmm, strange. Let's observe...

PR #6698, opened 2023-12-03
Fix: Content relations

martin-rueegg:
This PR addresses the circumstance, where the cached record in the polymorphic relationship of a content model does not correspond to the instance of the `ContentActiveRecord` that is accessing it's `content` property.

Furthermore, it address the fact, that the `ContentStateService::set()` method is calling a deprecated `setState()` method on it's content object, which then in turn again calls `ContentStateService::set()`. The endless loop is only prevented thanks to PHP avoiding calling the same `__set($name)` call recursively. So the loop happens only once.

Also v1.14 is affected by this.

## PR Admin

Could this still be added to v1.15 or even 1.14? Happy to rebase, if yes.

- Bugfix

- No

###The PR fulfills these requirements

- [ ] All [tests](#issuecomment-new) are passing
- [ ] Changelog was modified

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ HumHub Changelog
 
 1.15.1 (Unreleased)
 --------------------------------
+- Fix #6698: Content relations
 - Fix #6644: Fix push service
 - Fix #6645: File dropdown not visible at the bottom of the page
 - Fix #6639: Apply image inline styles in email message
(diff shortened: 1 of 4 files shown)


luke- (approved):

luke-:
@martin-rueegg Thanks, I'll think a fix for `v1.15+` is fine.

PR #6711, opened 2023-12-08
Enh: run migrations manually

martin-rueegg:
Currently, when navigating to Administration -> Information -> Database, any pending migration is applied immediately and leaves the administrator no choice of whether to actually apply them now.

- This PR loads and displays the pending migration and gives the user the option to apply them.
- Once migrations are applied, the user can reload the migration info (useful, if some migrations failed) and apply them again.
- If no migration is applied, no option to apply them is offered.
- If migrations are applied, cache is automatically cleared as well.

- Feature

- No

- [x] All tests are passing
- [x] Changelog was modified

### DIFF
--- a/protected/humhub/modules/admin/controllers/InformationController.php
+++ b/protected/humhub/modules/admin/controllers/InformationController.php
@@ -8,9 +8,12 @@
 
 namespace humhub\modules\admin\controllers;
 
+use humhub\commands\MigrateController;
+use humhub\libs\SelfTest;
 use humhub\modules\admin\components\Controller;
 use humhub\modules\admin\components\DatabaseInfo;
 use humhub\modules\admin\libs\HumHubAPI;
+use humhub\modules\admin\permissions\SeeAdminInformation;
 use humhub\modules\queue\driver\MySQL;
 use humhub\modules\queue\helpers\QueueHelper;
 use humhub\modules\queue\interfaces\QueueInfoInterface;
@@ -20,12 +23,15 @@ use ReflectionException;
 use Yii;
 
 /**
- * Informations
+ * Information
  *
  * @since 0.5
  */
 class InformationController extends Controller
 {
+    public const DB_ACTION_CHECK = 0;
+    public const DB_ACTION_RUN = 1;
+    public const DB_ACTION_PENDING = 2;
 
     /**
      * @inheritdoc
@@ -52,7 +58,7 @@ class InformationController extends Controller
     protected function getAccessRules()
     {
         return [
-            ['permissions' => \humhub\modules\admin\permissions\SeeAdminInformation::class],
+            ['permissions' => SeeAdminInformation::class],
         ];
     }
 
@@ -78,11 +84,25 @@ class InformationController extends Controller
 
     public function actionPrerequisites()
     {
-        return $this->render('prerequisites', ['checks' => \humhub\libs\SelfTest::getResults()]);
+        return $this->render('prerequisites', ['checks' => SelfTest::getResults()]);
     }
 
-    public function actionDatabase()
+    public function actionDatabase(int $migrate = self::DB_ACTION_CHECK)
     {
+        if ($migrate === self::DB_ACTION_RUN) {
+            $migrationOutput = sprintf(
+                "%s\n%s",
+                MigrateController::webMigrateAll(),
+                SettingController::flushCache()
+            );
+        } else {
+            $migrationOutput = MigrateController::webMigrateAll(MigrateController::MIGRATION_ACTION_NEW);
+
+            $migrate = str_contains($migrationOutput, 'No new migrations found.')
+                ? self::DB_ACTION_CHECK
+                : self::DB_ACTION_PENDING;
+        }
+
         $databaseInfo = new DatabaseInfo(Yii::$app->db->dsn);
 
         $rebuildSearchJob = new RebuildIndex();
@@ -95,7 +115,8 @@ class InformationController extends Controller
             [
                 'rebuildSearchRunning' => QueueHelper::isQueued($rebuildSearchJob),
                 'databaseName' => $databaseInfo->getDatabaseName(),
-                'migrate' => \humhub\commands\MigrateController::webMigrateAll(),
+                'migrationOutput' => $migrationOutput,
+                'migrationStatus' => $migrate,
             ]
         );
     }
@@ -105,8 +126,8 @@ class InformationController extends Controller
      */
     public function actionBackgroundJobs()
     {
-        $lastRunHourly = (int) Yii::$app->settings->getUncached('cronLastHourlyRun');
-        $lastRunDaily = (int) Yii::$app->settings->getUncached('cronLastDailyRun');
+        $lastRunHourly = (int)Yii::$app->settings->getUncached('cronLastHourlyRun');
+        $lastRunDaily = (int)Yii::$app->settings->getUncached('cronLastDailyRun');
 
         $queue = Yii::$app->queue;
 
@@ -115,7 +136,10 @@ class InformationController extends Controller
             $canClearQueue = true;
             if (Yii::$app->request->isPost && Yii::$app->request->get('clearQueue') == 1) {
                 $queue->clear();
-                $this->view->setStatusMessage('success', Yii::t('AdminModule.information', 'Queue successfully cleared.'));
+                $this->view->setStatusMessage(
+                    'success',
+                    Yii::t('AdminModule.information', 'Queue successfully cleared.')
+                );
             }
         }
 
@@ -137,7 +161,7 @@ class InformationController extends Controller
             $reflect = new ReflectionClass($queue);
             $driverName = $reflect->getShortName();
         } catch (ReflectionException $e) {
-            Yii::error('Could not determine queue driver: '. $e->getMessage());
+            Yii::error('Could not determine queue driver: ' . $e->getMessage());
         }
 
         return $this->render('background-jobs', [
@@ -151,5 +175,4 @@ class InformationController extends Controller
             'canClearQueue' => $canClearQueue
         ]);
     }
-
 }
--- a/protected/humhub/modules/admin/views/information/database.php
+++ b/protected/humhub/modules/admin/views/information/database.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * @link https://www.humhub.org/
  * @copyright Copyright (c) 2017 HumHub GmbH & Co. KG
@@ -6,13 +7,17 @@
  */
 
 use humhub\libs\Html;
+use humhub\modules\admin\controllers\InformationController;
+use humhub\modules\ui\view\components\View;
 
 /**
- * @var $this \humhub\modules\ui\view\components\View
+ * @var $this View
  * @var $databaseName string
- * @var $migrate string
+ * @var $migrationStatus int
+ * @var $migrationOutput string
  * @var $rebuildSearchRunning boolean
  */
+
 ?>
 <div>
     <p>
@@ -27,10 +32,42 @@ use humhub\libs\Html;
     </p>
 </div>
 
-<p><?= Yii::t('AdminModule.information', 'Database migration results:'); ?></p>
-
-<div class="well">
+<div>
+<?php if ($migrationStatus === InformationController::DB_ACTION_PENDING): ?>
+    <p><?= Yii::t('AdminModule.information', 'Outstanding database migrations:'); ?></p>
+    <div class="well">
     <pre>
-        <?= $migrate; ?>
+        <?= $migrationOutput ?>
     </pre>
+    </div>
+    <p><br>
+        <?= Html::a(
+            Yii::t('AdminModule.information', 'Refresh'),
+            ['/admin/information/database'],
+            [
+                'id' => 'migrationRun',
+                'class' => 'btn btn-primary pull-right',
+            ]
+        ); ?>
+    </p>
+<?php elseif ($migrationStatus === InformationController::DB_ACTION_RUN): ?>
+    <p><?= Yii::t('AdminModule.information', 'Database migration results:'); ?></p>
+    <div class="well">
+    <pre>
+        <?= $migrationOutput ?>
+    </pre>
+    </div>
+    <p><br>
+        <?= Html::a(
+            Yii::t('AdminModule.information', 'Update Database'),
+            ['/admin/information/database', 'migrate' => 1],
+            [
+                'id' => 'migrationRun',
+                'class' => 'btn btn-primary pull-right',
+            ]
+        ); ?>
+    </p>
+<?php else: ?>
+    <p><?= Yii::t('AdminModule.information', 'Your database is <b>up-to-date</b>.'); ?></p>
+<?php endif; ?>
 </div>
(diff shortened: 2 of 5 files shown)


martin-rueegg:
All failing tests have to do with
- `http://localhost/` vs
- `http://localhost:8080/`

I have no clue how this could be related to the changes in this PR.

martin-rueegg:
> All failing tests have to do with
> 
>     * `http://localhost/` vs
> 
>     * `http://localhost:8080/`
> 
> 
> I have no clue how this could be related to the changes in this PR.

@luke- Could that have to do with your merges from master to develop?

luke- (changes requested):
@martin-rueegg Thanks, below my review.

protected/humhub/modules/admin/controllers/InformationController.php
```diff
@@ -78,15 +82,73 @@ public function actionAbout()
...
+            Yii::t('AdminModule.information', 'Update migration information'),
+            ['/admin/information/database', 'migration' => 'check'],
+            $linkOptions
+        );
+
+        if ($request->isPjax && $migrationCheck = $request->get('migration')) {
+            if ($migrationCheck === 'run') {
+                $return = MigrateController::webMigrateAll();
+
+                $return .= "\n" . "Flushing cache ...";
```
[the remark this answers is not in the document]
martin-rueegg: Not sure I understand your sentence.

 You'd prefer
```php
 $return .= "\nFlushing cache ...";
```
Or heredoc or something different?

protected/humhub/modules/admin/views/information/database.php
```diff
@@ -1,36 +1,64 @@
...
 
 <p><?= Yii::t('AdminModule.information', 'Database migration results:'); ?></p>
 
-<div class="well">
-    <pre>
-        <?= $migrate; ?>
-    </pre>
+<div>
+    <?php
+    Pjax::begin([
```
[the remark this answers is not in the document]
martin-rueegg: Do I understand correctly, that you want to get rid of the entire PJAX in this PR?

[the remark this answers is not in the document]
luke-: Yes, currently we only use PJAX to avoid reloading of content container layouts or the whole page layout. And some users disabled this. (Don't know why). I don't want to introduce such a new approach like this by the way. 

For me, it is important that everything is implemented as identically as possible. Anything else makes hard to maintain.

I would like to discuss & decide this separately, if we should se PJAX in more places in the core/modules.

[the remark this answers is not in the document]
martin-rueegg: > Yes, currently we only use PJAX to avoid reloading of content container layouts or the whole page layout. And some users disabled this. (Don't know why). I don't want to introduce such a new approach like this by the way.

Ok, I understand.

The reason why I chose PJAX was twofold:
1. Page loading speed/reactivity
2. Reducing the code loaded before migrations are happening

The first one is a general point. As a user I'd expect the page to move somehow once I click on a link. Here, nothing seems to happen (with PJAX enabled) until the request returns and the page is replaced. Hence, loading the page content quickly and then the migration (check) provides more responsiveness. However, this is just a detail in user experience.

The second point ties in with the topic of upgrading modules: The less code is loaded before the migration is happening, the less likely some code fails (e.g. because a field is not yet there). And the fact that some migration has not been a applied yet is an indication that something went wrong already before this point, as otherwise migrations would have been applied during installation/update.

As such, the PJAX sub-request only loads the controller and the view and runs the migration. There is no checking on any other part. And in an emergency, the link could also be called directly from the browser's address line (e.g. in a support case) to run the migration (if shell access isn't easy available to run it there). And as such some issues could be solved.

Now this just as a background for you. I'm happy to convert the code into plain page calls. Or, alternatively, in [JavaScript client](https://docs.humhub.org/docs/develop/javascript-client) calls. Would that be any better than PJAX? (Under the hood, PJAX is just hitchhiking the client xhr request. So I'm not sure if those who disable PJAX also disable any xhr.)

> For me, it is important that everything is implemented as identically as possible. Anything else makes hard to maintain.

Sure, got you.

protected/humhub/modules/admin/controllers/InformationController.php
```diff
@@ -78,15 +83,42 @@ public function actionAbout()
...
+
+        if ($migrations === self::DB_ACTION_RUN) {
+            $migrationOutput = sprintf(
+                "%s\n%s",
+                MigrateController::webMigrateAll(),
+                SettingController::flushCache()
+            );
+
+            $linkRunMigrations = [
+                'caption' => Yii::t('AdminModule.information', 'Update migration information'),
```
luke-, line 102: Do we really need to have different captions/buttons/action here?

I see only two situations:

- There are new migrations to execute. --> Show button
- There are no migration. --> No button

luke- (changes requested):

protected/humhub/modules/admin/views/information/database.php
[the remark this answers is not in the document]
luke-: @martin-rueegg Thanks for the arguments and the changes. Perhaps we should really push for greater use of PJAX on the roadmap.

protected/humhub/modules/admin/controllers/InformationController.php
martin-rueegg: There is a semi-third state:

-  migrations are run and the result displayed in the window. So the "Update migration information" is show to re-check. would be the same as reloading the page.

Is kinda verifying that everything is fine now. Or after a migration failed to see what is now actually pending.

luke- (changes requested):

protected/humhub/modules/admin/controllers/InformationController.php
luke-: > > But then, why not simply add a "Reload" button to the view on condition `!empty($migrationOutput)`.
> 
> Actually, `$migrationOutput` will never be empty, as "empty" is when it says "no migrations". But I get your point.
> 
> I'm not quite sure, why "reload" would be more indicative than "Update migration information". I find it a bit ambiguois, as it could also mean "reload the migrations".
> 
> How about "refresh"?

"Refresh" would be also fine for me. We could also do something like `str_replace('No new migrations!', '')` on controller level. And on view: 

```
<?php if (empty($migrationOutput)): ?>

Some more user friendly and translated message, that the database is up 2 date.

<?php endif; ?> 
```

> 
> > Why do we need `DB_ACTION_CHECK` constants and controller logic? Are these perhaps leftovers from the Ajax solution?
> 
> I have used constants, as I did not want to just use free-text strings. But now that we have abolished PJAX (which needed to distinguish three states: a) primary load, b) migration check, and c) migration run) we do only have to distinguish two states (load with check vs. load with run). So we could do it similar to the refresh search thingy and just use `1`, if you prefer.
> 
> > controller logic
> 
> We need that anyway to distinguish the two requests and run different migration commands (`up` vs `new`)

Please see the view file below, for an idea, of controller logic. Did I missed something?

protected/humhub/modules/admin/views/information/database.php
```diff
@@ -31,6 +48,7 @@
 
 <div class="well">
```
luke-, line 49: Hmm, instead of preparing the button on the controller side, I would prefer to add some vars, to get a simple and readable view e.g. 

```php
<?php if ($newMigrationsPerformed): ?>
       Database Migration Result:
      <pre>
        <?= $migrationOutput ?>
      </pre>
      <?= Html::a('Update Database' , ['/admin/information/database'], ['class' => 'btn btn-default']); ?>
<?php elseif ($newMigrationsAvailable): ?>
      <pre>
        <?= $migrationOutput ?>
      </pre>
      <?= Html::a('Refresh' , ['/admin/information/database', 'migrate' => 1], ['class' => 'btn btn-default']); ?>
<?php else: ?>
      <!-- Show your database is up2date message -->
<?php endif; ?>
```

luke- (changes requested):

martin-rueegg:
@luke- Thank for the review. I hope I have now been able to adjust accordingly.

protected/humhub/modules/admin/controllers/InformationController.php
```diff
@@ -78,15 +84,32 @@ public function actionAbout()
...
+            $migrate = str_contains($migrationOutput, 'No new migrations found.')
+                ? self::DB_ACTION_CHECK
+                : self::DB_ACTION_PENDING;
+        }
+
         $databaseInfo = new DatabaseInfo(Yii::$app->db->dsn);
 
         $rebuildSearchJob = new RebuildIndex();
-        if (Yii::$app->request->isPost && Yii::$app->request->get('rebuildSearch') == 1) {
+        if ($request->isPost && $request->get('rebuildSearch') == 1) {
```
luke-, line 112: Just a small note: I understand that we often have very deep wiring that is a bit hard to read, but in the case of `Yii::$app->request` or other app components I would like to keep it that way. :-)

luke- (approved):

protected/humhub/modules/admin/controllers/InformationController.php
martin-rueegg: If `Yii::$app->request` was an actual class field, I'd understand. And I agree that in this case line 93 could be at 112 to keep it more readable.

But what is be the benefit of going through the whole get-component thing twice, just to read two properties of the very same object? Isn't that a bunch of function calls we can safe?

I appreciate that this should not be done if the object might change in between, e.g. `Yii::$app->controller` but in this case I just make to successive calls to `isPost` and `get`.

Shall I undo the change or move the $request asignment to line 112?

luke-: I would like to undo this, I don't want to be too picky.

I believe the Yii2 ServiceLocator is quite efficient: 
https://github.com/yiisoft/yii2/blob/master/framework/di/ServiceLocator.php#L127-L129

martin-rueegg: done.

martin-rueegg: > I believe the Yii2 ServiceLocator is quite efficient:
> https://github.com/yiisoft/yii2/blob/master/framework/di/ServiceLocator.php#L127-L129

Btw, internally it's more something like:
1. Application: property_exists('request')
3. Application: get_parent_class()
4. WebApplication: property_exists('request')
3. WebApplication: get_parent_class()
4. BaseApplication: property_exists('request')
3. BaseApplication: get_parent_class()
4. Module: property_exists('request')
3. Module: get_parent_class()
4. ServiceLocator: property_exists('request')
3. ServiceLocator: get_parent_class()
4. Component: property_exists('request')
3. Component: get_parent_class()
4. BaseObject: property_exists('request')
3. BaseObject: get_parent_class()
5. Application: method_exists('__get')
3. Application: get_parent_class()
4. WebApplication: method_exists('__get')
3. WebApplication: get_parent_class()
4. BaseApplication: method_exists('__get')
3. BaseApplication: get_parent_class()
4. Module: method_exists('__get')
3. Module: get_parent_class()
4. ServiceLocator: method_exists('__get')
4. ServiceLocator::__get('request')
6. Application: method_exists('has')
3. Application: get_parent_class()
4. WebApplication: method_exists('has')
3. WebApplication: get_parent_class()
4. BaseApplication: method_exists('has')
3. BaseApplication: get_parent_class()
4. Module: method_exists('has')
3. Module: get_parent_class()
4. ServiceLocator: method_exists('has')
5. isset($this->_components['has']) with array lookup
6. return true
6. Application: method_exists('get')
3. Application: get_parent_class()
4. WebApplication: method_exists('get')
3. WebApplication: get_parent_class()
4. BaseApplication: method_exists('get')
3. BaseApplication: get_parent_class()
4. Module: method_exists('get')
3. Module: get_parent_class()
4. ServiceLocator: method_exists('get')
5. ServiceLocator::get('request')
5. isset($this->_components['has']) with array lookup
5. return $this->_components['has'] with array lookup

It may be though, that the first location of the property (steps 1-14) are more efficient for an instantiated object. I have not checked the source code. But if it's similar as reflection works, that would basically be the procedure.

And all of that twice in the current method. Only once, if it's not a post request. :-)

Using `Yii::$app->get('request')` would already be massively better. But then we loose autocomplete.

So I certainly think, storing the variable is way more efficient. Particularly if you do it many times in the same scope, and then on many levels ...

luke-: Thanks for the overview, dood to know, indeed, it doesn't look very efficient. Hope there are some Optimizations in the background. :-)

But for me and for now, the important thing is to maintain a consistent style. But for sure there is much room for improvements in future

luke- (approved):

yurabakhtin:
@luke- @martin-rueegg Probably the bug is related somehow:

migration-bug

martin-rueegg:
@yurabakhtin would you be able to describe the procedure?

Does the "not secure" info in the addressbar show up before or after you click "Update Database"?

I'd suggest that the problem may be a http vs https incongruency.

yurabakhtin:
@martin-rueegg I just started to work on new branch https://github.com/humhub/humhub/tree/enh/search where we have new migration `m240203_112155_search` so after run the update I got the errors, and even after several page refreshes I see the errors.

> Does the "not secure" info in the addressbar show up before or after you click "Update Database"?

My local test server works on `http://`

local-site

I can reproduce the broken layout even with simple opening the url like `http://humhub.local/admin/information/database?migrate=1` in new browser tab.
If you don't see this then don't spend your time, maybe it is only from my browser side, I will investigate this later deeply.
I just posted this here because it looks more related and in order to don't forget to fix this, thanks.

martin-rueegg:
Ah, I see. So I guess it has nothing to do with the "not secure" warning.
Also, looking at this PR I don't see how this could be the causing an error. Unless the MigrateController would somehow force https, rather than http in the `up` action.

I can't reproduce it here with https. And when I use http, it automatically switches to https (browser and server side). So I'd have to change my setup.

But you could print a screenshot of the requests sent (visible in the network tab of the developer console of the browser). I'm sure there are some css (and maybe other) files that are not correctly loaded.

yurabakhtin:
@martin-rueegg 

- `/admin/information/database?migrate=1`

files-404

strange why the files are not loaded, but if I copy their url into a new browser tab, I see the files are loaded correctly:

fontawesome
bootstrap

- `/admin/information/database`

no-migrate-param

I see only one difference in the param `v`: `v=1.16.0` vs `v=1707498088`

I have this bug in Chrome and FireFox browsers.

The branch `master` v1.15.3 has no such issue, only `develop` v1.16.0

martin-rueegg:
@yurabakhtin This is indeed very peculiar.

Would you mind pasting the source code of the two main pages here? I.e. the raw response from the following requests:

- /admin/information/database?migrate=1
- /admin/information/database

And do you have a chance to see if the requests actually hit the webserver?
Is the cache disabled in the chrome browser?

yurabakhtin:
@martin-rueegg I have detected the problem was in the code `Yii::$app->assetManager->clear()` which is called from `SettingController::flushCache()`, on my side the code removes all folders from the folder `/assets` so it seems some of them cannot be recreated when page is loaded right after the clearing.
I have decided to redirect a page with storing a migration result in session, please check my solution in PR https://github.com/humhub/humhub/pull/6881.

martin-rueegg:
Thank you, @yurabakhtin. See my further comments in the respective PR.

PR #6713, opened 2023-12-08
Keep string class option for Pjax Widget

gevorgmansuryan:
Keep string class option for Pjax Widget


martin-rueegg:
Nice solution, thank @gevorgmansuryan.

However, it should be merged as a fix into `master`, shouldn't it?

gevorgmansuryan:
@martin-rueegg yes it will be merged to master with 1.16.0 release

martin-rueegg:
It is a bug, but a minor one.

martin-rueegg:
> I'm not deep enough into PJAX topic right now to be able to assess what the implications are here. That's why I prefer `develop`.

It's less related to PJAX than to simple widget creation, or more specific, some `ArrayHelper::merge()` behavior as laid out here:
- https://github.com/humhub/humhub/pull/6277#discussion_r1419548525

Here an illustration of the bug:

```php
// classes defined with array notation
\humhub\widgets\Pjax::widget(['options' => ['class' => /* array here -> */ ['oneClass', 'anotherClass']]);
```
will correctly produce:
```html
<div class="oneClass anotherClass exclude-from-pjax-client"></div>
```

However, without the fix in this PR, in v1.15, the following would fail
```php
// classes defined with string notation
\humhub\widgets\Pjax::widget(['options' => ['class' => /* string here -> */ 'oneClass anotherClass']);
```
as the custem classes would simply be left out:
```html
<div class="exclude-from-pjax-client"></div>
```


Unfortunately, this code not covered by tests. But then most likely no one is using that code anyway. Certainly it is not used in the core. But since this "feature" has been introduced in 1.15, I'd to also fix it in 1.15.

That's my 2 cents. It is your call.

luke-:
@martin-rueegg Thanks for the explanation. I haven't seen any core uses of this class either. Only in the `news` module I found something.

If you need the fix in the `master` then we can do a rebase. Otherwise I would leave it at `develop`.

ArchBlood:
> > I'm not deep enough into PJAX topic right now to be able to assess what the implications are here. That's why I prefer `develop`.
> 
> It's less related to PJAX than to simple widget creation, or more specific, some `ArrayHelper::merge()` behavior as laid out here:
> 
> * [Enh/bring back pjax #6277 (comment)](https://github.com/humhub/humhub/pull/6277#discussion_r1419548525)
> 
> Here an illustration of the bug:
> 
> ```
> // classes defined with array notation
> \humhub\widgets\Pjax::widget(['options' => ['class' => /* array here -> */ ['oneClass', 'anotherClass']]);
> ```
> 
> will correctly produce:
> 
> ```
> <div class="oneClass anotherClass exclude-from-pjax-client"></div>
> ```
> 
> However, without the fix in this PR, in v1.15, the following would fail
> 
> ```
> // classes defined with string notation
> \humhub\widgets\Pjax::widget(['options' => ['class' => /* string here -> */ 'oneClass anotherClass']);
> ```
> 
> as the custem classes would simply be left out:
> 
> ```
> <div class="exclude-from-pjax-client"></div>
> ```
> 
> Unfortunately, this code not covered by tests. But then most likely no one is using that code anyway. Certainly it is not used in the core. But since this "feature" has been introduced in 1.15, I'd to also fix it in 1.15.
> 
> That's my 2 cents. It is your call.

I'd also like to agree here, currently the discussions module uses the Yii2 Pjax widget due to issues like this, also if the Pjax class was reintroduced in v1.15 and this is a bug fix for the HumHub extended class it should be rebased for `master`, but again this is up to the HumHub team, but would be very much appreciated for the next minor release. 🤔

martin-rueegg:
> If you need the fix in the `master` then we can do a rebase. Otherwise I would leave it at `develop`.

I personally do not _need_ it in master, as I know how to work around it. However
- I don't quite understand the resistance to rebase it to master:
  - I appreciate you want to have things tested rigorously before they go into master. However, this is clearly a bugfix, not a feature. So _not_ rebasing it to master keeps the buggy version in production.
  - Apparently the buggy feature was introduces without anyone noticing or reporting the bug. So how would the fixed version cause more problems than the buggy version?
  - Also, the fix only adds an array value type conversion for the key `class`, which prevents `ArrayHelper::merge()` from _overwriting_ a per-existing string value, but actually _merges_ them.
- It would make module-devs' life easier if the fix is merged in the version the bug was introduced. Because it reduces the number of cases we have to account for. If you merge it into 1.16, in the module we must test for
  - version `<1.15`: use my workaround as per wiki documentation
  - version `=1.15`: use half-baked solution, but first make sure class attribute is provided as an array
  - version `>1.15`: use fixed solution


Besides that, I think the code could have been written in a way that is easier to read and more efficient - as it does not need to merge three arrays:

```php
$class = (array)ArrayHelper::getValue($this->options, 'class', []);
$class[] = 'exclude-from-pjax-client';
$this->options['class'] = $class;
```

(All we really want it to do is to _add_ the `exclude-from-pjax-client` class to the widget's element, _without replacing_ any existing class provided by the widget configuration.)

yurabakhtin:
@luke- I don't think we need the change of this PR.

I tested with code:

```php
\humhub\widgets\Pjax::widget(['options' => ['class' => /* array here -> */ ['oneClass', 'anotherClass']]]);
```

1. Current PR's code returns an array as:

```php
$this->options['class'] = [
  0 => 'oneClass',
  1 => 'anotherClass',
  2 => 'oneClass',
  3 => 'anotherClass',
  4 => 'exclude-from-pjax-client'
];
```

2. If I revert the change to:

```php
$this->options = ArrayHelper::merge($this->options, [
    'class' => ['exclude-from-pjax-client'],
]);
```

I get expeced result:

```php
$this->options['class'] = [
  0 => 'oneClass',
  1 => 'anotherClass',
  3 => 'exclude-from-pjax-client'
];
```

3. Also the same correct result can be returned by suggested code here https://github.com/humhub/humhub/pull/6713#issuecomment-1855389596:

```php
$class = (array)ArrayHelper::getValue($this->options, 'class', []);
$class[] = 'exclude-from-pjax-client';
$this->options['class'] = $class;
```

result:

```php
$this->options['class'] = [
  0 => 'oneClass',
  1 => 'anotherClass',
  3 => 'exclude-from-pjax-client'
];
```

So we should either don't merge this PR or use the suggested code by @martin-rueegg.

martin-rueegg:
@yurabakhtin thanks for looking into this.

> I tested with code:
> 
> ```
> \humhub\widgets\Pjax::widget(['options' => ['class' => /* array here -> */ ['oneClass', 'anotherClass']]]);
> ```

This is not the problematic code. Array notation is the one that already works.

But this one would fail:

```php
// classes defined with string notation
\humhub\widgets\Pjax::widget(['options' => ['class' => /* string here -> */ 'oneClass anotherClass']);
```

The string `"oneClass anotherClass"` is expected to be overwritten without the correction. Due to the follwing line:
https://github.com/yiisoft/yii2/blob/d43341a73ab9ab711c42b74c117e076275d9b8c0/framework/helpers/BaseArrayHelper.php#L137:
```php
is_array($v) && isset($res[$k]) && is_array($res[$k])
```
because `$v` is an array, but the existing entry `$res[$k]` is a string. Hence in Line 140 the value is overwritten.

yurabakhtin:
> This is not the problematic code. Array notation is the one that already works.
> 
> But this one would fail:

@martin-rueegg Thank you for explanation, then we should use your approach:

```php
$class = (array)ArrayHelper::getValue($this->options, 'class', []);
$class[] = 'exclude-from-pjax-client';
$this->options['class'] = $class;
```

because it works well on both cases and it doesn't create duplicates.

protected/humhub/widgets/Pjax.php
```diff
@@ -11,6 +11,8 @@ class Pjax extends \yii\widgets\Pjax
     public function init()
     {
         $this->options = ArrayHelper::merge($this->options, [
+            'class' => (array) ArrayHelper::getValue($this->options, 'class', [])
+        ], [
```
yurabakhtin, line 15: We should use a suggested code from here https://github.com/humhub/humhub/pull/6713#issuecomment-1855389596:

```php
$class = (array)ArrayHelper::getValue($this->options, 'class', []);
$class[] = 'exclude-from-pjax-client';
$this->options['class'] = $class;
```

yurabakhtin (changes requested):

protected/humhub/widgets/Pjax.php
gevorgmansuryan: @yurabakhtin why? both are doing the same thing

martin-rueegg: @gevorgmansuryan if you'd read the conversation above, you'd see that 
a) your code creates duplicate entries (https://github.com/humhub/humhub/pull/6713#issuecomment-1855619202)
b) does unnecessary array creation and merges and hence isn't very efficient (https://github.com/humhub/humhub/pull/6713#issuecomment-1855389596)

Of course only if you replace the entire code (Lines 13-17).

Personal remark: Sometimes its worth double-checking if someone else spotted a mistake. I already pointed it out a few times before you actually looked at it. Then you made a quick-shot-fix without due diligence. And now again you just assume your code is correct. However, it's not ... Just a gentle reminder.

martin-rueegg (commented):
Yes. I'd still rebase it to `master`, but that's @luke-'s call

protected/humhub/widgets/Pjax.php
gevorgmansuryan: @martin-rueegg Thanks for remarks :)

yurabakhtin (approved):

martin-rueegg:
@luke- Would it be possible to merge this?

Again, preferably into `master`, given it's a bug-fix. But at least into `develop` if `master` is out of question.

Thanks!

luke- (approved):

luke-:
@martin-rueegg Since this PR is against `develop` I've merged to it. Feel free to cherry-pick also to `master`  if required.

Issue #6714, opened 2023-12-09
`develop` tests fail


martin-rueegg:
For some reasons, the tests fail since the recent merges.

e.g. 
- https://github.com/humhub/humhub/actions/runs/7150405383/job/19473640939?pr=6711
- https://github.com/humhub/humhub/actions/runs/7145740499/job/19462051162?pr=6713

All failing tests have to do with

-  `http://localhost/` vs
-  `http://localhost:8080/`

They are unrelated to the above PRs.

luke-:
Probably related to this commit: 
https://github.com/humhub/humhub/commit/ebe45bacfcaecbbb5f703652f3c8d83bce74a0a9c

martin-rueegg:
It appears that the `AutoTourCest` is [failing](https://github.com/humhub/humhub/actions/runs/7181825119/job/19557011099#step:18:1531) on various PRs on `develop`.

@yurabakhtin, since you re-enabled it in

- #6685

you might find quickly what's the reason?

yurabakhtin:
> @yurabakhtin, since you re-enabled it in
> 
> * [Refactor auto start tour for new user #6685](https://github.com/humhub/humhub/pull/6685)
> 
> you might find quickly what's the reason?

@martin-rueegg Yes, I still could not find how to fix the test, because locally it works without errors, but on github side it has a random effect, e.g. you can find the latest commit in the PR https://github.com/humhub/humhub/pull/6685/commits/dd3c56ad5a455bfd139fa8f561ec64512170a791 and all tests are green there, but after merging they all are red.

@luke- I think we should disable the test again - https://github.com/humhub/humhub/pull/6731.

martin-rueegg:
> > @yurabakhtin, since you re-enabled it in
> > 
> > * [Refactor auto start tour for new user #6685](https://github.com/humhub/humhub/pull/6685)
> > 
> > you might find quickly what's the reason?
> 
> @martin-rueegg Yes, I still could not find how to fix the test, because locally it works without errors, but on github side it has a random effect, e.g. you can find the latest commit in the PR [dd3c56a](https://github.com/humhub/humhub/commit/dd3c56ad5a455bfd139fa8f561ec64512170a791) and all tests are green there, but after merging they all are red.

Oh jeah, I know that situation. Somehow frustrating, isn't it? :-)

> @luke- I think we should disable the test again - #6731.

It would also be great to find a solution for the port in the host address that makes several other tests fail. Again, before the merge all was green.

[...]

PR #6719, opened 2023-12-10
[WIP] Enh: Set running migration to `Yii::$app->currentMigration` (ignore for now)

martin-rueegg:
Newly introduced automations (e.g. `afterSave()`) may depend on fields that are only about to be installed by the migrations. However, if those automations are called during initialization of the app, they will fail. The current PR sets a property `Yii::$app->currentMigration` to the currently executed migration. Hence, an automation can (not) execute some code if currently migrations (or a specific migration) is being executed.

- Feature

- No

- [x] All tests are passing
- [x] Changelog was modified

**Other information:**

### DIFF
--- a/protected/humhub/components/ApplicationTrait.php
+++ b/protected/humhub/components/ApplicationTrait.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * @link      https://www.humhub.org/
  * @copyright Copyright (c) 2023 HumHub GmbH & Co. KG
@@ -8,6 +9,7 @@
 namespace humhub\components;
 
 use humhub\interfaces\MailerInterface;
+use yii\db\MigrationInterface;
 use yii\helpers\Url;
 
 trait ApplicationTrait
@@ -26,6 +28,12 @@ trait ApplicationTrait
      * @var string Minimum PHP version that may works but probably with small issues
      */
     public $minSupportedPhpVersion;
+
+    /**
+     * @var MigrationInterface|null Will be set to the instance of the currently running migration
+     * @see Migration::up() and Migration::down()
+     */
+    public ?MigrationInterface $currentMigration = null;
 
     /**
      * @inheritdoc
--- a/protected/humhub/components/Migration.php
+++ b/protected/humhub/components/Migration.php
@@ -60,7 +60,11 @@ class Migration extends \yii\db\Migration
      */
     public function up()
     {
-        return $this->saveUpDown([$this, 'safeUp']);
+        Yii::$app->currentMigration = $this;
+        $result = $this->saveUpDown([$this, 'safeUp']);
+        Yii::$app->currentMigration = null;
+
+        return $result;
     }
 
     /**
@@ -69,7 +73,11 @@ class Migration extends \yii\db\Migration
      */
     public function down()
     {
-        return $this->saveUpDown([$this, 'safeDown']);
+        Yii::$app->currentMigration = $this;
+        $result = $this->saveUpDown([$this, 'safeDown']);
+        Yii::$app->currentMigration = null;
+
+        return $result;
     }
 
     /**
@@ -482,9 +490,19 @@ class Migration extends \yii\db\Migration
      */
     public function integerReferenceKey(): ColumnSchemaBuilder
     {
-        return $this->integer(11)
-            ->notNull()
-            ;
+        return $this->integer(11)->notNull();
+    }
+
+    /**
+     * Returns the field configuration for a FK field
+     *
+     * @return ColumnSchemaBuilder
+     * @since 1.16
+     * @noinspection PhpUnused
+     */
+    public function integerReferenceKeyUnsigned(): ColumnSchemaBuilder
+    {
+        return $this->integerReferenceKey()->unsigned();
     }
 
     /**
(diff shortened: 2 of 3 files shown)


martin-rueegg:
All failing errors are related to
- #6714

protected/humhub/components/ApplicationTrait.php
```diff
@@ -27,6 +29,12 @@ trait ApplicationTrait
      */
     public $minSupportedPhpVersion;
 
+    /**
+     * @var MigrationInterface|null Will be set to the instance of the currently running migration
+     * @see Migration::up() and Migration::down()
+     */
+    public ?MigrationInterface $currentMigration = null;
```
luke-, line 36: What do you think about using the new `MigrationService` for this?

```
public static ?MigrateInterface $currentMigration = null;
```

protected/humhub/components/Migration.php
```diff
@@ -60,7 +60,11 @@ public function init()
      */
     public function up()
     {
-        return $this->saveUpDown([$this, 'safeUp']);
+        Yii::$app->currentMigration = $this;
```
luke-, line 63:
```suggestion
       humhub\services\MigrationService::$currentMigration = $this;
```

luke- (changes requested):

protected/humhub/components/ApplicationTrait.php
martin-rueegg: That's an interesting idea.

- pro:
  - In terms of logic (as in what belongs where) I love the idea
- contra
  - in terms of efficiency: the `MigrationService` class would have to be loaded not only during migration, but also during regular request cycles, as the _consuming_ automation would load the class just to find the field empty. The application object is already loaded ... (for the same reason I've not stored it as a "component", but directly as a fixed field).

Happy to adjust to your verdict ... :-)

martin-rueegg: > * the _consuming_ automation would load the class just to find the field empty

Well, we could do a 
```php
if (class_exists(humhub\services\MigrationService::$class, false) && MigrationService::$currentMigration !== null) {
   // do what needs to be done during migration
} else {
   // to the regular stuff
}
```

luke-: @martin-rueegg Can you give an example of when we need to use this feature?

Do I understand correctly that some events have to check whether migrations are currently being executed?

martin-rueegg: If had similar situations before.

Currently I use it here, where I would update a polymorphic relation to synchronize `guid` and `gid`. But this can only happen _after_ the migration is applied. Yet, I use those object so migrate the data. An alternative approach would be to create a mock object during migration ...

https://github.com/humhub/humhub/pull/6718/files#diff-86ef7d668695f939c218131eb996b6c3218e640f46349304efc41b3d51babe2fR54-R81

luke-: Sorry, I didn't quite get it.
Normally the migrations are executed separately and the models do not have to check this?

martin-rueegg: The thing is, that I wanted to add the `gid` field to `contentcontainer`, `space`, and `user` tables. To fill them with initial values, I have to generate the GlobalId record of which I then read the `gid` and save it to the respective table.

But when if I do one table at the time (since their migrations are in different modules) some fields do not yet exist, that are updated from the `GlobalId::afterSave()` method.

But I can have a look to see if there's another way to solve this.

luke-: I hadn't time yet, to take a closer look at the Global ID PR and concept. 

But in principle, a migration for core models would be like this.

// Example Migration
``` 
function migrateUp() {
    foreach (GetRowsInUserTable() as $row) {
               $id = InsertRowInGidTable();
               UpdateUserRowWithGid();
     }
}
```

 It would be good if DB migrations work as low-level as possible without Models (afterSave etc.), as otherwise Model changes may break future migrations.

martin-rueegg: Yes, that makes sense. Thank you.

so the `GetRowsInUserTable();` and `InsertRowInGidTable();` would be something to include in the migration class, right?

luke-: I would usually use code like this in the migration: 

```php
    function up()
    {
        foreach ((new \yii\db\Query())->select("*")->from('user')->all() as $row) {
            $this->insert('gid', [/* DATA FOR GID Table */]);
            $this->update('user', ['gid' => $db->lastInsertId()], ['id' => $row['id']]);
        }
    }
```

In case for global id and also the need to migrate lot of module data in future, I would add a method like this:

// humhub\components\Migration
```php
   function createGidRecord($class, $id): int {
            $this->insert('gid', [/* DATA FOR GID Table */]);
           return $db->lastInsertId();
   }
``` 
 
// in migration e.g. for Users 
```php
    function up()
    {
        foreach ((new \yii\db\Query())->select("*")->from('user')->all() as $row) {
            $this->update('user', ['gid' => $this->createGidRecord(User::class, $row['id'])], ['id' => $row['id']]);
        }
    }
```

martin-rueegg: Thank you for your extended example. Much appreciated!

PR #6720, opened 2023-12-10
Enh: Consolidate `(is|set)(Database)Installed()` to `ApplicationTrait`

martin-rueegg:
- Code de-duplication
- move code away from `Settings`, that has nothing to do with `Settings`
- move it to `BaseSettingsManager` where also `isDatabaseInstalled()` resides

- Refactor

- No

- [x] All tests are passing
- [x] Changelog was modified

**Other information:**

### DIFF
--- a/MIGRATE-DEV.md
+++ b/MIGRATE-DEV.md
@@ -11,6 +11,8 @@ Version 1.16 (Unreleased)
 - `\humhub\modules\content\components\ContentAddonActiveRecord::canWrite()`
 - `\humhub\modules\file\models\File::canRead()` use `canView()` instead
 - `\humhub\modules\content\components\ContentAddonActiveRecord::canRead()` use `canView()` instead
+- `\humhub\models\Setting::isInstalled()` use `Yii::$app->isInstalled()` instead
+- `\humhub\libs\BaseSettingsManager::isDatabaseInstalled()` use `Yii::$app->isDatabaseInstalled()` instead
 
 ### Type restrictions
 - `\humhub\modules\comment\models\Comment` on `canDelete()`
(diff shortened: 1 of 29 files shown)


martin-rueegg:
All failing errors seem to stem from
- #6714

MIGRATE-DEV.md
```diff
@@ -11,6 +11,7 @@ Version 1.16 (Unreleased)
 - `\humhub\modules\content\components\ContentAddonActiveRecord::canWrite()`
 - `\humhub\modules\file\models\File::canRead()` use `canView()` instead
 - `\humhub\modules\content\components\ContentAddonActiveRecord::canRead()` use `canView()` instead
+- `\humhub\models\Setting::isInstalled()` use `BaseSettingsManager::isInstalled()` instead
```
luke-, line 14: I like the improvement. Just as a thought, could this also be something for `ApplicationTrait`?  

e.g. `Yii::$app->isInstalled()`. 


```php
    public static function isInstalled(): bool
    {
        return isset($this->params['installed']) && $this->params['installed'];
    }

```

luke- (changes requested):

MIGRATE-DEV.md
martin-rueegg: Yes, I guess that would be an even better solution! (Although the method would then have to be non-static)

If we do this, I'd like to also move the `BaseSettingsManager::isDatabaseInstalled()` there. And maybe even the `DynamicSetting:set(Database)Installed()`, too.

Would you agree?

luke-: Hmmm, good point.

Yes, then let's move all 4 methods `is|set[Database]Installed()` to `ApplicationTrait`. Seems to be the best place.

martin-rueegg: on it!

luke- (approved):

luke-:
@martin-rueegg It's much better that way! Thanks for the refactoring!

PR #6729, opened 2023-12-12
 Fix #6720 Regression re isDatabaseInstalled

martin-rueegg:
Since `protected/humhub/config/common.php` defined `Yii::$app->params['databaseInstalled']` to be false, `Yii::$app->isDatabaseInstalled()` would return false, unless the value is overwritten by `@config/dynamic.php`.

During tests, `@config/dynamic.php` does not set the paramter to true.

The current solution is to remove the paramter alltogether from the default set of config. As such, the value remains undefined unless set in `@config/dynamic.php`.

`Yii::$app->isDatabaseInstalled()` will now check the real database connection not only if the `$checkConnection` parameter is set to true, but also if the parameter has not yet been defined.

- Bugfix

- No

- [ ] All tests are passing
- [ ] Changelog was modified

### DIFF
--- a/protected/humhub/components/ApplicationTrait.php
+++ b/protected/humhub/components/ApplicationTrait.php
@@ -102,9 +102,9 @@ trait ApplicationTrait
      */
     public function isDatabaseInstalled(bool $checkConnection = false): bool
     {
-        $dieOnError = isset(Yii::$app->params['databaseInstalled']) && $this->params['databaseInstalled'];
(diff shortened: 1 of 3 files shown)


martin-rueegg:
Ok, appart from the `AutoTourCest` failure, the other failing tests do actually stem from 
-  #6714

luke- (approved):

PR #6743, opened 2023-12-15
Document return type of `component\User::getIdentity()` as `model\User`

martin-rueegg:
Just a small documentation fix so IDE code completion is more accurate.

### DIFF
--- a/protected/humhub/modules/user/components/User.php
+++ b/protected/humhub/modules/user/components/User.php
@@ -18,13 +18,15 @@ use yii\authclient\ClientInterface;
 
 /**
  * Description of User
+ *
  * @property UserModel|null $identity
+ * @method  UserModel|null getIdentity(bool $autoRenew = true)
  * @mixin Impersonator
  * @author luke
  */
 class User extends \yii\web\User
 {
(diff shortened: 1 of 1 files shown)


luke- (approved):

luke-:
FYI maybe we add something like this in future: https://github.com/humhub/humhub/pull/6640/files#diff-c7dc1c8253329945eab67a27e7891af7ec6a9fbda52d18e93de4d90be3a11e4bR57

martin-rueegg:
> FYI maybe we add something like this in future: https://github.com/humhub/humhub/pull/6640/files#diff-c7dc1c8253329945eab67a27e7891af7ec6a9fbda52d18e93de4d90be3a11e4bR57

I see. Thank you for the heads up.

Is this because the `identityClass` can be overwritten by the configuration? Would it not make sense to require and verify that `identityClass` always is/extends from User?

luke-:
`identityClass` only needs to implement a `IdentityInterface`.
We can override this via PHPDoc to `User`, but I don't like the solution, fixed return types are better :-)

PR #6745, opened 2023-12-15
Enh: harmonise term enabled/disabled vs active/inactive for modules

martin-rueegg:
Renaming thought the code (of core) the terms that relate to Module's enabled state:
- `active` => `enabled`
- `activate` => `enable`
- `deactivate` => `disable`

- Code style update
- Refactor

- Yes

documented in MIGRATE-DEV.md

- [x] All tests are passing
- [x] Changelog was modified

**Other information:**

See
- https://github.com/humhub/humhub/pull/6550#issuecomment-1829335643 (@luke-)
  > > > * I like the idea to remove the term `active` and always work with `enabled`. If you like feel free to create a PR and mark `activate*` as deprecated for now. @yurabakhtin ok from your side?
  > > 
  > > 
  > > Turns out, `\humhub\modules\friendship\Module::getIsEnabled()` is used for a different purpose. Shall we rename this or does that cause too much ripple?
  > 
  > Hmm, I think we can rename it. Any idea? isActivated? :-)
- https://github.com/humhub/humhub/pull/6550#issuecomment-1831571901 (@luke-)
  > > Otherwise, I'd use a more specific term, e.g. `isFriendshipEnabled()` (I'd not use "active" as the functionality may be "active" and still no one uses them, so there is no "active" friendship. "enabled" or "available" would reflect the potentiality.
  > 
  > For now I would prefer `isFriendshipEnabled()`. The `friendship` module is quite special. It's a core module, always activated, but can be disabled via Admin Settings.
- https://github.com/humhub/humhub/pull/6550#issuecomment-1832191248
  > > @luke-
  > > There are also a few places, where the term `enabled` or `isEnabled` is used
  > > 
  > > * `\humhub\modules\file\models\FileHistory::isEnabled()` refers to the functionality, not the module (should be fine)
  > >   👍
  > 
  > > * `\humhub\modules\marketplace\Module::isEnabled()` connected to
  > > * `\humhub\modules\marketplace\Module::$enabled` which is set to false if `Yii::$app->params['humhub']['apiEnabled']` is not `true`, and could also be disable through configuration
  > 
  > TBD
  > 
  > > * `\humhub\modules\content\components\ContentContainerModuleManager::isEnabled($id)` checks if module `$id` is enabled on the container
  > 
  > Is ok or?
  > 
  > > So the there is some potential for confusion between
  > > 
  > > * `\humhub\modules\marketplace\Module::isEnabled()`, which refers to a configured state, and it's inherited
  > > * `\humhub\components\Module::getIsEnabled()`, which refers to the module enabled state (but which is always true for a core module like the marketplace.
  > > 
  > > Would it make sense to also rename `\humhub\modules\marketplace\Module::isEnabled()` into `\humhub\modules\marketplace\Module::getIsEnabled()`, overriding the `\humhub\components\Module::getIsEnabled()`? In a way it seems to have a similar meaning. Or does that increase the confusion?
  > > Alternatively, we could obviously leave it, as it is, or rename it to `\humhub\components\Module::isMarketplaceEnabled()`, analogous to the friendship module.
  > 
  > Hmm, I would prefer to introduce `\humhub\modules\marketplace\Module::isMarketplaceEnabled()` analgour to the new `\humhub\modules\friendship\Module::isFriendshipEnabled()`. What do you think @yurabakhtin ?
  > 
  > For `Module::isFriendshipEnabled()` we need to do some refactorings in the `calendar`, `cfiles`, `content-bookmarks`, `mail` , `task`.

### DIFF
--- a/protected/humhub/modules/friendship/Module.php
+++ b/protected/humhub/modules/friendship/Module.php
@@ -26,7 +26,7 @@ class Module extends \humhub\components\Module
      *
      * @return boolean is enabled
      */
-    public function getIsEnabled()
+    public function isFriendshipEnabled(): bool
     {
         if (Yii::$app->getModule('friendship')->settings->get('enable')) {
             return true;
--- a/protected/humhub/modules/marketplace/Module.php
+++ b/protected/humhub/modules/marketplace/Module.php
@@ -11,8 +11,8 @@ namespace humhub\modules\marketplace;
 use humhub\components\Module as BaseModule;
 use humhub\modules\marketplace\components\HumHubApiClient;
 use humhub\modules\marketplace\components\LicenceManager;
-use humhub\modules\marketplace\models\Licence;
 use humhub\modules\marketplace\components\OnlineModuleManager;
+use humhub\modules\marketplace\models\Licence;
 use Yii;
 
 /**
@@ -62,6 +62,8 @@ class Module extends BaseModule
      * @var array A list of module ids that cannot be installed.
      */
     public $moduleBlacklist = [];
+    private $_onlineModuleManager = null;
+    private $_humhubApi = null;
 
     /**
      * @inheritdoc
@@ -70,10 +72,6 @@ class Module extends BaseModule
     {
         return Yii::t('MarketplaceModule.base', 'Marketplace');
     }
-
-    private $_onlineModuleManager = null;
-
-    private $_humhubApi = null;
 
     /**
      * @inheritDoc
@@ -93,8 +91,18 @@ class Module extends BaseModule
 
     /**
      * @return bool
+     * @deprecated since v1.16; use `static::isMarketplaceEnabled()` instead
+     * @see static::isMarketplaceEnabled()
      */
     public static function isEnabled(): bool
+    {
+        return static::isMarketplaceEnabled();
+    }
+
+    /**
+     * @return bool
+     */
+    public static function isMarketplaceEnabled(): bool
     {
         /* @var Module $marketplaceModule */
         $marketplaceModule = Yii::$app->getModule('marketplace');
--- a/CHANGELOG-DEV.md
+++ b/CHANGELOG-DEV.md
@@ -3,6 +3,7 @@ HumHub Changelog
 
 1.16.0 (Unreleased)
 -------------------
+- Enh #6745: Harmonise term `enabled/disabled` vs `active/inactive` for modules
 - Fix #6754: Regression due to return type (#6550)
 - Enh #6550: Improve module migrations
 - Fix #6237: Migration errors during module activation are ignored
--- a/MIGRATE-DEV.md
+++ b/MIGRATE-DEV.md
@@ -14,6 +14,10 @@ Version 1.16 (Unreleased)
 - `\humhub\modules\content\components\ContentAddonActiveRecord::canRead()` use `canView()` instead
 - `\humhub\modules\content\components\ContentAddonActiveRecord::canWrite()`
 - `\humhub\modules\file\models\File::canRead()` use `canView()` instead
+- `\humhub\modules\friendship\Module::getIsEnabled()` use `isFriendshipEnabled()` instead
+  (note: `\humhub\modules\friendship\Module::getIsEnabled()` and the virtual
+  property `\humhub\modules\friendship\Module::isEnabled` now return the status of the module -
+  which yields always true for core modules.)
 
 ### Type restrictions
 - `\humhub\commands\MigrateController` enforces types on fields, method parameters, & return types
--- a/MIGRATE-DEV.md
+++ b/MIGRATE-DEV.md
@@ -18,6 +18,7 @@ Version 1.16 (Unreleased)
   (note: `\humhub\modules\friendship\Module::getIsEnabled()` and the virtual
   property `\humhub\modules\friendship\Module::isEnabled` now return the status of the module -
   which yields always true for core modules.)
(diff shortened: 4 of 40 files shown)


protected/humhub/modules/marketplace/Module.php
```diff
@@ -71,10 +73,6 @@ public function getName()
         return Yii::t('MarketplaceModule.base', 'Marketplace');
     }
 
-    private $_onlineModuleManager = null;
-
-    private $_humhubApi = null;
-
```
martin-rueegg, line 77: Funny place to locate field variables. :-) They have now been moved to the top ....

protected/humhub/modules/friendship/Module.php
```diff
@@ -26,7 +26,7 @@ class Module extends \humhub\components\Module
      *
      * @return boolean is enabled
      */
-    public function getIsEnabled()
```
yurabakhtin, line 29: @martin-rueegg It seems we should store this method as deprecated as you did for `isMarketplaceEnabled()`.

yurabakhtin (changes requested):

protected/humhub/modules/friendship/Module.php
martin-rueegg: Thank you @yurabakhtin for looking into this.

The issue is:
- `marketplace\Module` extends `humhub\components\Module`, whereas
- `friendship\Module` extends `\humhub\components\Module`

Now `\humhub\components\Module` newly defines `getIsEnabled()` to return the state if the _module_ is enabled. However, the old `friendship\Module::getIsEnabled()` returned whether the _functionality_ is enabled

So we have a real conflict here. Not an ideal situation. But as I've understood @luke- [here](https://github.com/humhub/humhub/pull/6550#issuecomment-1832191248), that was acceptable.

yurabakhtin: @martin-rueegg Ok, got it, thank you for the explanation.

yurabakhtin (approved):

protected/humhub/modules/friendship/Module.php
martin-rueegg: You're welcome.

Thanks for the review!

martin-rueegg:
Have rebased on top of `develop` and fixed the merge conflict.

luke-:
@martin-rueegg Thanks!

We should also prepare PRs for the following modules that check whether the friendship system is enabled.

- [ ] Calendar
- [ ] CFiles
- [ ] ContentBookmarks
- [ ] Tasks
- [ ] Mail

Did I understand correctly, that `\humhub\components\Module::getIsEnabled()`  is always present, thus this change does not cause a fatal error on modules. But unmigrated modules always detect the friendship module as enabled. (Core Module).

luke-:
I couldn't find any other uses of `isActivated` in our modules.

@yurabakhtinDo you have any ideas where there is a need for migration?

yurabakhtin:
> I couldn't find any other uses of `isActivated` in our modules.
> 
> @yurabakhtinDo you have any ideas where there is a need for migration?

@luke- I could not find where `isActivated` is used in modules.

martin-rueegg:
> Did I understand correctly, that `\humhub\components\Module::getIsEnabled()` is always present, thus this change does not cause a fatal error on modules.

Yes, that's correct, as long as the module extends from `\humhub\components\Module::getIsEnabled()`, which is certainly true for the friendship module - if not for all modules.

> But unmigrated modules always detect the friendship module as enabled. (Core Module).

Yes, I guess also this is true as the friendship module is enabled by default and cannot be disabled either.

luke-:
Ok great, then the migration effort only affects the modules which checks for enabled friendship:  https://github.com/humhub/humhub/pull/6745#issuecomment-1860939908

martin-rueegg:
@luke- what is currently holding back this PR?

luke-:
@martin-rueegg Missing PR's for the modules listed here: https://github.com/humhub/humhub/pull/6745#issuecomment-1860939908


Besides that, the PR looks good for me and is ready to merge!

martin-rueegg:
> We should also prepare PRs for the following modules that check whether the friendship system is enabled.
> 
> * [x]  Calendar
>             https://github.com/humhub/calendar/pull/442
> * [x]  CFiles
>             https://github.com/humhub/cfiles/pull/200
> * [ ]  ContentBookmarks
>             No usage found
> * [x]  Tasks
>              https://github.com/humhub/tasks/pull/248
> * [ ]  Mail
>         No usage found

martin-rueegg:
@luke- I've created some PRs to the best of my knowledge. I'm not sure, if older HH versions should be supported, in which case a version check would have to be made before calling the method.

 The tests of cours fail, as this current PR has not yet been merged.

luke-:
@martin-rueegg Thanks for the PRs. We'll merge them and release a module version shortly before the v1.16 release.

luke- (approved):

martin-rueegg:
> @martin-rueegg Thanks for the PRs. We'll merge them and release a module version shortly before the v1.16 release.

Thanks for your info. I guess it would make most sense that the two terms are identical. Hence, if you decide to keep "enable/disable" in the UI, maybe the same should be the case for the code.

Either way, the adjective "successful" in line https://github.com/humhub/humhub/pull/6745/files#diff-ac4e86a568e376b4163bc52621367fad02dfdf2341771df1ed39132d7ecea53cR24 should still be an adverb (successfully). :-)

PR #6767, opened 2023-12-19
Fix #6766: Since Humhub 1.15.1, getting content polymorphic relation …

### LINKED ISSUE #6766: Since Humhub 1.15.1, getting content polymorphic relation changes the content object_model to the wrong class

1. Create `class One extends ContentActiveRecord`
2. Create `class Two extends One`
3. `$two = new Tow(); $two->save(); $id = $thow->id;`
4. `echo $two->content->object_model;`
5. `$two = Two::findOne($id); echo $two->content->object_model; $two->save();`

4. One::class
5. One::class
6. In the DB, content table, object_model = One::class

### What do you get instead?

4. One::class
5. Two::class
6. In the DB, content table, object_model = Two::class

### Additional info

This is because on commit https://github.com/humhub/humhub/commit/fa78598f4d3eb252d63db15b4dfbc388830bc991 which has now been released in Humhub 1.15, `self::getObjectModel()` has been replaced with `get_class()`.

Problem: all content modified by classes such as Two are progressively loosing the content relation.
So content get lost progressively...
It requires module updates with migration to put back the correct object_model.
But in some cases, the content got duplicated in the DB...
So it's now complex to fix all DB of people that installed 1.15.1...

| Q                | A
| ---------------- | ---
| HumHub version   | 1.15.1

marc-farre:
Fix #6766: Since Humhub 1.15.1, getting content polymorphic relation changes the content object_model to the wrong class

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**

Issue https://github.com/humhub/humhub/issues/6766

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ HumHub Changelog
 - Enh #6734: Trim Base URL on save
 - Fix #6708: Fix displaying of alert when RichText is changed on refresh a page
 - Fix #6747: Fix loading of default container permissions from cache
+- Fix #6766: Since Humhub 1.15.1, getting content polymorphic relation changes the content object_model to the wrong class
 
 1.15.1 (December 13, 2023)
 -------------------------
--- a/protected/humhub/components/behaviors/PolymorphicRelation.php
+++ b/protected/humhub/components/behaviors/PolymorphicRelation.php
@@ -101,8 +101,8 @@ class PolymorphicRelation extends Behavior
             $this->cached = $object;
 
             if ($object instanceof ActiveRecord) {
-                $class = get_class($object);
-                if ($cached === null || get_class($cached) !== $class) {
+                $class = self::getObjectModel($object);
+                if ($cached === null || self::getObjectModel($cached) !== $class) {
                     $this->owner->setAttribute($this->classAttribute, $class);
                 }
 
--- a/protected/humhub/modules/comment/widgets/Comments.php
+++ b/protected/humhub/modules/comment/widgets/Comments.php
@@ -2,10 +2,11 @@
 
 namespace humhub\modules\comment\widgets;
 
+use humhub\components\behaviors\PolymorphicRelation;
+use humhub\components\Widget;
 use humhub\modules\comment\models\Comment as CommentModel;
 use humhub\modules\comment\Module;
 use humhub\modules\content\components\ContentActiveRecord;
-use humhub\components\Widget;
 use humhub\modules\content\widgets\stream\StreamEntryOptions;
 use humhub\modules\content\widgets\stream\WallStreamEntryOptions;
 use Yii;
@@ -62,7 +63,7 @@ class Comments extends Widget
      */
     public function run()
     {
-        $objectModel = get_class($this->object);
+        $objectModel = PolymorphicRelation::getObjectModel($this->object);
         $objectId = $this->object->getPrimaryKey();
 
         $streamQuery = Yii::$app->request->getQueryParam('StreamQuery');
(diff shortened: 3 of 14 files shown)


luke- (approved):

martin-rueegg:
I personally would not remove the `PolymorphicRelation::getObjectModel` for content containers. Quite the opposite. a) for reasons of consistency, and b) because it might be required at some point.

It's true that currently in the core only `User` and `Space` is used. However, in my module I do extend from `Space`. For me that's fine, as I _want_ to have my own class used in `contentcontainer.class`. But in other circumstances you'd want the same mechanism than with `PolymorphicRelation::getObjectModel` and then it needed to be changed only in one place.

marc-farre:
> I personally would not remove the `PolymorphicRelation::getObjectModel` for content containers.

@luke- what is your thought?

Currently, `getObjectModel()` only works for `ContentActiveRecord` and `ContentAddonActiveRecord`. See
https://github.com/humhub/humhub/blob/master/protected/humhub/components/behaviors/PolymorphicRelation.php#L119

I think that if we introduce `getObjectModel()` for containers, it should be done on the develop branch.

martin-rueegg:
> I think that if we introduce `getObjectModel()` for containers, it should be done on the develop branch.

Sure, that should be done on `develop`. I was just intending to be on the safe side ...

I hope I did not create any confusion.

luke-:
> > I personally would not remove the `PolymorphicRelation::getObjectModel` for content containers.
> 
> @luke- what is your thought?
> 
> Currently, `getObjectModel()` only works for `ContentActiveRecord` and `ContentAddonActiveRecord`. See https://github.com/humhub/humhub/blob/master/protected/humhub/components/behaviors/PolymorphicRelation.php#L119
> 
> I think that if we introduce `getObjectModel()` for containers, it should be done on the develop branch.

The current implementation should also work for `ContentContainers` via the `get_class()` fallback?

What is currently not possible: To allow inherited `ContentContainer` classes, as in the case of `Content`. But I don't think we need to support that either. Or have I overlooked something here?

luke- (approved):
@marc-farre Looks good for me.

@yurabakhtin Can you please also take a look here?

If ok, we'll release the 1.15.2

martin-rueegg:
All I wanted to avoid is another situation as were're just trying to resolve with this PR: Although it looks perfectly fine substituting `getObjectModel` with `get_class` I would have reccommended not doing this in a hotfix.

On the long run, and for `develop`, I would indeed suggest to support to allow inherited ContentContainer ...

marc-farre:
> The current implementation should also work for `ContentContainers` via the `get_class()` fallback?

Yes, but it's useless as https://github.com/humhub/humhub/blob/master/protected/humhub/components/behaviors/PolymorphicRelation.php#L119 only triggers `ContentActiveRecord` and `ContentAddonActiveRecord`.

> I don't think we need to support that either

Currently, I see no use case for it.

So I think the code is fine for now. We do not implement `ContentContainers::getObjectModel()`.

marc-farre:
> substituting `getObjectModel` with `get_class`

@martin-rueegg I haven't done this on https://github.com/humhub/humhub/pull/6767/files
Only substituting `get_class` with `getObjectModel`

martin-rueegg:
> > substituting `getObjectModel` with `get_class`
> 
> @martin-rueegg I haven't done this on https://github.com/humhub/humhub/pull/6767/files Only substituting `get_class` with `getObjectModel`

I know. You've reverted the change. My comment was from earlier ...

yurabakhtin:
> @yurabakhtin Can you please also take a look here?
> 
> If ok, we'll release the 1.15.2

@luke- I think all good here, it can be merged.

PR #6768, opened 2023-12-20
Fix MultiSelect when empty selection is given

martin-rueegg:
Currently, the `BasePicker` and the `MultiSelect` widgets try to get the value from the module, if `$config['selection']` is `empty` (which includes an empty array!).

To avoid such a lookup by the widget, one would expect to provide an empty array (which may actually constitute the current selection). However, this will be ignored due to the above.

This PR does treat an empty array as a valid selection list. Thus, avoiding looking up the value.

Additionally, if the field value is an `ArrayAccess` instance, rather than an `array`, the current code fails with a fatal error. Using `ArrayHelper::keyExists()` rather than `in_array()` solves this issue.

- Bugfix

- Maybe Yes
- Probably No

Not sure if there are use cases, where and empty array has been provided as `selection` and was expected to be replaced by the attribute value of the model or some other source. To me this would not make sense. Because if you expect the `selection` to be filled by the field/widget, then why would you provide an empty array, rather then just leaving the config alone?

That's why I currently filed the PR against `master`.

- [ ] All tests [tests](#issuecomment-new) are passing
- [x] Changelog was modified

**Other information:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,11 @@
 HumHub Changelog
 ================
 
-1.15.2 (December 19, 2023)
+1.15.3 (unreleased)
+--------------------------
+- Fix #6768 MultiSelect when empty selection is given
+
+- 1.15.2 (December 19, 2023)
 --------------------------
 - Fix #6753: Non-unique key used for permission caching
 - Fix #6741: Fix no pretty url of password recovery link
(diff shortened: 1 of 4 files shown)


luke-:
@martin-rueegg Thanks for the fix.

I would prefer to merge such fixes into the `develop` branch. As long as it has no immediate impact on productive installations.

Please let me know if you absolutely need it (or if the function is affected in any way), then we can merge it into `master` branch.

marc-farre:
@luke- I've added a comment here about an issue introduced by this PR: https://github.com/humhub/humhub/commit/dd2d23058dbffa53fd1718123f9f61c9b79588fe#r141253611

PR #6773, opened 2023-12-21
Fix: polymorphic relation lookup (Fix #6772)

### LINKED ISSUE #6772: Humhub 1.16.0 - Error with `PolymorphicRelation::getPolymorphicRelation()`

I cannot tell, but on my Humhub instance, it occurred when viewing the notification list.

The problem is that `$this->owner->getAttribute()` can return null:
https://github.com/humhub/humhub/blob/1db793da1f0d680053713b0654bdd5f32a5339cd/protected/humhub/components/behaviors/PolymorphicRelation.php#L70-L70

But the first argument of `static::loadActiveRecord()` must be a string:
https://github.com/humhub/humhub/blob/1db793da1f0d680053713b0654bdd5f32a5339cd/protected/humhub/components/behaviors/PolymorphicRelation.php#L177C50-L177C50

### Additional info

Log:
```
TypeError: humhub\components\behaviors\PolymorphicRelation::loadActiveRecord(): Argument #1 ($className) must be of type string, null given, called in /var/www/humhub/protected/humhub/components/behaviors/PolymorphicRelation.php on line 69 and defined in /var/www/humhub/protected/humhub/components/behaviors/PolymorphicRelation.php:177
Stack trace:
#0 /var/www/humhub/protected/humhub/components/behaviors/PolymorphicRelation.php(69): humhub\components\behaviors\PolymorphicRelation::loadActiveRecord(NULL, NULL)
#1 [internal function]: humhub\components\behaviors\PolymorphicRelation->getPolymorphicRelation()
#2 /var/www/humhub/protected/vendor/yiisoft/yii2/base/Component.php(297): call_user_func_array(Array, Array)
#3 /var/www/humhub/protected/humhub/modules/notification/models/Notification.php(130): yii\base\Component->__call('getPolymorphicR...', Array)
#4 /var/www/humhub/protected/humhub/modules/notification/controllers/ListController.php(46): humhub\modules\notification\models\Notification->getBaseModel()
#5 [internal function]: humhub\modules\notification\controllers\ListController->actionIndex()
#6 /var/www/humhub/protected/vendor/yiisoft/yii2/base/InlineAction.php(57): call_user_func_array(Array, Array)
#7 /var/www/humhub/protected/vendor/yiisoft/yii2/base/Controller.php(178): yii\base\InlineAction->runWithParams(Array)
#8 /var/www/humhub/protected/vendor/yiisoft/yii2/base/Module.php(552): yii\base\Controller->runAction('', Array)
#9 /var/www/humhub/protected/vendor/yiisoft/yii2/web/Application.php(103): yii\base\Module->runAction('notification/li...', Array)
#10 /var/www/humhub/protected/vendor/yiisoft/yii2/base/Application.php(384): yii\web\Application->handleRequest(Object(humhub\components\Request))
#11 /var/www/humhub/index.php(28): yii\base\Application->run()
#12 {main}
```

| Q                | A
| ---------------- | ---
| HumHub version   | 1.16.0
| PHP version      | 8.2

martin-rueegg:
- Fix #6772
- Regression #6587

- Bugfix

- No

- [ ] All tests [tests](#issuecomment-new) are passing
- [x] Changelog was modified

**Other information:**

### DIFF
--- a/protected/humhub/components/behaviors/PolymorphicRelation.php
+++ b/protected/humhub/components/behaviors/PolymorphicRelation.php
@@ -66,16 +66,16 @@ class PolymorphicRelation extends Behavior
             return $this->cached;
         }
 
-        $record = static::loadActiveRecord(
-            $this->owner->getAttribute($this->classAttribute),
-            $this->owner->getAttribute($this->pkAttribute)
-        );
+        $className = $this->owner->getAttribute($this->classAttribute);
+        $primaryKey = $this->owner->getAttribute($this->pkAttribute);
+
+        $record = static::loadActiveRecord($className, $primaryKey);
 
         if ($this->strict && !$record && !empty($this->classAttribute) && !empty($this->pkAttribute)) {
             throw new IntegrityException(
                 'Call to an inconsistent polymorphic relation detected on '
                 . ($this->owner === null ? 'NULL' : get_class($this->owner))
-                . ' (' . $this->owner->getAttribute($this->classAttribute) . ':' . $this->owner->getAttribute($this->pkAttribute) . ')'
+                . ' (' . $className . ':' . $primaryKey . ')'
             );
         }
 
@@ -169,12 +169,12 @@ class PolymorphicRelation extends Behavior
     /**
      * Loads an active record based on classname and primary key.
      *
-     * @param string $className
+     * @param string|null $className
      * @param string|int $primaryKey
      *
      * @return null|ActiveRecord|ActiveRecordInterface
      */
-    public static function loadActiveRecord(string $className, $primaryKey): ?ActiveRecordInterface
+    public static function loadActiveRecord(?string $className, $primaryKey): ?ActiveRecordInterface
     {
         if (empty($className) || empty($primaryKey)) {
             return null;
--- a/CHANGELOG-DEV.md
+++ b/CHANGELOG-DEV.md
@@ -3,6 +3,7 @@ HumHub Changelog
 
 1.16.0 (Unreleased)
 -------------------
+- Fix #6772: Polymorphic relation lookup (Regression #6587)
 - Enh #6745: Harmonise term `enabled/disabled` vs `active/inactive` for modules
 - Fix #6754: Regression due to return type (#6550)
 - Enh #6550: Improve module migrations


martin-rueegg:
Regression from https://github.com/humhub/humhub/commit/428c7a3ac58a9942fa816769134e3b685cfb29bc#diff-63b6b65bbd93ed9d3bb6f24c5ede589d08cf030f8887b055599815653bb4697bR153
- #6587

protected/humhub/components/behaviors/PolymorphicRelation.php
```diff
@@ -169,14 +169,14 @@ private function validateUnderlyingObjectType(?object $object)
...
      * Loads an active record based on classname and primary key.
      *
-     * @param string $className
+     * @param string|mixed $className
      * @param string|int $primaryKey
      *
      * @return null|ActiveRecord|ActiveRecordInterface
      */
-    public static function loadActiveRecord(string $className, $primaryKey): ?ActiveRecordInterface
+    public static function loadActiveRecord($className, $primaryKey): ?ActiveRecordInterface
```
luke-, line 177: Would this be also an option? This would allow us to remove the `is_string` check.

```suggestion
    public static function loadActiveRecord(?string $className, $primaryKey): ?ActiveRecordInterface
```

protected/humhub/components/behaviors/PolymorphicRelation.php
```diff
@@ -169,14 +169,14 @@ private function validateUnderlyingObjectType(?object $object)
...
     {
-        if (empty($className) || empty($primaryKey)) {
+        if (($primaryKey !== 0 && empty($primaryKey)) || !is_string($className) || trim($className) === '') {
```
luke-, line 179: I would actually prefer the previous variant here for reasons of readability.

Not sure we need to consider a PK with ID 0 or the string of spaces.  Or do you see a specific reason for this?

luke- (changes requested):
@martin-rueegg Thanks for the quick fix.

protected/humhub/components/behaviors/PolymorphicRelation.php
martin-rueegg: We can try. There is no restriction on the `getAttribute()` method. In fact, in some cases, objects are returned. This would lead to a fatal error again. So I wanted to be on the safe side. But happy to apply `?string` ...

protected/humhub/components/behaviors/PolymorphicRelation.php
martin-rueegg: `0` is a valid primary key. I know that in most places you use `empty()`-checks, but I think that should be avoided.

The `trim()` is not required, of course. Happy to remove.

luke-: You are right, it is a valid PK, but normally there are no situations in which this PK is assigned. e.g. `auto_increment` starts with 1.

I want to keep the codebase as consistent and simple as possible - so I don't want to implement an extra check for `0` here. If it becomes necessary to allow 0, we need to do changes in more places.

protected/humhub/components/behaviors/PolymorphicRelation.php
luke-: Thank you, being on the safe side is usually a good thing :-)

But let's risk it here.

luke- (approved):

Issue #6780, opened 2024-01-07
Create new user by SSO without allowing email self-registrations.


jonthorne:
### Is your feature request related to a problem? Please describe.

Using SSO login with a valid user in Keycloak, but not yet in humhub and "New users can register" is disabled, this blocks the new user registration. I only want Keycloak users to be able to login and don't want others to be able to self register by email.

### Describe the solution you'd like

I would like a way to disable user registration on humhub via the "register form" while still allowing the SSO flow to provision users if they do not exists in humhub. I want to control the registration of new users via another system, not on humhub. And allow any valid Keycloak user to be able to login to hubhub.

### Describe alternatives you've considered

The only option I can find for now is to enable "New users can register". But try and find a way to block registration by email account.

ArchBlood:
In your theme view overwrite you can remove the following lines as alternative;

https://github.com/humhub/humhub/blob/56d42191afecf4c06c45614f9d154dc67f6c1889/protected/humhub/modules/user/views/auth/login.php#L99-L113

https://github.com/humhub/humhub/blob/56d42191afecf4c06c45614f9d154dc67f6c1889/protected/humhub/modules/user/views/auth/login_modal.php#L110-L125

jonthorne:
Hi @ArchBlood

Thanks for this prompt reply. This is a very helpful way to get around the issue for now. Although I will have to make sure to keep modifying this after any software update. A separate setting for this scenario would still be helpful if at all possible.

Jon.

marc-farre:
I use CSS to deal with this problem:
1. You need Humhub 1.16 or to patch your 1.15 with this: https://github.com/humhub/humhub/pull/6616/files
2. Add this CSS code to your custom theme:

```css
#user-auth-login-modal #tabs,
#create-account-form[data-has-auth-client="0"] #registration-form,
form[action="/user/auth/login"],
#register-form,
#account-login-form,
.or-container {
  display: none !important;
}
```

See comment on https://community.humhub.com/comment/perma?id=46424 (You must be a member of the https://community.humhub.com/s/installation-and-setup space).

You can also skip the login form this way (won't work for account creation):
https://community.humhub.com/s/installation-and-setup/wiki/Configuration+Examples#skip-login-form

Maybe we could leave this issue open to provide later a better solution.

luke-:
We could either add a new checkbox "New user can register via SSO" or create exceptions for SSO providers via the configuration file.

marc-farre:
[...]

PR #6788, opened 2024-01-09
Test Disable Login Form

luke-:
Related: https://github.com/humhub/humhub/issues/6780#issuecomment-1881296805

configuration file example
```php
return [
    'container' => [
        'definitions' => [
            \humhub\modules\user\controllers\AuthController::class => [
                'showLoginForm' => false,
                'allowUserRegistrationFromAuthClientIds' => ['keycloak', 'saml']
            ]
        ]
    ]
];
```

### DIFF
--- a/CHANGELOG-DEV.md
+++ b/CHANGELOG-DEV.md
@@ -44,3 +44,5 @@ HumHub Changelog
 - Fix #6777: Allow access to pending registrations to non-system admins
 - Enh #6783: Lowercase user email before save
 - Enh #6786: Improve password hashing
+- Enh #6788: Allow to disable login/registration form via configuration file
+- Enh #6788: Allow new user registration of specified SSO providers via configuration file 
--- a/protected/humhub/modules/user/Module.php
+++ b/protected/humhub/modules/user/Module.php
@@ -161,6 +161,33 @@ class Module extends \humhub\components\Module
     public $includeEmailInSearch = true;
 
     /**
+     * Should the login form be displayed. This can be deactivated, e.g. to display only SSO providers.
+     * With the parameter `?showLoginForm=1` the login form can still be displayed as a fallback.
+     *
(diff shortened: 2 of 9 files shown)


marc-farre:
Thanks very much @luke- !
It works fine and will be very helpful.

What about also adding the `showRegistrationForm` param?
It would allow removing whats in the pink rectangles (we also should remove the "or" horizontal line):

image

image

image

If it helps, I can contribute with a commit for this in a few days.

luke-:
@marc-farre I have added a `showRegistrationForm` variable. Do we need to show the `AuthClients` here again? Wouldn't it be sufficient to show the AuthClients once on Login?

Would be great when you can find some time to test this. Thanks!

marc-farre:
Thanks, that's awesome!

> Do we need to show the `AuthClients` here again? Wouldn't it be sufficient to show the AuthClients once on Login?

Yes, it's sufficient and perfect doing it this way.

I've removed the OR horizontal bar in commit https://github.com/humhub/humhub/pull/6788/commits/060d112e493a559dd640b3b17714e4e1c7b26c46

image

Now we need to make that work for email and link invites:

`/user/registration/by-link?token=xxxxxxxxx`:
image

`/user/registration?token=xxxxxxxxx`:
image

Problem: the controller is not `AuthController`. So maybe we should move `$showLoginForm` and `$showRegistrationForm` to the User `Module` class?

marc-farre:
@luke- Thanks for you PR.
I've added this commit: https://github.com/humhub/humhub/pull/6788/commits/1b3fabd8788376fc63655299bdbebe032d281071
It allows hiding the form for email invites too.
I've also added the `showOrDivider` attribute to the `AuthChoice` widget allowing to show the OR divider only when we have both the SSO button and the regular login form.
I think this PR is fine now.
Thanks for you work on it!

PR #6809, opened 2024-01-21
Fix: Iso3166Codes class calls

ArchBlood:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] All tests are passing
- [x] Changelog was modified

**Other information:**

### DIFF
--- a/protected/humhub/modules/user/models/fieldtype/CountrySelect.php
+++ b/protected/humhub/modules/user/models/fieldtype/CountrySelect.php
@@ -10,6 +10,7 @@ namespace humhub\modules\user\models\fieldtype;
 
 use humhub\libs\Iso3166Codes;
 use humhub\modules\user\models\User;
+use humhub\libs\Html;
 use Yii;
 
 /**
@@ -65,7 +66,7 @@ class CountrySelect extends Select
                 $key = trim($code);
                 $value = Iso3166Codes::country($key, true);
                 if (!empty($key) && $key !== $value) {
-                    $items[trim($key)] = trim($value);
+                    $items[$key] = trim($value);
                 }
             }
         }
@@ -86,7 +87,7 @@ class CountrySelect extends Select
         $value = $user->profile->$internalName;
 
         if (!$raw) {
-            return \yii\helpers\Html::encode(Iso3166Codes::country($value));
+            return Html::encode(Iso3166Codes::country($value));
         }
 
         return $value;
--- a/protected/humhub/modules/user/models/fieldtype/CountrySelect.php
+++ b/protected/humhub/modules/user/models/fieldtype/CountrySelect.php
@@ -15,7 +15,7 @@ use Yii;
 /**
  * ProfileFieldTypeSelect handles numeric profile fields.
  *
- * @package humhub.modules_core.user.models
+ * @package humhub\modules\user\models\fieldtype
  * @since 0.5
  */
 class CountrySelect extends Select
@@ -29,18 +29,18 @@ class CountrySelect extends Select
     public function getFormDefinition($definition = [])
     {
         return parent::getFormDefinition([
-                    get_class($this) => [
-                        'type' => 'form',
-                        'title' => Yii::t('UserModule.profile', 'Supported ISO3166 country codes'),
-                        'elements' => [
-                            'options' => [
-                                'type' => 'textarea',
-                                'label' => Yii::t('UserModule.profile', 'Possible values'),
-                                'class' => 'form-control',
-                                'hint' => Yii::t('UserModule.profile', 'Comma separated country codes, e.g. DE,EN,AU')
-                            ]
-                        ]
+            get_class($this) => [
+                'type' => 'form',
+                'title' => Yii::t('UserModule.profile', 'Supported ISO3166 country codes'),
+                'elements' => [
+                    'options' => [
+                        'type' => 'textarea',
+                        'label' => Yii::t('UserModule.profile', 'Possible values'),
+                        'class' => 'form-control',
+                        'hint' => Yii::t('UserModule.profile', 'Comma separated country codes, e.g. DE,EN,AU')
                     ]
+                ]
+            ]
         ]);
     }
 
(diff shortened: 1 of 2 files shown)


protected/humhub/modules/user/models/fieldtype/CountrySelect.php
```diff
@@ -55,22 +55,22 @@ public function getSelectItems()
...
+                $items[Iso3166Codes::country($code)] = Iso3166Codes::country($code);
             }
         } else {
             foreach (explode(",", $this->options) as $code) {
-
                 $key = trim($code);
-                $value = iso3166Codes::country($key, true);
+                $value = Iso3166Codes::country($key, true);
                 if (!empty($key) && $key !== $value) {
                     $items[trim($key)] = trim($value);
```
ArchBlood, line 68: I'm not sure why `$key` is being `trim()` here of `$key` equals `$code` which is already `trim()` :thinking:

protected/humhub/modules/user/models/fieldtype/CountrySelect.php
```diff
@@ -86,7 +86,7 @@ public function getUserValue(User $user, $raw = true): ?string
         $value = $user->profile->$internalName;
 
         if (!$raw) {
-            return \yii\helpers\Html::encode(iso3166Codes::country($value));
+            return \yii\helpers\Html::encode(Iso3166Codes::country($value));
```
ArchBlood, line 89: Probably should use HumHub's Html class instead of Yii's base class. :thinking:

marc-farre:
@ArchBlood thanks, I agree the case didn't match the referenced class.

protected/humhub/modules/user/models/fieldtype/CountrySelect.php
ArchBlood: If the variable `$key` is already equal to a trimmed version of another variable, there is no need to trim it again when assigning it to the `$items` array.  If `$key` is obtained through `$key = trim($code)`, and later you check if it's not empty and different from `$value`, you can directly use `$key` without trimming it again when assigning it to `$items`.

If the variable is already equal to a trimmed version, you don't need to trim it again when using it in subsequent operations unless there are specific reasons or requirements to do so?

ArchBlood:
> @ArchBlood thanks, I agree the case didn't match the referenced class.

I also believe that the Iso3166Codes class should be refactored soon as well, but I currently don't have much time on my hands with next month being around the corner, and having most if not all Green Meteor modules updated for PHP 8. 🙃

luke- (approved):

ArchBlood:
@luke- what's your thoughts on this?
```php
public static function country($code, $translate = true)
{
    $locale = '_' . $code;

    // Replace with your desired fallback language if translation is not available
    $fallbackLanguage = 'en-US';

    $region = Locale::getRegion($locale);

    if ($region !== null) {
        // $region now contains the region code
        $regionDisplayName = Locale::getDisplayRegion($locale, $translate ? Yii::$app->language : $fallbackLanguage);

        // $regionDisplayName contains the display name of the region
        return $regionDisplayName;
    } else {
        // Unable to determine region
        return '';
    }
}
```
We should also take note that in PHP 8, the `Locale` class is now a part of the global namespace, so we don't really need the leading backslash `\`. But if we're slowly easing out of PHP 7.4 then we can always include the leading backslash till PHP 8 is set as a full on requirement.

Then in the other functions the following can be done, I'd like @yurabakhtin's overall opinion on this though;
```php
public static function phoneCode($code)
{
    $uppercaseCode = strtoupper($code);
    return isset(static::$phoneCodes[$uppercaseCode]) ? static::$phoneCodes[$uppercaseCode] : 'N/A';
}

public static function isValid($code)
{
    $uppercaseCode = strtoupper($code);
    return isset(static::$countries[$uppercaseCode]);
}
```

luke-:
> @luke- what's your thoughts on this?
> 
> ```
> public static function country($code, $translate = true)
> {
>     $locale = '_' . $code;
> 
>     // Replace with your desired fallback language if translation is not available
>     $fallbackLanguage = 'en-US';
> 
>     $region = Locale::getRegion($locale);
> 
>     if ($region !== null) {
>         // $region now contains the region code
>         $regionDisplayName = Locale::getDisplayRegion($locale, $translate ? Yii::$app->language : $fallbackLanguage);
> 
>         // $regionDisplayName contains the display name of the region
>         return $regionDisplayName;
>     } else {
>         // Unable to determine region
>         return '';
>     }
> }
> ```

Can you please explain what you want to accomplish with the change?

ArchBlood:
> > @luke- what's your thoughts on this?
> > ```
> > public static function country($code, $translate = true)
> > {
> >     $locale = '_' . $code;
> > 
> >     // Replace with your desired fallback language if translation is not available
> >     $fallbackLanguage = 'en-US';
> > 
> >     $region = Locale::getRegion($locale);
> > 
> >     if ($region !== null) {
> >         // $region now contains the region code
> >         $regionDisplayName = Locale::getDisplayRegion($locale, $translate ? Yii::$app->language : $fallbackLanguage);
> > 
> >         // $regionDisplayName contains the display name of the region
> >         return $regionDisplayName;
> >     } else {
> >         // Unable to determine region
> >         return '';
> >     }
> > }
> > ```
> 
> Can you please explain what you want to accomplish with the change?

Take a country code as input and returns the display name of the corresponding region. Here's a breakdown of what the code does:

1. **Input Parameters:**
   - `$code`: The country code for which you want to retrieve the region display name.
   - `$translate`: A boolean parameter indicating whether to translate the display name. If set to `true`, it will attempt to translate the display name using the Yii::$app->language; otherwise, it uses a fallback language ('en-US').

2. **Locale Setup:**
   - The function appends an underscore to the country code to create a locale string (e.g., "_US").
   - It sets up a fallback language ('en-US').

3. **Region Retrieval:**
   - It uses the `Locale::getRegion` function to attempt to retrieve the region code based on the provided locale.
   - If a region is found (`$region !== null`), it uses `Locale::getDisplayRegion` to get the display name of the region.
   - If `$translate` is set to `true`, it attempts to translate the display name using Yii::$app->language; otherwise, it uses the fallback language.

4. **Return Values:**
   - If a region is found, it returns the display name of the region.
   - If no region is found, it returns an empty string.

luke-:
What will your modification change compared to the current function?

ArchBlood:
> What will your modification change compared to the current function?

The main difference would be complexity vs simplicity, the current version is more simplicity, my modification could use a try-catch method which would be better for best practice, although I don't think current HumHub is ready for this kind of change as it would also require more changes to other files as well, so I think we could probably keep it at this for now and revisit this topic at a later date. 🤔

PR #6816, opened 2024-01-24
Prototype: Check module requirements before install/update

luke-:


protected/humhub/modules/marketplace/components/OnlineModuleManager.php
```diff
@@ -122,6 +101,51 @@ private function unzip($file, $folder)
         return true;
     }
 
+    private function checkRequirements($moduleId, $moduleZipFile) {
```
ArchBlood, line 104: I'd like to maybe provide some thoughts through the following modifications

### Original Code:

```php
private function checkRequirements($moduleId, $moduleZipFile) {
    $config = require('zip://' . $moduleZipFile . '#' . $moduleId . '/config.php');
    if (isset($config['requirementCheck']) && is_callable($config['requirementCheck'])) {
        $result = call_user_func($config['requirementCheck']);
        if ($result !== null) {
            throw new HttpException('500', $result);
        }
    }
}
```

### Modified Code:

```php
use yii\base\InvalidArgumentException;
use yii\base\InvalidConfigException;
use yii\web\HttpException;

private function checkRequirements($moduleId, $moduleZipFile) {
    // Check if the ZIP extension is enabled
   // This is so that in cases where the ZIP extension isn't loaded.
   // If not then log the error, another method would be to check against SelfTest for simplicity.
    if (!extension_loaded('zip')) {
        throw new \RuntimeException('PHP ZIP extension is not enabled.');
    }

    try {
        // Check if the module ZIP file exists and is readable
        if (!file_exists($moduleZipFile) || !is_readable($moduleZipFile)) {
            throw new InvalidArgumentException('Unable to access module ZIP file.');
        }

        // Read the 'config.php' file from the specified location within the ZIP archive
        $config = require('zip://' . $moduleZipFile . '#' . $moduleId . '/config.php');

        // Check if the retrieved configuration is an array
        if (!is_array($config)) {
            throw new InvalidConfigException('Invalid configuration format in module.');
        }

        // Check if the configuration specifies a 'requirementCheck' function
        if (isset($config['requirementCheck']) && is_callable($config['requirementCheck'])) {
            // Call the specified 'requirementCheck' function
            $result = call_user_func($config['requirementCheck']);

            // If the 'requirementCheck' function returns a non-null result, throw an exception
            if ($result !== null) {
                // Log the error details using logging
                Yii::error('Requirement check failed: ' . $result, __METHOD__);

                // Throw an HttpException with a status code of '500'
                throw new HttpException('500');
            }
        }
    } catch (\Exception $e) {
        // Handle any exceptions
        Yii::error($e->getMessage(), __METHOD__);
        throw $e;
    }
}
```

### Differences:

1. **Exception Classes:**
   - **Original Code:** Uses a generic `HttpException` for all exceptions.
   - **Modified Code:** Utilizes specific exception classes (`InvalidArgumentTypeException`, `InvalidConfigTypeException`, `HttpException`, `RuntimeException`) for more specific error types.

2. **Error Handling and Logging:**
   - **Original Code:** Directly throws an `HttpException` without detailed logging.
   - **Modified Code:** Implements structured error handling with specific exception classes. Uses logging (`Yii::error`) to log error details.

3. **Try-Catch Block:**
   - **Original Code:** No try-catch block for handling exceptions.
   - **Modified Code:** Introduces a try-catch block to catch exceptions, log them, and then re-throw the exception to allow error handling to take over.

4. **Consistency with Standards:**
   - **Original Code:** Doesn't follow conventions for error handling.
   - **Modified Code:** Adheres to standards for exception handling, making it more consistent.

The modified error handling provides more specific exception classes, structured error handling, and consistent logging, enhancing code quality and maintainability.

luke-:
@yurabakhtin Can you please add some tests for this here? (e.g. try to install a module with missing requirements)
Would be also cool, to have a sample always valid requirements.php file here:
 https://github.com/humhub/example-basic

protected/humhub/components/bootstrap/ModuleAutoLoader.php
```diff
@@ -118,6 +119,12 @@ private static function findModules(iterable $paths): array
         return $modules;
     }
 
+    private static function getModuleConfigByPath(string $modulePath): array
+    {
+        return include $modulePath . DIRECTORY_SEPARATOR . self::CONFIGURATION_FILE;
```
ArchBlood, line 124: After running some tests for this addition I'm getting the following error;

```logs
TypeError: humhub\components\bootstrap\ModuleAutoLoader::getModuleConfigByPath(): Return value must be of type array, int returned in /home/LOCAL/protected/humhub/components/bootstrap/ModuleAutoLoader.php:124
Stack trace:
#0 /home/LOCAL/protected/humhub/components/bootstrap/ModuleAutoLoader.php(87): humhub\components\bootstrap\ModuleAutoLoader::getModuleConfigByPath()
#1 /home/LOCAL/protected/humhub/components/bootstrap/ModuleAutoLoader.php(55): humhub\components\bootstrap\ModuleAutoLoader::findModules()
#2 /home/LOCAL/protected/humhub/components/bootstrap/ModuleAutoLoader.php(39): humhub\components\bootstrap\ModuleAutoLoader::locateModules()
#3 /home/LOCAL/protected/vendor/yiisoft/yii2/base/Application.php(325): humhub\components\bootstrap\ModuleAutoLoader->bootstrap()
#4 /home/LOCAL/protected/vendor/yiisoft/yii2/web/Application.php(69): yii\base\Application->bootstrap()
#5 /home/LOCAL/protected/humhub/components/Application.php(61): yii\web\Application->bootstrap()
#6 /home/LOCAL/protected/vendor/yiisoft/yii2/base/Application.php(271): humhub\components\Application->bootstrap()
#7 /home/LOCAL/protected/humhub/components/Application.php(42): yii\base\Application->init()
#8 /home/LOCAL/protected/vendor/yiisoft/yii2/base/BaseObject.php(109): humhub\components\Application->init()
#9 /home/LOCAL/protected/vendor/yiisoft/yii2/base/Application.php(204): yii\base\BaseObject->__construct()
#10 /home/LOCAL/protected/humhub/components/ApplicationTrait.php(38): yii\base\Application->__construct()
#11 /home/LOCAL/index.php(25): humhub\components\Application->__construct()
#12 {main}
```

To fix this and display better error handling here's a fix that will show exactly which `config.php` that causes any issues so that they can be fixed;
```php
    /**
     * Retrieves the module configuration from the specified path.
     *
     * @param string $modulePath The path to the module directory.
     * @return array The module configuration.
     * @throws \RuntimeException If the configuration file is not found or is invalid.
     */
    private static function getModuleConfigByPath(string $modulePath): array
    {
        $configFilePath = $modulePath . DIRECTORY_SEPARATOR . self::CONFIGURATION_FILE;

        // Check if the configuration file exists
        if (!file_exists($configFilePath)) {
            throw new \RuntimeException('Configuration file not found: ' . $configFilePath);
        }

        // Include the configuration file and ensure it returns an array
        $config = include $configFilePath;

        if (!is_array($config)) {
            throw new \RuntimeException('Invalid configuration file format: ' . $configFilePath);
        }

        return $config;
    }
```

If this method isn't wanted then I'd suggest only check `config.php` of enabled modules.

ArchBlood: To reproduce the error you must have a module that has a blank or missing `config.php` file.

luke-:
@yurabakhtin No, it is not about checking a module with an broken config.

It is about the fact that modules can now have a `requirements.php` that can check system requirements. 

e.g. https://github.com/humhub/rest/blob/master/requirements.php

--

My idea was to create a module with `requirements.php` similar to this one:
https://github.com/humhub/humhub/tree/develop/protected/humhub/tests/codeception/_data/ModuleConfig/module2

In addition, a test that checks that the `requirements.php` is executed and checked.

---

Can you try that out? Otherwise we'll put the issue on hold for now.

yurabakhtin:
@luke- I have tested an enabling the module [example-basic](https://github.com/humhub/example-basic) with created locally the file `requirements.php`, but the `requirements.php` is checked only on install a module from Marketplace.
I don't find the module `example-basic` in Marketplace, i.e. even if we will add the `requirements.php` for the module it will never be used because this module cannot be installede from Marketplace.

---

We could test it with module `rest` where we already have the file `requirements.php` in the branch `master` but the version is not released yet.
I think the file `requirements.php` could be tested if:
- release new version of the module `rest`
- try to install the `rest` module from Marketplace on run tests when pretty URLs are disabled, however it seems tests can be work only with enabled pretty URLs.

---

> My idea was to create a module with `requirements.php` similar to this one:
> https://github.com/humhub/humhub/tree/develop/protected/humhub/tests/codeception/_data/ModuleConfig/module2

ok, thanks for the info about testing modules, but the new method `OnlineModuleManager->checkRequirements()` is called only on run install process which download a module by ID from Marketplace. We can try just test a code `requre 'requirements.php'` of already installed test module but it will not be the testing of the method `OnlineModuleManager->checkRequirements()`, that is required as I understand.
Either we should run the `checkRequirements()` when module is enabling instead of current when module is installing.

luke-:
@marc-farre What do you think about testing the `requirements.php` also when the module is activated, as suggested by Yura?

Then we can integrate this relatively well into the tests.

Otherwise, I would leave the tests out for now, as we don't currently have a Marketplace API mock that can return a ZIP module.

marc-farre:
> What do you think about testing the `requirements.php` also when the module is activated, as suggested by Yura?

I think it's a good idea.

We could even take this opportunity to create a method to check active module requirements which can then be used, in addition to tests, in the "Prerequisites" admin page.

Because `requirements.php` is tested when the module is installed and updated, but the server config can change at any time.

E.g, after changing the PHP version of the server, it could be nice to be able to check if "Prerequisites" are still fine (mainly because sometimes, when upgrading PHP version, some PHP modules are missing on the new version).

yurabakhtin:
@luke- you wrote:

> ... **also** when the module is activated

I think we should completely remove the `requirements.php` testing on install/update module and keep it only on activating, because as I understand currently it works like this:

- download a module package
- unpack the package
- check `requirements.php`
- if error then don't install the module, so user should fix the error and then try again to download the package. For me better is to allow downloading and installing without any restrictions but do this only on activating.

marc-farre:
> keep it only on activating

@yurabakhtin why not preventing update if requirements are not OK?

It's working well with this PR and I think it's important to also check on module update to avoid installing a new version which doesn't work with the current system (e.g, the new module version requires a more recent version of PHP).

luke-:
I also think that the `requirements.php' should be checked before updates.

@yurabakhtin  Can you add it to the tests when we check it during the enable?

yurabakhtin:
@marc-farre @luke- 

> why not preventing update if requirements are not OK?
> 
> I also think that the `requirements.php' should be checked before updates.

Yes, you are right, we should check it before update already installed and activated module.

marc-farre:
> we can remove the checking from the installing process and keep it only for updating and enabling?

From my point of view, I agree with this. It's not a problem to install an incompatible module. Once installed, you can then do what's required to be able to enable the module.

luke-:
@marc-farre @yurabakhtin Is there anything against always check it, i.e. for Install, Update & Enable?

marc-farre:
> @marc-farre @yurabakhtin Is there anything against always check it, i.e. for Install, Update & Enable?

I think checking on Install is not necessary, but it's not a problem too. So it's OK for me to always check it.

luke-:
@yurabakhtin Also ok for you? Then I'll merge this.

yurabakhtin:
> @yurabakhtin Also ok for you? Then I'll merge this.

@luke- yes ok, you can merge it. The checking of requirements doesn't create issue on install, only one case may be not comfortable for user because the package will be downloaded from server many times until all requirements are met.

Issue #3936, opened 2020-03-23
Hashtags


buddh4:
Implement hashtag richtext plugin.

nsturtz2:
I found the code it is at protected/vendor/npm-asset/at.js and there is a hashtags file protected/vendor/npm-asset/at.js/examples/hashtags.html how do I enable it to use in Humhub?

MasterMindNET:
What would you like to tag using a hashtag? there is an `at` (@) sign to reference someone or space.
What would be the main purpose of this plugin (hashtag richtext plugin)?

buddh4:
Similar to facebook hashtags users can use hashtags as #summer #newyear in the richtext which will link to a dashboard like stream with all content also tagged with this hashtag.

Carl2213:
Hashtags are like topics, only universal. Currently, the topics feature is contained within the content container of the stream (the space or user's profile).

Hashtags would function exactly the same way but on a global level. This would help with discoverability and curating stream content. Hashtags are also easier to use. Simply #summer instead of adding a topic to the post (hashtags can also be triggered in comments) or having to manage topics.

I think it would be possible to rework the Topics feature into Hashtags. Remove the content container limit to make them global, then add #logic to the richtext in order to automatically create a search/stream filter with the hashtag when the # character is detected.

Content from the user's profile or joined spaces can be joined at the top. Further filters (friends/spaces) can further customize the stream/search results.

Carl2213:
Further optional extensions:

- calculate the number of mentions of a particular hashtag to feed a #Trending widget with the most used/discussed hashtags
- Like mentioning, auto suggest hashtags in the richtext when the # character is typed, along with the current usage number of each hashtag

buddh4:
Also the topic feature can be used as moderator only tool to tag and categorize content.

Carl2213:
Maybe both can be kept, but the underlying logic can be copied and modified and renamed to hashtags?

ArchBlood:
@buddh4 on the topic of topics feature should also be able to hide topics as well, but another issue should be opened for that :thinking:

buddh4:
@ArchBlood not sure what you mean, but topics should be rather hidden for users who can't create topics unless there are topics available in a space.

For me the differences between hashtags and topics are:

- Hashtag are user generated and not moderated
- Hashtags are added directly in a richtext field and are rendered as simple links
- Hashtags are global (not container related) 
- There probably will be more hashtags than topics
[...]

PR #6822, opened 2024-01-29
Enh: Trending Topics

ArchBlood:
As seen in the community site: [here](https://community.humhub.com/s/general-discussion/post/post/view?id=16077)
Screenshot_1

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] All tests are passing
- [ ] Changelog was modified

**Other information:**

### DIFF
--- a/protected/humhub/modules/dashboard/views/dashboard/index.php
+++ b/protected/humhub/modules/dashboard/views/dashboard/index.php
@@ -4,6 +4,7 @@
  * @var bool $showProfilePostForm
  */
 
+use humhub\modules\topic\widgets\TrendingTopics;
 use humhub\modules\activity\widgets\ActivityStreamViewer;
 use humhub\modules\dashboard\widgets\DashboardContent;
 use humhub\modules\dashboard\widgets\Sidebar;
@@ -27,7 +28,11 @@ use humhub\libs\Html;
                 [
                     ActivityStreamViewer::class,
                     ['streamAction' => '/dashboard/dashboard/activity-stream'],
-                    ['sortOrder' => 150]
+                    ['sortOrder' => 150],
+                ],
+                [
+                    TrendingTopics::class,
+                    [],
                 ]
             ]
         ]);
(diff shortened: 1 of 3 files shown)


luke-:
Wouldn't that be better than small modules? This could also be useful in some Spaces.

ArchBlood:
> Wouldn't that be better than small modules? This could also be useful in some Spaces. 

I thought of doing that, but thought it would have been better being a part of the core for ease of extending from other modules.

ArchBlood:
> Hmm, in which cases does it make sense to extend this?

- Topic logic that implements variants to Topics;

In some cases where Topics are actually used, requests for Topic variants are also requested, such as Hashtags, this being a part of the core would mean that this could be extended from the Topic variant instead of having to have multiple modules installed, case in point would be the blog module in humhub-contrib, this module requires both the blog module and the custom pages module to be installed to be used.

I can still create this in a module if the HumHub team thinks that would be a better option, however I can't say much about making it extendable for other modules to use. :thinking:

ArchBlood:
@luke- as of your advice I've created the module, currently I can't release it on the marketplace due to PHP restrictions being PHP 8+ for the module and knowing some people are still using PHP 7.4;

So I'll close this P/R in favor of moving it into a module.

https://github.com/GreenMeteor/trending

ArchBlood:
> @ArchBlood Ok great, maybe a more speaking module id would be fine: e.g. `trending-topics`
I did have that thought at first, but also wanted to keep the module id simple and not with a `-` if possible, I had also thought of using `trendingTopic` but decided against this option.

The overall goal now is the trending part of the module, not necessarily the Topics part, I was thinking of making it so that it could be extended in ways that not only Topics were displayed but the option could be configurable trends. 🤔

ArchBlood:
Another thought, what would HumHub's stance be on changing the icon of the Topics module to `fa-hashtag` instead of `fa-star`?
Screenshot_1

This would work as a better alternative for the growing request for #3936

PR #6872, opened 2024-02-29
Fix #6871: Cache is not deleted after deleting a value in the settings

### LINKED ISSUE #6871: Cache is not deleted after deleting a value in the settings

Common to the 3 following controllers:
```php
$moduleUserSettings = Yii::$app->getModule('my-module')->settings->user(Yii::$app->user->identity);
```

Controller 1:
```php
$moduleUserSettings->set('test-value', true);
```

Controller 2:
```php
$moduleUserSettings->delete('test-value');
```

Controller 1:
```php
$val = $moduleUserSettings->get('test-value');
var_dump($val);
```

`null`

### What do you get instead?

`1`

### Additional info

Related to https://github.com/humhub/humhub/issues/6189 ?

| Q                | A
| ---------------- | ---
| HumHub version   | 1.15.3

marc-farre:
- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**

Issue https://github.com/humhub/humhub/issues/6871

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,9 @@ HumHub Changelog
 
 1.15.4 (Unreleased)
 -------------------
-- Fix #6851: Update HumHub Prosemirror (fixes legacy emoji support) 
+- Fix #6851: Update HumHub Prosemirror (fixes legacy emoji support)
 - Fix #6859: Hide scheduled and draft content on other user profile pages
+- Fix #6871: Cache is not deleted after deleting a value in the settings
 
 1.15.3 (February 8, 2024)
 -------------------------
--- a/protected/humhub/libs/BaseSettingsManager.php
+++ b/protected/humhub/libs/BaseSettingsManager.php
@@ -15,7 +15,6 @@ use Yii;
 use yii\base\Component;
 use yii\base\InvalidArgumentException;
 use yii\base\InvalidConfigException;
-use yii\db\conditions\LikeCondition;
 use yii\db\StaleObjectException;
 use yii\helpers\Json;
 
@@ -81,7 +80,7 @@ abstract class BaseSettingsManager extends Component
         }
 
         if ($value === null) {
-             $this->delete($name);
+            $this->delete($name);
             return;
         }
 
@@ -194,6 +193,8 @@ abstract class BaseSettingsManager extends Component
         if (isset($this->_loaded[$name])) {
             unset($this->_loaded[$name]);
         }
+
+        $this->invalidateCache();
     }
 
     /**


ArchBlood (approved):

marc-farre:
Thanks @martin-rueegg for the explanations.
I cannot reproduce on my local instance the issue I explain on https://github.com/humhub/humhub/issues/6871
So I think that your code might be working.

However, I clearly had the issue on a production instance. The setting was removed from the database (I've checked with phpMyAdmin) and even after `$moduleUserSettings->delete('test-value');` I still had the old value with `$moduleUserSettings->get('test-value');`.

After purging the cache of the instance, `$moduleUserSettings->get('test-value');` was returning null.
So even if I cannot reproduce, I'm sure there is a bug with the cache when deleting a setting.

I also think the code should be easy to understand.
In `BaseSettingsManager.php`, the `invalidateCache()` is triggered in the `set()` method.
So I think it should also be triggered in the `delete()` method.

Otherwise, it's quite difficult to understand that for the `delete()` method the cache is deleted somewhere else, in another file.
The more simple is the code, the less bug we have!

martin-rueegg:
Thanks @marc-farre, for your comments.

> The more simple is the code, the less bug we have!

Yes and no. Simplicity is often also a question of perspective. I would argue, that in OOP, simplicity is defined by placing the functionality as close to the source as possible. In the current case, however, there is a "problem" insofar as the cache entry combines multiple values (all settings from one `module_id`), whereas every value us represented by single database record.

So for me, the perspective was that deleting a record using it's default model should also invalidate the cache, as the model can be used to manipulate records even without the settings manager. I agree, this should also apply to updates. Then, only direct SQL-queries would not touch the cache. But if someone imitates the `setting` model, he or she would see in the model's code that also the cache needs to be cleared. If the clearing of the cache was left only to the settings manager, this would not be transparent. The code would be in another file, for that matter.

Your solution might fix your current issue. But you do not know the actual source of the issue. Hence, the bug still exists. For example, could it be that the value is deleted in the db (without using the `setting` model) before being deleted in the settings manger?

Generally, I would agree to the statement of "more simple is the code, the less bug we have". However, I would add, that the more tests we have, the less bugs we have or they will surface more quickly!

Looking at the test, I realize that [`testGetCached()`](https://github.com/humhub/humhub/blob/05097e881fc341fa3fda4fcaa446da1984b1ea46/protected/humhub/tests/codeception/unit/components/SettingsManagerTest.php#L317) does not test deletion. And [testDelete()](https://github.com/humhub/humhub/blob/05097e881fc341fa3fda4fcaa446da1984b1ea46/protected/humhub/tests/codeception/unit/components/SettingsManagerTest.php#L248) is not cached. So I think it would be worth enhancing the tests to check both update and deletion of cached values (both in the db as well as in the cache).

marc-farre:
> placing the functionality as close to the source as possible

Yes, that's fine.

But the complexity for me is to have cache managed in 2 different places (in `BaseSettingsManager` and `SettingActiveRecord`).
And mainly for the `BaseSettingsManager::set()` and `BaseSettingsManager::delete()` methods which doesn't refresh the cache the same way.

> could it be that the value is deleted in the db (without using the setting model) before being deleted in the settings manger

No. It was deleted via `Yii::$app->getModule('my-module')->settings->user(Yii::$app->user->identity)->delete('test-value');` (I'm sure of this). And it worked, because I've checked the DB, and the value was gone.

luke-:
Thank you guys for fixing the bug.

> I've not looked into this in further detail, but I noticed that you have undone the change from the above PR (#6190). Which implies that cleaning up the cache in `SettingActiveRecord` seems not to be sufficient.
> 
> Do you encounter the bug only when [`$record`](https://github.com/humhub/humhub/pull/6872/files#diff-cc3085d54fa643468beac823005c268bbb78e9cdf279cf096db4d0be3cff3ef1R182) is not found or an [error](https://github.com/humhub/humhub/pull/6872/files#diff-cc3085d54fa643468beac823005c268bbb78e9cdf279cf096db4d0be3cff3ef1R186-R190) is encountered? As in any other case, `$record->delete();` should already delete the cache.

A quick question, am I blind, or where is or where should the cache be cleared in the `BaseSettingsManager::delete()` method? https://github.com/humhub/humhub/blob/master/protected/humhub/libs/BaseSettingsManager.php#L181

---

I haven't looked any deeper into the topic yet, but of course it would be nicer if the caching is only done in one place.

I think that calls to `Setting::deleteAll()` are more legacy code and could be replaced by `SettingsManager::deleteAll()`. We could then say/define, all settings MUST be accessed/manipulated using the SettingsManager (not directly using ARs).  Wouldn't we then have caching in one place? With the exception that the cache key comes from the AR.

marc-farre:
@martin-rueegg @luke-  To summarize:
- `$this->invalidateCache();` was removed from `BaseSettingsManager::delete()` in PR  https://github.com/humhub/humhub/pull/6190/files#diff-cc3085d54fa643468beac823005c268bbb78e9cdf279cf096db4d0be3cff3ef1L170
- A new cache was added to `SettingActiveRecord` in same PR: https://github.com/metaworx/humhub/blob/3b4c44176d01612370b54ba87c9da4a1edacbb58/protected/humhub/components/SettingActiveRecord.php#L30
- But this new cache doesn't use the same Key

So we are speaking about 2 different caches in 2 different classes.

This PR puts back the cache deletion to `BaseSettingsManager` when deleting a setting, for the cache key of this class.

@martin-rueegg do you see a better idea?
As what I understand, you wanted to move the cache from `BaseSettingsManager` to `SettingActiveRecord`, but currently we have both with different keys.

luke- (approved):
@marc-farre Thanks, I would merge the PR now, if we find a better solution later, we can do this  afterwards in a separate PR. But for me it's clear,  that the cache clear is missing here.

martin-rueegg:
@marc-farre

> * `$this->invalidateCache();` was removed from `BaseSettingsManager::delete()` in PR  https://github.com/humhub/humhub/pull/6190/files#diff-cc3085d54fa643468beac823005c268bbb78e9cdf279cf096db4d0be3cff3ef1L170
> * A new cache was added to `SettingActiveRecord` in same PR: https://github.com/metaworx/humhub/blob/3b4c44176d01612370b54ba87c9da4a1edacbb58/protected/humhub/components/SettingActiveRecord.php#L30
> * But this new cache doesn't use the same Key
> 
> So we are speaking about 2 different caches in 2 different classes.

This is actually incorrect. The key is now _always_ generated by `SettingActiveRecord::getCacheKey()` based on the `SettingActiveRecord::CACHE_KEY_FORMAT`. So it is the same cache, using the same key.

But what has changed is, that the cache was now deleted in `SettingActiveRecord::deleteAll()`, rather than `BaseSettingsManager::delete()`, since the former is called by the latter (please note, that `\yii\db\ActiveRecord::delete()` calls `ActiveRecord::deleteInternal()` which calls `ActiveRecord::deleteAll()` - or `SettingActiveRecord::deleteAll()` in this case - in turn).

Which begs the question, why would the cache not be deleted, if `BaseSettingsManager::delete()` was called!?

As mentioned before, there can only be two reasons:

1. The given record was no longer in the db, and hence [not found](https://github.com/metaworx/humhub/blob/3b4c44176d01612370b54ba87c9da4a1edacbb58/protected/humhub/libs/BaseSettingsManager.php#L156), or
2. The deletion [failed](https://github.com/metaworx/humhub/blob/3b4c44176d01612370b54ba87c9da4a1edacbb58/protected/humhub/libs/BaseSettingsManager.php#L159) with an exception,

... and hence the above mentioned `deleteAll()` is not called.

That's why I said, it would be better to fix the test and see where the actual error comes from. With the current solution, you might seemingly "fix" the symptom, but by doing so, the true error gets masked.

> @martin-rueegg do you see a better idea? As what I understand, you wanted to move the cache from `BaseSettingsManager` to `SettingActiveRecord`, but currently we have both with different keys.

- Firstly, the cache key is the same  - as explained above
- Secondly, I personally would 
  - leave the _creation_ of the cache in `BaseSettingsManager` as it loads (and contains) all keys of a given module at once. So that's closest to it's creation and usage.
  - clear the cache in `SettingActiveRecord`, but not only on delete (`deleteAll()`), but also on update (`updateAll()`). Which again is closest to the effect that requires the cache to be cleared.

That was indeed my initial plan. But apparently, I left the update part untouched.


> @marc-farre (...) But for me it's clear, that the cache clear is missing here.

Not really. As I tried to show above. :-) Clearing the cache is implemented. The bug is, that for whatever reason, it's not executed. Could be, that the module in question is overwriting or hooking into the `beforeDelete()` event or something else. This is not clear. Hence, again, I think it would be wiser to update the tests and find the real cause.

luke-:
> @luke-
> 
> > (...) But for me it's clear, that the cache clear is missing here.
> 
> Not really. As I tried to show above. :-) Clearing the cache is implemented. The bug is, that for whatever reason, it's not executed. Could be, that the module in question is overwriting or hooking into the `beforeDelete()` event or something else. This is not clear. Hence, again, I think it would be wiser to update the tests and find the real cause.

I can't see any line either in `SettingsManager::delete()` or `SettingsActiveRecord::[before|after]delete()` (before this PR) that should clear the cache.  Am I missing something here?

martin-rueegg:
@luke- 

It is in `SettingActiveRecord::deleteAll()`:

https://github.com/humhub/humhub/blob/99cb0fa7d11254d81a562e05069a66d114f5bad4/protected/humhub/components/SettingActiveRecord.php#L56-L60


- `SettingActiveRecord::deleteAll()` is called by
  - `ActiveRecord::deleteInternal()` which is called by
    -  `ActiveRecord::delete()` which is called by
      - `BaseSettingsManager::delete()`:
        https://github.com/humhub/humhub/blob/99cb0fa7d11254d81a562e05069a66d114f5bad4/protected/humhub/libs/BaseSettingsManager.php#L180-L185

luke-:
@martin-rueegg Thanks for the fast reply.

Yes, the cache is deleted correctly via your example path.
But if the path starts with `SettingsManager::delete()`, the cache is not deleted at all. But the PR has fixed this.

martin-rueegg:
@luke- 

> Yes, the cache is deleted correctly via your example path. But if the path starts with `SettingsManager::delete()`, the cache is not deleted at all. But the PR has fixed this.

But `\humhub\components\SettingsManager` extends `\humhub\libs\BaseSettingsManager`. And since `SettingsManager` is not overloading the `delete()`-method, it _is_ actually the `BaseSettingsManager::delete()` that is called, isn't it?

martin-rueegg:
As a matter of fact, _this_ PR also "fixes" `BaseSettingsManager::delete()`, not `SettingsManager` directly ...

luke-:
Yes, that's right, it's about `BaseSettingsManager`.

But when `BaseSettingsManager::delete()` was called, there was no flush in the cache.

martin-rueegg:
@luke- 

> But when `BaseSettingsManager::delete()` was called, there was no flush in the cache.

I don't think this is valid as a general statement. But there seem to be situations, where the cache is not cleared. I have shown in my previous messages, what could be the reason for this. 

Again, I think updating the tests would help to verify this. See last paragraph of [this comment](https://github.com/humhub/humhub/pull/6872#issuecomment-1971433832).

Unfortunately, right now I don't have the time to enhance the test.

luke-:
@martin-rueegg Ok thanks, I need some time to dig deeper into this.

marc-farre:
> there can only be two reasons:
> 
> 1. The given record was no longer in the db, and hence [not found](https://github.com/metaworx/humhub/blob/3b4c44176d01612370b54ba87c9da4a1edacbb58/protected/humhub/libs/BaseSettingsManager.php#L156), or
> 2. The deletion [failed](https://github.com/metaworx/humhub/blob/3b4c44176d01612370b54ba87c9da4a1edacbb58/protected/humhub/libs/BaseSettingsManager.php#L159) with an exception,

Yes, when I had the issue, the record was deleted in the DB.
It explains why when `BaseSettingsManager::delete()` was executed, the cache wasn't deleted.

However, the record was deleted by a previous `BaseSettingsManager::delete()` (because I know my code and I manage this instance, so it cannot be from another reason): https://github.com/humhub/humhub/blob/master/protected/humhub/libs/BaseSettingsManager.php#L185
And I don't have any exception.

So I think that https://github.com/humhub/humhub/blob/master/protected/humhub/libs/BaseSettingsManager.php#L197 still makes sense, because:
- it's easier to understand than the logic that explains it's deleted at the end from `SettingActiveRecord::deleteAll()`
- if a similar issue arrives, when `BaseSettingsManager::delete()` is executed again, the cache is cleared the second time

ArchBlood:
> @martin-rueegg You're right. I've done a test, and `SettingActiveRecord::deleteAll()` is called which deletes the cache with the same key because `BaseSettingsManager::getCacheKey()` gets the key from `SettingActiveRecord` class: https://github.com/humhub/humhub/blob/master/protected/humhub/libs/BaseSettingsManager.php#L244
> 
> So I cannot explain the issue I had, and maybe this PR is useless.
> 
> However:
> 
> * I'm sure I had this issue (I've done some tests when it was happening), so they might be some cases where `array_walk` doesn't deletes the cache: https://github.com/humhub/humhub/blob/99cb0fa7d11254d81a562e05069a66d114f5bad4/protected/humhub/components/SettingActiveRecord.php#L56-L60
> * I think the way the cache is managed is complicated to understand, and I think it shouldn't be done in both `BaseSettingsManager` and `SettingActiveRecord` classes (E.g, `BaseSettingsManager::set()` deletes the cache in `BaseSettingsManager`, but `BaseSettingsManager::delete()` deletes the cache from `SettingActiveRecord`)
> 
> @luke- should I create an issue to simplify the way cache is managed for settings?

After reading this it's sounding like a race condition where cache is being modified before the execution of the code, maybe something simple like this could help?

```php
array_walk($modulesOrContainers, static function (ActiveRecord $rec) {
    try {
        $key = static::getCacheKey(...array_values($rec->toArray())); 
        if (Yii::$app->cache->exists($key)) {
            Yii::$app->cache->delete($key);
        } else {
            Yii::warning("Cache entry with key '$key' does not exist.");
        }
    } catch (Exception $e) {
        Yii::error("Error deleting cache entry: " . $e->getMessage());
    }
});
```

Issue #6874, opened 2024-03-06
Use OpenSans from Composer


luke-:
- Update OpenSans to latest version if available
- Include 500 weight
- Use from Composer as Asset https://www.npmjs.com/package/@fontsource/open-sans?activeTab=readme

yurabakhtin:
@luke- Done in PR https://github.com/humhub/humhub/pull/6876.

PR #6876, opened 2024-03-07
Update open-sans font from composer

yurabakhtin:

- [x] Feature

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
https://github.com/humhub/humhub/issues/6874


luke- (changes requested):
@yurabakhtin Can't we then also delete the `static/fonts/` directory?

Which font weights are imported?  If I have understood correctly, the available weights must be imported individually? Like before in open-sans.css

yurabakhtin:
> Can't we then also delete the `static/fonts/` directory?

@luke- Ah yes, I forgot to do this in first commit, fixed in https://github.com/humhub/humhub/pull/6876/commits/f2b8d438524a3a29a1474e310a5a30e468ffe2ac.

> Which font weights are imported?

- Latin `font-style: normal` with `font-weight: 300-800` - file `latin.css`
- Latin `font-style: italic` with `font-weight: 300-800` - file `latin-italic.css`

>  If I have understood correctly, the available weights must be imported individually? Like before in open-sans.css

I saw the old file `open-sans.css` had fonts only for normal and latin with weight from 300 to 800, it is why I have to include the 2 files `latin,css` and `latin-italic.css`. It would be better if the package has a single file where the fonts were included but I don't find such one there.

I am not sure why each weight must be imported individually, yes the package has a separate file per each weight, but better to load them all from single css file because when page is loading each file is one http requests, so better keep all css in single bundled file in order to decrease a number of loaded external files such as css or js.

luke- (approved):

PR #6918, opened 2024-04-04
Add `composer.lock` to `.gitignore`

marc-farre:


luke-:
@marc-farre Why should the `composer.lock` be ignored? I would prefer to keep it as it is.

marc-farre:
OK. So I close this PR.
On my side, it prevents me from switching branch after a `composer update` (I need to to a `git stash` because the `composer.lock` has changed).
But maybe I'm not doing it the right way.
Anyway, it's not important.

PR #6920, opened 2024-04-04
Enhancing of meta search for advanced search providers

yurabakhtin:

- [x] Feature

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**
Related to https://github.com/humhub/advanced-search/issues/1


luke- (approved):

ArchBlood:
Page Not Found for `/people` should return `/user/people`

protected/humhub/widgets/MetaSearchWidget.php
```diff
@@ -65,4 +66,30 @@ public function addProvider($provider)
     {
         $this->providers[] = $provider;
     }
+
+    /**
+     * @return MetaSearchProviderInterface[]
+     */
+    protected function getSortedProviders(): array
```
luke-, line 73: private?

protected/humhub/interfaces/MetaSearchProviderInterface.php
```diff
@@ -63,4 +70,11 @@ public function getService(): MetaSearchService;
      * @return string|null
      */
     public function getKeyword(): ?string;
+
+    /**
+     * Get additional params which may be used to initialize the provider
+     *
+     * @return array|null
+     */
+    public function getParams(): ?array;
```
luke-, line 79: I don't quite understand what we need the parameter for.

Couldn't it just be integrated into `getRoute()`? 

If it is needed to identify different `AdvancedServiceProvider` instances, couldn't we also use the route here?

protected/humhub/services/MetaSearchService.php
```diff
@@ -97,6 +107,16 @@ public function getTotal(): int
         return isset($this->totalCount) ? (int) $this->totalCount : 0;
     }
 
+    /**
+     * Get total number results as short integer number
+     *
+     * @return string
+     */
+    public function getShortTotal(): string
```
luke-, line 115: Why don't we use the formatter directly here?

luke- (changes requested):

protected/humhub/services/MetaSearchService.php
yurabakhtin: @luke- Do you mean to remove the merhod `getShortTotal()` and keep only the meyhod `getTotal()` with new code like this?

```php
public function getTotal(): string
{
    return Yii::$app->formatter->asShortInteger($this->totalCount);
}
```

Really I thought to keep the method `getTotal()` which returns now only the value from the protected property `$this->totalCount` for some external access case in future. But if you think we don't need it I can remove the `getShortTotal()`, let me know.

protected/humhub/interfaces/MetaSearchProviderInterface.php
yurabakhtin: @luke- 

> Couldn't it just be integrated into `getRoute()`?

No, the `getRoute()` is used only for link "Seach all results":

- Content: `/content/search`
- Space: `/space/spaces`
- User: `/user/people`
- Advanced Search (Google): `https://www.google.com/search?q=%keyword%`

> If it is needed to identify different `AdvancedServiceProvider` instances, couldn't we also use the route here?

When the meta searching is run we send an AJAX request per each provider:

request-provider

the data `provider` and `params` are taken from html attributes of each provider:

html-provider

before the module Advanced Search Providers we didn't need the additional params because each provider has unique class name, but now we need to know what Advanced Search Provider is really requested, i.e. we need now GET params `provider` + `params` in order to init a search provider on the controller `humhub\controllers\MetaSearchController` side.

A search provider is built [here](https://github.com/humhub/humhub/blob/enh/advanced-meta-search/protected/humhub/widgets/MetaSearchProviderWidget.php#L45-L49):

```php
$this->provider = Yii::createObject([
    'class' => $this->provider,
    'params' => $this->params,
    'keyword' => $this->keyword
]);
```

then Advanced Search Provider is initialized [here](https://github.com/humhub/advanced-search/blob/main/models/AdvancedSearchProvider.php#L111) by ID:

```php
$this->provider = Provider::findOne(['id' => (int) $this->params['id']]);
```

I.e. we need pass some unique data for the AJAX request, and I am not sure we can use there a result from `getRoute()`.

protected/humhub/widgets/MetaSearchWidget.php
yurabakhtin: @luke- Yes, it may be updated to `private` without any problems for now, but I did it as `protected` for future case if the class will be used as parent class from some external module so it would be good allow to run `$this->getSortedProviders()` from a child class.
But if you think we don't need this I can make it private, just please tell me, thanks.

luke-: @yurabakhtin Let's make it "private" for now. When we need it we can change the visibility to "protected".

protected/humhub/services/MetaSearchService.php
luke-: @yurabakhtin I thought we could directly use the method 
`Yii::$app->formatter->asShortInteger($this->totalCount)` instead of a extra method. 
 But it's not that important. We can leave it as it is.

protected/humhub/widgets/views/metaSearchProvider.php
```diff
@@ -16,7 +16,7 @@
     <div class="search-provider-title">
         <?= $provider->getName() ?>
         <?php if ($provider->getService()->isSearched()) : ?>
-            <?= Html::tag('span', '(' . $provider->getService()->getTotal() . ')') ?>
+            <?= Html::tag('span', '(' . Yii::$app->formatter->asShortInteger($provider->getService()->getTotal()) . ')') ?>
```
luke-, line 19: @yurabakhtin Yes, that's what I meant :-)

protected/humhub/interfaces/MetaSearchProviderInterface.php
yurabakhtin: or allow only `id` as additional param:

- `<li data-provider="humhub\modules\advanced_search\search\SearchProvider:1">...</li>`

```php
$provider = explode(':', $this->provider, 2);
$params = [
    'class' => $provider[0],
    'keyword' => $this->keyword,
];
if (!empty($provider[1])) {
    $params['id'] = $provider[1];
}
$this->provider = Yii::createObject($params);
```

luke-: @yurabakhtin Thank you for your explanation.

> > Wouldn't it also be a solution:
> > 
> > * Instead of the `getParams()` value, just use the `getRoute()` in all places. (e.g. in the Ajax request, Data Attributes)
> > * `AdvancedServiceProvider` could make a unique route for each provider by `/advanced-search/url?id=1` and then redirect to the actual route. This would ensure that each AdvancedSearchProvider has a unique route.
> > 
> > This would allow us to do without `getParam()`. I really only see a use for the `advanced-search` module here.
> 
> @luke- I have investigated and I think it will be a not good code on the a parsing side of the Advanced Search Provider, I try to explain why:
> 
> There is the widget [MetaSearchProviderWidget](https://github.com/humhub/humhub/blob/enh/advanced-meta-search/protected/humhub/widgets/MetaSearchProviderWidget.php). It has the property `$provider` which can be pass as either object `MetaSearchProviderInterface` or as `string`.
> 
> * `MetaSearchProviderInterface`: this case for normal page loading, i.e. here we init it as `MetaSearchProviderWidget::widget(['provider' => new ContentSearchProvider()])` or `MetaSearchProviderWidget::widget(['provider' => AdvancedSearchProvider::findOne(['id' => 123])])`
> * `string`: Ajax case loading, i.e. `'provider' => 'humhub\modules\content\search\ContentSearchProvider'`, and here we need to build new object `MetaSearchProviderInterface` from the passed string, it is ok for core unique providers, but for Advanced Providers we need at least new additional param `id`, please see how we create the provider object [here](https://github.com/humhub/humhub/blob/enh/advanced-meta-search/protected/humhub/widgets/MetaSearchProviderWidget.php#L45-L49):
> 
> ```
> $this->provider = Yii::createObject([
>     'class' => $this->provider,
>     'params' => $this->params,
>     'keyword' => $this->keyword,
> ]);
> ```
> 
> If we will pass a route value like this: `'route' => '/advanced-search/url?id=1'` instead of `'params' => '{"id":1}'`, then [here](https://github.com/humhub/advanced-search/blob/main/search/SearchProvider.php#L112):
> 
> ```
> $this->provider = Provider::findOne(['id' => (int) $this->params['id']]);
> ```
> 
> we should parse the passed route string `/advanced-search/url?id=1` in order to extract the `id`, yes it is possible but I think it will looks like a dirty solution.
> 

I understand that it is a rather dirty solution from the point of view of the `advanced-search` module. But from a core perspective, I think a `getParams()` which is probably only necessary in the `advanced-search` module is also not that ideal.

```php 
    /**
     * Finds Provider by URL
     * This is necessary to distinguish between several instances of AdvancedSearchProvider.
     */
    private function getProviderByUrl(string $url): ?Provider {
        $id = str_replace(Url::to(['/advanced-search/url', 'id' => null]), '', $url);
        return Provider::findOne(['id' => intval($id)]);
    }
```

> > If this is not possible, I would prefer a `getIdentifier()` method. Which should also work, right?
> 
> For core providers we keep an identifier as is, like `humhub\modules\content\search\ContentSearchProvider`, but for advanced we could set it as `humhub\modules\advanced_search\search\SearchProvider:{"id":1}`. Html attibutes:
> 
> * `<li data-provider="humhub\modules\advanced_search\search\SearchProvider:{"id":1}">...</li>`
> * `<li data-provider="humhub\modules\content\search\ContentSearchProvider">...</li>`
> 
> Then [here](https://github.com/humhub/humhub/blob/enh/advanced-meta-search/protected/humhub/widgets/MetaSearchProviderWidget.php#L45-L49) we could init provider like this:
> 
> ```
> $provider = explode(':', $this->provider, 2);
> $params = [
>     'class' => $provider[0],
>     'keyword' => $this->keyword,
> ];
> if (!empty($provider[1])) {
>     $params = array_merge($params, json_decode($provider[1]));
> }
> $this->provider = Yii::createObject($params);
> ```

I think you misunderstood me here.

My alternative suggestion would be to simply rename `getParams()`.

For example, we could rename it from
`public function getParams(): ?array;` to `public function getMultipleInstanceIdentifier(): ?int`.

For me, it would be clearer from the naming.
Core providers and all others except presumably `advanced-search` could still return `null`.

yurabakhtin: @luke- I have implemented your solution with removing the method `getParams()` completely and use only `getRoute()` for unique detecting, I just made the result of the method as `string` or `array`, so we don't need the dirty `str_replace` because for advanced search provider the `route` param is sent as array:

```
provider: humhub\modules\advanced_search\search\SearchProvider
keyword: two
route[0]: /advanced_search/url
route[id]: 1
```

for core providers it is passed as `string`:

```
provider: humhub\modules\content\search\ContentSearchProvider
keyword: two
route: /content/search
```

Commits https://github.com/humhub/humhub/pull/6920/commits/4f21e8e823e16d378c0aed994c1ca4f8d4a610d1 and https://github.com/humhub/advanced-search/commit/cfefcc8a67cc616b1c2405ac170d65115be7696b.

So now the code looks much better than with `getParams()`, thanks again for the advices.

luke-: @yurabakhtin Thanks, I like the code much better from a core perspective.

Issue #5419, opened 2021-11-04
Lock search index while reindexing...


luke-:
### Is your feature request related to a problem? Please describe.

While reindexing the search index via console, new entries may added by queue processing.

See: https://github.com/humhub/humhub/issues/3847#issuecomment-958712895

### Describe the solution you'd like

Set some sort of lock.

### Describe alternatives you've considered

-

funkycram:
It also can happen with the `UpdateDocument` jobs executed by queue/run when there are very numerous in the queue: ` File '/home/web/www/humhub/protected/runtime/searchdb/_oat.fdx' is not readable.`:
```
2023-03-15 11:48:01 [361965] humhub\modules\search\jobs\UpdateDocument (attempt: 1, pid: 1756612) - Done (0.119 s, 14.06 MiB)
PHP Fatal error:  Uncaught ZendSearch\Lucene\Exception\InvalidArgumentException: File '/home/web/www/humhub/protected/runtime/searchdb/_oat.fdx' is not readable. in /home/web/www/humhub/protected/vendor/matthewbdaly/zendsearch/src/ZendSearch/Lucene/Storage/File/Filesystem.php:46
Stack trace:
#0 /home/web/www/humhub/protected/vendor/matthewbdaly/zendsearch/src/ZendSearch/Lucene/Storage/Directory/Filesystem.php(354): ZendSearch\Lucene\Storage\File\Filesystem->__construct()
#1 /home/web/www/humhub/protected/vendor/matthewbdaly/zendsearch/src/ZendSearch/Lucene/Index/SegmentWriter/AbstractSegmentWriter.php(611): ZendSearch\Lucene\Storage\Directory\Filesystem->getFileObject()
#2 /home/web/www/humhub/protected/vendor/matthewbdaly/zendsearch/src/ZendSearch/Lucene/Index/SegmentWriter/DocumentWriter.php(211): ZendSearch\Lucene\Index\SegmentWriter\AbstractSegmentWriter->_generateCFS()
#3 /home/web/www/humhub/protected/vendor/matthewbdaly/zendsearch/src/ZendSearch/Lucene/Index/Writer.php(772): ZendSearch\Lucene\Index\SegmentWriter\DocumentWriter->close()
#4 /home/web/www/humhub/protected/vendor/matthewbdaly/zendsearch/src/ZendSearch/Lucene/Index.php(1199): ZendSearch\Lucene\Index\Writer->commit()
[...]

PR #6934, opened 2024-04-08
Lock search index while reindexing

yurabakhtin:

- [x] Bugfix

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
https://github.com/humhub/humhub/issues/5419

### DIFF
--- a/CHANGELOG-DEV.md
+++ b/CHANGELOG-DEV.md
@@ -4,6 +4,7 @@ HumHub Changelog
 1.16.0-beta.2 (TBD)
 -----------------------------
 - Fix #6928: Disabled users without "ignoredDN" settings
+- Fix #5419: Lock search index while reindexing
 
 1.16.0-beta.1 (April 5, 2024)
 -----------------------------
--- a/protected/humhub/modules/content/jobs/SearchDeleteDocument.php
+++ b/protected/humhub/modules/content/jobs/SearchDeleteDocument.php
@@ -1,15 +1,19 @@
 <?php
+/**
+ * @link https://www.humhub.org/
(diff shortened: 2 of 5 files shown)


luke- (changes requested):
@yurabakhtin Is there a waiting time between retries?

yurabakhtin:
> Is there a waiting time between retries?

@luke- The content search jobs are extended from `LongRunningActiveJob` which uses [the value](https://github.com/humhub/humhub/blob/develop/protected/humhub/modules/queue/Module.php#L26) for ttr: `$longRunningJobTtr = 60 * 60`.

yurabakhtin:
@luke- Maybe do you mean [RetryAcquireTrait](https://www.yiiframework.com/doc/api/2.0/yii-mutex-retryacquiretrait)?
But I see it is used only for `FileMutex` and `PgsqlMutex` but currently the `MysqlMutex` is used in [config](https://github.com/humhub/humhub/blob/master/protected/humhub/config/common.php#L170-L172).

luke-:
@yurabakhtin Yes, I mean something like the Mutex Retry Timeout.

`Ttr` just seems to be the maximum execution time.

With your solution I am afraid that the retry will happen immediately (e.g. if there is a queue with workers).

Ideally a time of e.g. 5 minutes passes until the next retry.

yurabakhtin:
@luke- 

> Yes, I mean something like the Mutex Retry Timeout.

Done in the commit https://github.com/humhub/humhub/pull/6934/commits/e810cb8e0576b54dec672abee590471d98c41f32.

luke- (approved):

PR #6941, opened 2024-04-09
Fix: Attempt to read property "jsWidget" on null

### LINKED ISSUE #6940: Attempt to read property "jsWidget" on null
Currently I'm not sure where to start;

I try accessing the new search function then click "Show all results" for content and get a error.

The search should be done correctly.

### What do you get instead?
```logs
yii\base\ErrorException: Attempt to read property "jsWidget" on null in /home/USER/SITE/protected/humhub/modules/content/widgets/stream/StreamEntryWidget.php:109
Stack trace:
#0 /home/USER/SITE/protected/humhub/modules/content/widgets/stream/StreamEntryWidget.php(109): yii\base\ErrorHandler->handleError()
#1 /home/USER/SITE/protected/humhub/modules/content/widgets/stream/StreamEntryWidget.php(81): humhub\modules\content\widgets\stream\StreamEntryWidget::renderLegacyWallEntry()
#2 /home/USER/SITE/protected/humhub/modules/content/controllers/SearchController.php(69): humhub\modules\content\widgets\stream\StreamEntryWidget::renderStreamEntry()
#3 /home/USER/SITE/protected/humhub/modules/content/controllers/SearchController.php(49): humhub\modules\content\controllers\SearchController->renderResults()
#4 [internal function]: humhub\modules\content\controllers\SearchController->actionResults()
#5 /home/USER/SITE/protected/vendor/yiisoft/yii2/base/InlineAction.php(58): call_user_func_array()
#6 /home/USER/SITE/protected/vendor/yiisoft/yii2/base/Controller.php(179): yii\base\InlineAction->runWithParams()
#7 /home/USER/SITE/protected/vendor/yiisoft/yii2/base/Module.php(553): yii\base\Controller->runAction()
#8 /home/USER/SITE/protected/vendor/yiisoft/yii2/web/Application.php(104): yii\base\Module->runAction()
#9 /home/USER/SITE/protected/vendor/yiisoft/yii2/base/Application.php(385): yii\web\Application->handleRequest()
#10 /home/USER/SITE/index.php(28): yii\base\Application->run()
#11 {main}
```
### Additional info

| Q                | A
| ---------------- | ---
| HumHub version   | 1.16.0-beta.2
| PHP version      | 8.1.18
| Operating system | CloudLinux

ArchBlood:
fixes #6940

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] All tests are passing
- [ ] Changelog was modified

**Other information:**

### DIFF
--- a/protected/humhub/modules/content/widgets/stream/StreamEntryWidget.php
+++ b/protected/humhub/modules/content/widgets/stream/StreamEntryWidget.php
@@ -116,7 +116,7 @@ abstract class StreamEntryWidget extends JsWidget
         }
 
         $params = [
-            'content' => $record instanceof ContentActiveRecord ? $record->getWallOut($options) : '',
+            'content' => $record->getWallOut($options),
             'jsWidget' => $jsWidget,
             'entry' => $record->content,
         ];
--- a/protected/humhub/modules/content/widgets/stream/StreamEntryWidget.php
+++ b/protected/humhub/modules/content/widgets/stream/StreamEntryWidget.php
@@ -102,15 +102,31 @@ abstract class StreamEntryWidget extends JsWidget
             $options = [];
         }
 
+        // Ensure that $record->getWallEntryWidget() returns a valid object
+        $wallEntryWidget = $record->getWallEntryWidget();
+        if (!$wallEntryWidget || !$wallEntryWidget->renderOptions) {
+            // Handle the case where $wallEntryWidget or renderOptions is null
+            return;
+        }
+
+        // Ensure that $wallEntryWidget->renderOptions is properly initialized before accessing its properties
+        $jsWidget = null;
         if (isset($options['jsWidget'])) {
             $jsWidget = $options['jsWidget'];
             unset($options['jsWidget']);
         } else {
-            $jsWidget = $record->getWallEntryWidget()->jsWidget;
+            $jsWidget = $wallEntryWidget->renderOptions->jsWidget;
+        }
+
+        // Check if $record has getWallOut method
+        if (method_exists($record, 'getWallOut')) {
+            $content = $record->getWallOut($options);
+        } else {
+            $content = '';
         }
 
         $params = [
-            'content' => $record->getWallOut($options),
+            'content' => $content,
             'jsWidget' => $jsWidget,
             'entry' => $record->content
         ];
--- a/protected/humhub/modules/content/widgets/stream/StreamEntryWidget.php
+++ b/protected/humhub/modules/content/widgets/stream/StreamEntryWidget.php
@@ -102,33 +102,23 @@ abstract class StreamEntryWidget extends JsWidget
             $options = [];
         }
 
-        // Ensure that $record->getWallEntryWidget() returns a valid object
-        $wallEntryWidget = $record->getWallEntryWidget();
-        if (!$wallEntryWidget || !$wallEntryWidget->renderOptions) {
-            // Handle the case where $wallEntryWidget or renderOptions is null
-            return;
-        }
-
-        // Ensure that $wallEntryWidget->renderOptions is properly initialized before accessing its properties
-        $jsWidget = null;
         if (isset($options['jsWidget'])) {
             $jsWidget = $options['jsWidget'];
             unset($options['jsWidget']);
         } else {
-            $jsWidget = $wallEntryWidget->renderOptions->jsWidget;
+            $wallEntryWidget = $record->getWallEntryWidget();
+            $jsWidget = $wallEntryWidget instanceof StreamEntryWidget ? $wallEntryWidget->jsWidget : null;
         }
 
-        // Check if $record has getWallOut method
-        if (method_exists($record, 'getWallOut')) {
-            $content = $record->getWallOut($options);
-        } else {
-            $content = '';
+        if ($jsWidget === null) {
+            Yii::error('Model ' . get_class($record) . ' must define $wallEntryClass or set $streamChannel to null!', 'content');
+            return '';
         }
 
         $params = [
-            'content' => $content,
+            'content' => $record instanceof ContentActiveRecord ? $record->getWallOut($options) : '',
             'jsWidget' => $jsWidget,
-            'entry' => $record->content
+            'entry' => $record->content,
         ];
 
         return Yii::$app->controller->renderPartial('@humhub/modules/content/views/layouts/wallEntry', $params);


protected/humhub/modules/content/widgets/stream/StreamEntryWidget.php
```diff
@@ -102,15 +102,31 @@ private static function renderLegacyWallEntry(ContentActiveRecord $record, $opti
             $options = [];
         }
 
+        // Ensure that $record->getWallEntryWidget() returns a valid object
+        $wallEntryWidget = $record->getWallEntryWidget();
```
yurabakhtin, line 106: I think the `$wallEntryWidget` should be defined only for case below when `!isset($options['jsWidget'])`.

protected/humhub/modules/content/widgets/stream/StreamEntryWidget.php
```diff
@@ -102,15 +102,31 @@ private static function renderLegacyWallEntry(ContentActiveRecord $record, $opti
...
+        if (!$wallEntryWidget || !$wallEntryWidget->renderOptions) {
+            // Handle the case where $wallEntryWidget or renderOptions is null
+            return;
+        }
+
+        // Ensure that $wallEntryWidget->renderOptions is properly initialized before accessing its properties
+        $jsWidget = null;
```
yurabakhtin, line 113: no need to define the `$jsWidget` here, because below the var is defined anyway.

protected/humhub/modules/content/widgets/stream/StreamEntryWidget.php
```diff
@@ -102,15 +102,31 @@ private static function renderLegacyWallEntry(ContentActiveRecord $record, $opti
...
         if (isset($options['jsWidget'])) {
             $jsWidget = $options['jsWidget'];
             unset($options['jsWidget']);
         } else {
-            $jsWidget = $record->getWallEntryWidget()->jsWidget;
+            $jsWidget = $wallEntryWidget->renderOptions->jsWidget;
```
yurabakhtin, line 118: Why here should we get the property `jsWidget` from `wallEntryWidget->renderOptions` when in previous version we get it directly from the `wallEntryWidget`?
I don't find the class `humhub\modules\content\widgets\stream\StreamEntryOptions` has the property `jsWidget`.
But I see the property `jsWidget` exists for `humhub\modules\content\widgets\stream\StreamEntryWidget` which is used for `wallEntryWidget` usually.

I think we should use here a code like this:

```php
$wallEntryWidget = $record->getWallEntryWidget();
$jsWidget = $wallEntryWidget instanceof StreamEntryWidget ? $wallEntryWidget->jsWidget : null;
```

protected/humhub/modules/content/widgets/stream/StreamEntryWidget.php
```diff
@@ -102,15 +102,31 @@ private static function renderLegacyWallEntry(ContentActiveRecord $record, $opti
...
+        // Check if $record has getWallOut method
+        if (method_exists($record, 'getWallOut')) {
+            $content = $record->getWallOut($options);
+        } else {
+            $content = '';
         }
 
         $params = [
-            'content' => $record->getWallOut($options),
+            'content' => $content,
```
yurabakhtin, line 129: Instead of the `method_exists()` we prefer to use a code like:

```php
'content' => $record instanceof ContentActiveRecord ? $record->getWallOut($options) : ''
```

yurabakhtin:
@ArchBlood Please check my reviews, but before apply the changes do you know what model has an old `wallEntryClass` so it calls the deprecated method `StreamEntryWidget::renderLegacyWallEntry()` and some other inside?
It would be better to find such old model and update it to new version instead of a modifying of the deprecated methods:

- `StreamEntryWidget::renderLegacyWallEntry()`
- `ContentActiveRecord ::getWallEntryWidget()`
- `ContentActiveRecord::getWallOut()`

yurabakhtin (changes requested):

protected/humhub/modules/content/widgets/stream/StreamEntryWidget.php
ArchBlood: I can confirm this is a better option!

protected/humhub/modules/content/widgets/stream/StreamEntryWidget.php
ArchBlood: Moving `$wallEntryWidget = $record->getWallEntryWidget();` here does nothing for the error reported in #6940 but I do see a possible need in some cases, but if you ask me it seems more of dead code here. 🤔

ArchBlood:
To answer the main question about which model, it seems it was due to my Ticket model using `implements \humhub\modules\content\interfaces\Searchable` and using the following still;

```php
    public function getSearchAttributes()
    {
        // Define the searchable attributes
        $attributes = [
            'id' => $this->id,
            'title' => $this->title,
            'description' => $this->description,
        ];

        return $attributes;
    }
```

yurabakhtin:
> ```
> getSearchAttributes
> ```

@ArchBlood As I understand you have a model like this:

```php
class Ticket extends ContentActiveRecord implements Searchable
{
    ...
    public function getSearchAttributes()
    {
        // Define the searchable attributes
        $attributes = [
            'id' => $this->id,
            'title' => $this->title,
            'description' => $this->description,
        ];

        return $attributes;
    }
    ...
}
```

But what does the class have for the properties? :

```php

    /**
     * @see StreamEntryWidget
     * @var string the StreamEntryWidget widget class
     */
    public $wallEntryClass;

    /**
     * The stream channel where this content should displayed.
     * Set to null when this content should not appear on streams.
     *
     * @since 1.2
     * @var string|null the stream channel
     */
    protected $streamChannel = 'default';
```

protected/humhub/modules/content/widgets/stream/StreamEntryWidget.php
yurabakhtin: I mean your code now is:

```php
        $wallEntryWidget = $record->getWallEntryWidget();
        ...
        if (isset($options['jsWidget'])) {
            $jsWidget = $options['jsWidget'];
            unset($options['jsWidget']);
        } else {
            $jsWidget = $wallEntryWidget->renderOptions->jsWidget;
        }
```

So why should we define the `$wallEntryWidget` even for case `if (isset($options['jsWidget'])) {` when the var is not used.
Why don't move it right after the line `} else {`?

protected/humhub/modules/content/widgets/stream/StreamEntryWidget.php
yurabakhtin: Main changing what I have suggested above is not the moving `$wallEntryWidget = $record->getWallEntryWidget();` but the code:

```php
$wallEntryWidget instanceof StreamEntryWidget ? $wallEntryWidget->jsWidget : null;
```

i.e. if the `$wallEntryWidget` instance of `StreamEntryWidget` then we can call `$wallEntryWidget->jsWidget` because the class `StreamEntryWidget` has it defined as `public $jsWidget = 'stream.StreamEntry';` by default.

And I cannot find a property with name `jsWidget` for the `$wallEntryWidget->renderOptions` which is instance of `humhub\modules\content\widgets\stream\StreamEntryOptions`.
It is why it looks starnge for me the code `$wallEntryWidget->renderOptions->jsWidget;`, I guess error may occurs here, or your specific model `Ticket` has such wall entry class with render options which has the defined property `jsWidget`, but in such case better update your model instead of touch the core code.

ArchBlood:
> @ArchBlood As I understand you have a model like this:
> 
> ```
> class Ticket extends ContentActiveRecord implements Searchable
> {
>     ...
>     public function getSearchAttributes()
>     {
>         // Define the searchable attributes
>         $attributes = [
>             'id' => $this->id,
>             'title' => $this->title,
>             'description' => $this->description,
>         ];
> 
>         return $attributes;
>     }
>     ...
> }
> ```
> 
> But what does the class have for the properties? :
> 
> ```
>     /**
>      * @see StreamEntryWidget
>      * @var string the StreamEntryWidget widget class
>      */
>     public $wallEntryClass;
> 
>     /**
>      * The stream channel where this content should displayed.
>      * Set to null when this content should not appear on streams.
>      *
>      * @since 1.2
>      * @var string|null the stream channel
>      */
>     protected $streamChannel = 'default';
> ```

Instead of explaining it I've sent you an invite for better visibility of the code.

protected/humhub/modules/content/widgets/stream/StreamEntryWidget.php
ArchBlood: > I mean your code now is:
> 
> ```
>         $wallEntryWidget = $record->getWallEntryWidget();
>         ...
>         if (isset($options['jsWidget'])) {
>             $jsWidget = $options['jsWidget'];
>             unset($options['jsWidget']);
>         } else {
>             $jsWidget = $wallEntryWidget->renderOptions->jsWidget;
>         }
> ```
> 
> So why should we define the `$wallEntryWidget` even for case `if (isset($options['jsWidget'])) {` when the var is not used. Why don't move it right after the line `} else {`?

This way, `$wallEntryWidget` is always defined, and you won't encounter the "Undefined variable" error when using the search for content.

ArchBlood: If it's made conditional the error will still happen, I've tested this and the error pops up when using the "Show all results" for content.

yurabakhtin:
@ArchBlood 

> Instead of explaining it I've sent you an invite for better visibility of the code.

Thank you for the access, what I have investigated:

- You have the class as searchable `class Ticket extends ContentActiveRecord implements Searchable`,
- and search attibutes are defined in the method `getSearchAttributes()`, it is ok,
- but it strange why the class has no `$wallEntryClass`,
- if I test a model `SomeRecord extends ContentActiveRecord implements Searchable` without the property `$wallEntryClass` on previous version 1.15 I get such picture:

no-wallEntryClass

I.e. the record is found but it is not rendered, yes we don't have any errors there, but it breaks at least the searching counter.

I advise you either implement the `$wallEntryClass` for the `Ticket` or add `protected $streamChannel = null;` so it excludes the ticket records from wall stream and it will not be visible on the searching page.

But please note the applying `$streamChannel = null` will works only for new created records because a value is stored in the database column `content:stream_channel`, so I don't recommend this way.
You can do this only if you really don't need to see the  tickets on a wall stream, in such case for proper work you should update the column `content:stream_channel` to `null` by migration or manually if it is only on your own DB.

---

To avoid errors I suggest to update the method `StreamEntryWidget::renderLegacyWallEntry` like this:

```php
private static function renderLegacyWallEntry(ContentActiveRecord $record, $options = [])
{
    if (!is_array($options)) {
        $options = [];
    }

    if (isset($options['jsWidget'])) {
        $jsWidget = $options['jsWidget'];
        unset($options['jsWidget']);
    } else {
        $wallEntryWidget = $record->getWallEntryWidget();
        $jsWidget = $wallEntryWidget instanceof StreamEntryWidget ? $wallEntryWidget->jsWidget : null;
    }

    if ($jsWidget === null) {
        Yii::error('Model ' . get_class($record) . ' must define $wallEntryClass or set $streamChannel to null!', 'content');
        return '';
    }

    $params = [
        'content' => $record instanceof ContentActiveRecord ? $record->getWallOut($options) : '',
        'jsWidget' => $jsWidget,
        'entry' => $record->content,
    ];

    return Yii::$app->controller->renderPartial('@humhub/modules/content/views/layouts/wallEntry', $params);
}
```

Could you please test how it works on your DB with already created Tickets?
If all ok I prefer to use this version for your current PR.
Thanks.

ArchBlood:
> @ArchBlood
> 
> > Instead of explaining it I've sent you an invite for better visibility of the code.
> 
> Thank you for the access, what I have investigated:
> 
> * You have the class as searchable `class Ticket extends ContentActiveRecord implements Searchable`,
> * and search attibutes are defined in the method `getSearchAttributes()`, it is ok,
> * but it strange why the class has no `$wallEntryClass`,
> * if I test a model `SomeRecord extends ContentActiveRecord implements Searchable` without the property `$wallEntryClass` on previous version 1.15 I get such picture:
> 
> no-wallEntryClass
> 
> I.e. the record is found but it is not rendered, yes we don't have any errors there, but it breaks at least the searching counter.
> 
> I advise you either implement the `$wallEntryClass` for the `Ticket` or add `protected $streamChannel = null;` so it excludes the ticket records from wall stream and it will not be visible on the searching page.
> 
> But please note the applying `$streamChannel = null` will works only for new created records because a value is stored in the database column `content:stream_channel`, so I don't recommend this way. You can do this only if you really don't need to see the tickets on a wall stream, in such case for proper work you should update the column `content:stream_channel` to `null` by migration or manually if it is only on your own DB.
> 
> To avoid errors I suggest to update the method `StreamEntryWidget::renderLegacyWallEntry` like this:
> 
> ```
> private static function renderLegacyWallEntry(ContentActiveRecord $record, $options = [])
> {
>     if (!is_array($options)) {
>         $options = [];
>     }
> 
>     if (isset($options['jsWidget'])) {
>         $jsWidget = $options['jsWidget'];
>         unset($options['jsWidget']);
>     } else {
>         $wallEntryWidget = $record->getWallEntryWidget();
>         $jsWidget = $wallEntryWidget instanceof StreamEntryWidget ? $wallEntryWidget->jsWidget : null;
>     }
> 
>     if ($jsWidget === null) {
>         Yii::error('Model ' . get_class($record) . ' must define $wallEntryClass or set $streamChannel to null!', 'content');
>         return '';
>     }
> 
>     $params = [
>         'content' => $record instanceof ContentActiveRecord ? $record->getWallOut($options) : '',
>         'jsWidget' => $jsWidget,
>         'entry' => $record->content,
>     ];
> 
>     return Yii::$app->controller->renderPartial('@humhub/modules/content/views/layouts/wallEntry', $params);
> }
> ```
> 
> Could you please test how it works on your DB with already created Tickets? If all ok I prefer to use this version for your current PR. Thanks.

I've tested your version with both `public $wallEntryClass` & `protected $streamChannel = null;` , where both throw an error as your code intends to do even if they are defined in Ticket model, not really what I was trying to achieve here as an error is still an error if not handled. I've also tried using the following modified version of your snippet with the same issue, so in turn this does not work;

```php
private static function renderLegacyWallEntry(ContentActiveRecord $record, $options = [])
{
    $jsWidget = $options['jsWidget'] ?? $record->getWallEntryWidget()->jsWidget ?? null;

    if ($jsWidget === null) {
        Yii::error('Model ' . get_class($record) . ' must define $wallEntryClass or set $streamChannel to null!', 'content');
        return '';
    }

    $content = $record instanceof ContentActiveRecord ? $record->getWallOut($options) : '';

    $params = [
        'content' => $content,
        'jsWidget' => $jsWidget,
        'entry' => $record->content,
    ];

    return Yii::$app->controller->renderPartial('@humhub/modules/content/views/layouts/wallEntry', $params);
}
```

yurabakhtin:
> ```php
> private static function renderLegacyWallEntry(ContentActiveRecord $record, $options = [])
> {
>     $jsWidget = $options['jsWidget'] ?? $record->getWallEntryWidget()->jsWidget ?? null;
> 
>     if ($jsWidget === null) {
>         Yii::error('Model ' . get_class($record) . ' must define $wallEntryClass or set $streamChannel to null!', 'content');
>         return '';
>     }
> 
>     $content = $record instanceof ContentActiveRecord ? $record->getWallOut($options) : '';
> 
>     $params = [
>         'content' => $content,
>         'jsWidget' => $jsWidget,
>         'entry' => $record->content,
>     ];
> 
>     return Yii::$app->controller->renderPartial('@humhub/modules/content/views/layouts/wallEntry', $params);
> }
> ```

I guess you have an error when you try to call `$record->getWallEntryWidget()->jsWidget` but `$record->getWallEntryWidget() === null`.

In my code I did this:

```php
$wallEntryWidget = $record->getWallEntryWidget();
$jsWidget = $wallEntryWidget instanceof StreamEntryWidget ? $wallEntryWidget->jsWidget : null;
```

Could you show me please what error did you get when you use my code?

---

> I've tested your version with both `public $wallEntryClass`

Did you add it like `public $wallEntryClass = WallEntry::class;`?
Did you create new class `WallEntry` for the `Ticket`, as it is implemented for other modules, for example: https://github.com/humhub/news/blob/main/widgets/WallEntry.php
Also you should create a view file like this https://github.com/humhub/news/blob/main/widgets/views/entry.php.

yurabakhtin:
@ArchBlood I have tested by installing the FAQ module and create a ticket:

ticket

Then I try to search it:

search

Probably we should fix the core class `humhub\modules\content\search\SearchRequest` to avoid such error for case when some model like Ticket returns `null` by the method `ContentActiveRecord::getContentName()`.
Please note usually such methods returns some static text, so I suggest to update the method of the `Ticket` to:

```php
public function getContentName()
{
    return $this->title ?? Yii::t('FaqModule.base', 'Ticket');
}
```

because it seems the core `SearchRequest` uses a code like `(new Ticket())->getContentName()`, so the property `Ticket->title` is `null` there.

---

ok, if I update the method as I suggested above I catch the error `Attempt to read property "jsWidget" on null`:

search_error

Then I update the method `StreamEntryWidget::renderLegacyWallEntry()` to my suggested code:

```php
private static function renderLegacyWallEntry(ContentActiveRecord $record, $options = [])
{
    if (!is_array($options)) {
        $options = [];
    }

    if (isset($options['jsWidget'])) {
        $jsWidget = $options['jsWidget'];
        unset($options['jsWidget']);
    } else {
        $wallEntryWidget = $record->getWallEntryWidget();
        $jsWidget = $wallEntryWidget instanceof StreamEntryWidget ? $wallEntryWidget->jsWidget : null;
    }

    if ($jsWidget === null) {
        Yii::error('Model ' . get_class($record) . ' must define $wallEntryClass or set $streamChannel to null!', 'content');
        return '';
    }

    $params = [
        'content' => $record instanceof ContentActiveRecord ? $record->getWallOut($options) : '',
        'jsWidget' => $jsWidget,
        'entry' => $record->content,
    ];

    return Yii::$app->controller->renderPartial('@humhub/modules/content/views/layouts/wallEntry', $params);
}
```

and I can run the search page without errors, only the Ticket record is not visible because WallEntry is not implemented for it yet.

search_no-errors

---

If you need to display the Ticket on the search list you should update the Ticket class by adding the line:

```php
public $wallEntryClass = WallEntry::class;
```

Then you should add the class WallEntry with code like this:

```php
<?php
namespace gm\humhub\modules\faq\widgets;

use gm\humhub\modules\faq\models\Ticket;
use humhub\modules\content\widgets\stream\WallStreamEntryOptions;
use humhub\modules\content\widgets\stream\WallStreamEntryWidget;
use humhub\modules\content\widgets\stream\WallStreamModuleEntryWidget;

class WallEntry extends WallStreamModuleEntryWidget
{
    /**
     * @inheritdoc
     */
    public $createRoute = '/faq/ticket/create';

    /**
     * @inheritdoc
     * @var Ticket
     */
    public $model;

    /**
     * @inheritdoc
     */
    public $editRoute = '/faq/ticket/update';

    /**
     * @inheritdoc
     */
    public $editMode = WallStreamEntryWidget::EDIT_MODE_NEW_WINDOW;

    /**
     * @inheritdoc
     */
    public function renderContent()
    {
        return $this->render('wall-entry', [
            'ticket' => $this->model,
            'isDetailView' => $this->renderOptions->isViewContext(WallStreamEntryOptions::VIEW_CONTEXT_DETAIL),
        ]);
    }

    /**
     * @inheritdoc
     */
    protected function getTitle()
    {
        return $this->model->title;
    }
}
```

and also a view file `/widgets/views/wall-entry.php`:

```php
<?php
use gm\humhub\modules\faq\models\Ticket;
use humhub\modules\content\widgets\richtext\RichText;

/* @var $ticket Ticket */
/* @var $isDetailView bool */
?>

<div data-news="<?= $ticket->id ?>" data-content-key="<?= $ticket->content->id ?>">
    <div data-ui-markdown<?= $isDetailView ? '' : ' data-ui-show-more' ?>>
        <?= RichText::output($ticket->description) ?>
    </div>
</div>
```

Then a Tciket record will be visbile on the search page like this:

view-ticket

I see the wall entry of the Ticket(as you can see on my screenshot above) even without the modifications in the core method, but it would be better to keep such fixes for some unexpected errors in the broken model like the Ticket, so I hope you update this PR to that my suggested code and I will approve this PR, otherwise please show me why it doesn't work for you, thanks.

---

But if you don't need to search ticket records then you can just remove `implements Searchable` and the method `public function getSearchAttributes()`.

Please let me know if have some questions else, thanks.

ArchBlood:
> But if you don't need to search ticket records then you can just remove implements Searchable and the method public function getSearchAttributes().
> 
> Please let me know if have some questions else, thanks.

The reason why the ticket was made searchable is so that it was easier for the user to find, of course we didn't want it to return the actual ticket information through a `WallEntry::class` due to privacy concerns;

Going back to using `protected $streamChannel = null;` this does the same as using `public $wallEntryClass = WallEntry::class;`

What @GreenMeteor wants to do is make use of the search function without the use of displaying the actual information in the search stream which both of the above does with the added error applied.
Screenshot_1
Screenshot_2

This P/R fixes the original reported error and makes it possible to use the search function how we needed it without using either `$wallEntryClass` or `$streamChannel`

ArchBlood:
I'll have to go with your suggestion and implement some restrictions @yurabakhtin so that only the creator can view their own Tickets.

yurabakhtin:
@ArchBlood I see you updated this PR as I recommended, thanks, I can approve it now, so it works like this without changes in the `Ticket` class:

no-wallEntryClass

It may be a confused in the counter part, because Ticket records are not visible.

But as I understand you don't want to implement `$wallEntryClass` fo the `Ticket` class because it will be visible on the Dashboard:

visible-on-dashboard

@luke- Currently we index content records only with `$streamChannel = 'default'`, the code is [here](https://github.com/humhub/humhub/blob/develop/protected/humhub/modules/content/services/ContentSearchService.php#L80): `$this->content->stream_channel === Content::STREAM_CHANNEL_DEFAULT`.

What do you think if we will introduce new `stream_channel = 'search'` and index such content records as well?

So then @ArchBlood could make the Tickets are visbile only on the search list and hidden on Dashboard stream:

```php
class Ticket extends ContentActiveRecord implements Searchable
{
    public $wallEntryClass = WallEntry::class;

    public $streamChannel = Content::STREAM_CHANNEL_SEARCH; // 'search'
```

yurabakhtin (approved):

luke-:
@yurabakhtin If the content should not appear on the dashboard, can't we just use the new Visbility flag?

Otherwise, we also have modules that simply implement a separate stream channel in such cases: https://github.com/humhub/translation/blob/master/models/TranslationLog.php#L41 

Do we need to change anything else in the PR in this regard?

protected/humhub/modules/content/widgets/stream/StreamEntryWidget.php
```diff
@@ -106,13 +106,19 @@ private static function renderLegacyWallEntry(ContentActiveRecord $record, $opti
...
+        }
+
+        if ($jsWidget === null) {
+            Yii::error('Model ' . get_class($record) . ' must define $wallEntryClass or set $streamChannel to null!', 'content');
+            return '';
         }
 
         $params = [
-            'content' => $record->getWallOut($options),
+            'content' => $record instanceof ContentActiveRecord ? $record->getWallOut($options) : '',
```
luke-, line 119: Should we really allow non `ContentActiveRecord`  in the stream?

protected/humhub/modules/content/widgets/stream/StreamEntryWidget.php
```diff
@@ -106,13 +106,19 @@ private static function renderLegacyWallEntry(ContentActiveRecord $record, $opti
             $jsWidget = $options['jsWidget'];
             unset($options['jsWidget']);
         } else {
-            $jsWidget = $record->getWallEntryWidget()->jsWidget;
+            $wallEntryWidget = $record->getWallEntryWidget();
+            $jsWidget = $wallEntryWidget instanceof StreamEntryWidget ? $wallEntryWidget->jsWidget : null;
```
luke-, line 110: Should we really allow non `StreamEntryWidget`  in the stream?

luke- (changes requested):

protected/humhub/modules/content/widgets/stream/StreamEntryWidget.php
ArchBlood: Personally speaking, I would have liked my original idea which would have made this so, but @yurabakhtin's suggestions were definitely a better option in regards to the error reported in #6940, the only issue I had with it was that it requires more coding on my end.

protected/humhub/modules/content/widgets/stream/StreamEntryWidget.php
ArchBlood: Same as https://github.com/humhub/humhub/pull/6941#discussion_r1566484418

protected/humhub/modules/content/widgets/stream/StreamEntryWidget.php
yurabakhtin: @luke- I saw in the first commit [here](https://github.com/humhub/humhub/pull/6941/commits/f97d73f609f5967c95d0d32ca3a03637b0ee7d6a#diff-708348c91e9bdcfac2cf80d59dbff60f3b13e95b8da5bdeb13cedc3783eec5bdR122) @ArchBlood wrote a code like `method_exists($record, 'getWallOut')`, so I suggested to replace it with `$record instanceof ContentActiveRecord`.

Yes, you are right, now we index only `ContentActiveRecord`, because only `Content` records are indexed [here](https://github.com/humhub/humhub/blob/develop/protected/humhub/modules/content/commands/SearchController.php#L42-L45):

```php
foreach (Content::find()->each() as $content) {
    (new ContentSearchService($content))->update(false);
}
```

So maybe we could revert the line back to `$record->getWallOut($options)`, because a main issue was only in a code above `$record->getWallEntryWidget()->jsWidget`.

---

As I understand all `ContentActiveRecord`s become searchable automatically even without `Searchable` interface since `v1.16` because you have added the method `getSearchAttributes` [here](https://github.com/humhub/humhub/commit/92ef2926d64e2cc50d07a0577ff4ca5f53ac88b3#diff-5fea8ec3ab977d4e8e1db5cf83b2509d7cdccff1e612ca16efb88c2a37e18793) for the class `ContentActiveRecord`.
So we can/should remove a code `implements Searchable` from all `ContentActiveRecord`s for versions 1.16+.

Then I think we should test how the searching works for `ContentActiveRecord` models without `implements Searchable` and with `$streamChannel = default` and without `$wallEntryClass = WallEntry::class`.
I find only one such this: [rss-importer/models/RssFeed](https://github.com/humhub/rss-importer/blob/main/models/RssFeed.php#L29).

protected/humhub/modules/content/widgets/stream/StreamEntryWidget.php
yurabakhtin: @luke- If we should not allow such records in a search stream then we should exclude them from indexing [here](https://github.com/humhub/humhub/blob/develop/protected/humhub/modules/content/services/ContentSearchService.php#L80), otherwise we may have such issues from external modules like the current one.

protected/humhub/modules/content/widgets/stream/StreamEntryWidget.php
yurabakhtin: @luke- I have fixed RSS Feed in the PR https://github.com/humhub/rss-importer/pull/19.

protected/humhub/modules/content/widgets/stream/StreamEntryWidget.php
ArchBlood: > @luke- If we should not allow such records in a search stream then we should exclude them from indexing [here](https://github.com/humhub/humhub/blob/develop/protected/humhub/modules/content/services/ContentSearchService.php#L80), otherwise we may have such issues from external modules like the current one.

I agree here, maybe we could include a `$exclude = [];` for such cases?

protected/humhub/modules/content/widgets/stream/StreamEntryWidget.php
ArchBlood: > @luke- I saw in the first commit [here](https://github.com/humhub/humhub/pull/6941/commits/f97d73f609f5967c95d0d32ca3a03637b0ee7d6a#diff-708348c91e9bdcfac2cf80d59dbff60f3b13e95b8da5bdeb13cedc3783eec5bdR122) @ArchBlood wrote a code like `method_exists($record, 'getWallOut')`, so I suggested to replace it with `$record instanceof ContentActiveRecord`.
> 
> Yes, you are right, now we index only `ContentActiveRecord`, because only `Content` records are indexed [here](https://github.com/humhub/humhub/blob/develop/protected/humhub/modules/content/commands/SearchController.php#L42-L45):
> 
> ```
> foreach (Content::find()->each() as $content) {
>     (new ContentSearchService($content))->update(false);
> }
> ```
> 
> So maybe we could revert the line back to `$record->getWallOut($options)`, because a main issue was only in a code above `$record->getWallEntryWidget()->jsWidget`.
> 
> As I understand all `ContentActiveRecord`s become searchable automatically even without `Searchable` interface since `v1.16` because you have added the method `getSearchAttributes` [here](https://github.com/humhub/humhub/commit/92ef2926d64e2cc50d07a0577ff4ca5f53ac88b3#diff-5fea8ec3ab977d4e8e1db5cf83b2509d7cdccff1e612ca16efb88c2a37e18793) for the class `ContentActiveRecord`. So we can/should remove a code `implements Searchable` from all `ContentActiveRecord`s for versions 1.16+.
> 
> Then I think we should test how the searching works for `ContentActiveRecord` models without `implements Searchable` and with `$streamChannel = default` and without `$wallEntryClass = WallEntry::class`. I find only one such this: [rss-importer/models/RssFeed](https://github.com/humhub/rss-importer/blob/main/models/RssFeed.php#L29).

With the current idea it will throw `must define $wallEntryClass or set $streamChannel to null!` which isn't desirable in the long run. I've tested it multiple times and have set `protected $streamChannel = null;` so the logic is off here, I can agree that `Searchable` isn't needed anymore as I've tested without it and it works as intended, but the issue still remains the same when it comes to the new changes as suggested, you **_MUST_** set `$wallEntryClass` this is not avoidable, so in this case setting `$streamChannel` does nothing at all and is useless.

protected/humhub/modules/content/widgets/stream/StreamEntryWidget.php
luke-: @yurabakhtin Can't we just additionally check here whether the Content AR provides a Wall Entry class?

If this is null, we can skip the indexing.  

https://github.com/humhub/humhub/blob/develop/protected/humhub/modules/content/services/ContentSearchService.php#L80

yurabakhtin: @luke- Yes, it is possible to check there by code like `empty($this->content->getModel()->wallEntryClass)`, but this code will does an additional database query to get `ContentActiveRecord`, it may does the indexing process slower.

I think better solution is to fix all wrong ContentActiveRecords which have `$streamChannel = 'default'` and have no `$wallEntryClass`.

yurabakhtin: @luke- Then you can merge this PR, it will informs about such wrong ContentActiveRecords with new log error `Yii::error('Model ' . get_class($record) . ' must define $wallEntryClass or set $streamChannel to null!', 'content');`.

luke-:
@yurabakhtin Why do we need the new checks for instanceof `StreamEntryWidget`  and `ContentActiveRecord`?

yurabakhtin:
> Why do we need the new checks for instanceof `StreamEntryWidget` and `ContentActiveRecord`?

@luke- I thought we should avoid a displaying of the [error](https://github.com/humhub/humhub/issues/6940), but if you prefer don't touch core code and fix the external model with wrong properties instead, then yes we can skip this PR.

luke-:
@yurabakhtin Ok the error is fine but why we need `'content' => $record instanceof ContentActiveRecord ? $record->getWallOut($options) : '',`?

yurabakhtin:
> @yurabakhtin Ok the error is fine but why we need `'content' => $record instanceof ContentActiveRecord ? $record->getWallOut($options) : '',`?

@luke- I confirm, we don't need this change at all because I don't know how such case may happens, I already answered about this [here](https://github.com/humhub/humhub/pull/6941#discussion_r1566814479), so the change can be reverted.

luke-:
Ok, thanks for the explanation:

From my side, we can either:

- keep the PR and revert the following line: https://github.com/humhub/humhub/pull/6941/files#diff-708348c91e9bdcfac2cf80d59dbff60f3b13e95b8da5bdeb13cedc3783eec5bdR119 - Then we would have an error message if StreamEntry is not defined

- close the PR and fix the module page if necessary

ArchBlood:
> Ok, thanks for the explanation:
> 
> From my side, we can either:
> 
> * keep the PR and revert the following line: https://github.com/humhub/humhub/pull/6941/files#diff-708348c91e9bdcfac2cf80d59dbff60f3b13e95b8da5bdeb13cedc3783eec5bdR119 - Then we would have an error message if StreamEntry is not defined
> * close the PR and fix the module page if necessary

I've reverted it back the line mentioned incase of this P/R being accepted.

luke- (approved):

PR #6957, opened 2024-04-19
PL translation update

nhlpl:
EN to PL


ArchBlood:
Such translations should be done via the [Translation community](https://translate.humhub.org/dashboard).

PR #6971, opened 2024-04-29
Meta search: Highlight keyword in single content view

yurabakhtin:

- [ ] Feature

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
https://github.com/humhub/humhub-internal/issues/158

### DIFF
--- a/protected/humhub/modules/content/Events.php
+++ b/protected/humhub/modules/content/Events.php
@@ -11,15 +11,14 @@ namespace humhub\modules\content;
 use humhub\commands\CronController;
 use humhub\commands\IntegrityController;
 use humhub\components\Event;
+use humhub\modules\content\assets\ContentHighlightAsset;
 use humhub\modules\content\components\ContentActiveRecord;
 use humhub\modules\content\models\Content;
 use humhub\modules\content\services\ContentSearchService;
 use humhub\modules\user\events\UserEvent;
 use Yii;
 use yii\base\BaseObject;
-use yii\console\Controller;
 use yii\helpers\Console;
-use yii\web\View;
 
 /**
  * Events provides callbacks to handle events.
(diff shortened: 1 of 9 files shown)


protected/humhub/modules/content/Events.php
```diff
@@ -143,6 +144,27 @@ public static function onCronBeforeAction($event): void
...
+        if ($highlight === null || $highlight === '') {
+            return;
+        }
+        Yii::$app->session->remove('contentHighlight');
+
+        /* @var View $view */
+        $view = $event->sender;
+
+        $index = time();
+        $view->registerJs('function contentHighlight' . $index . '() {
```
luke-, line 159: Couldn't we somehow outsource this JS to e.g. the `search` or `content` javascript?

protected/humhub/modules/content/Events.php
```diff
@@ -143,6 +144,27 @@ public static function onCronBeforeAction($event): void
...
+            "' . str_replace('"', '\"', $highlight) . '".split(" ")
```
luke-, line 160: I think we need better encoding here. I'm a bit afraid that this could lead to an injection into the JS that is not intercepted by the new nonce.


Can you please use something like 

`$this->registerJsVar('contentSearchKeyword', $keyword);` to safely pass the PHP Var to JS.

luke- (changes requested):
@yurabakhtin

protected/humhub/modules/content/Events.php
yurabakhtin: @luke- Updated in the commit https://github.com/humhub/humhub/pull/6971/commits/55381159b7a3ad050ebd6d02aa56a7bff41c7733.

protected/humhub/modules/content/Events.php
yurabakhtin: commit https://github.com/humhub/humhub/pull/6971/commits/55381159b7a3ad050ebd6d02aa56a7bff41c7733

luke- (approved):

Issue #6552, opened 2023-09-11
JS: call `init` function each time a modal box is reopened


marc-farre:
### Is your feature request related to a problem? Please describe.

- Controller: `return $this->renderAjax('my-modal-view');`
- View: `Assets::register($this);`
- JS:

```js
humhub.module('myModule.test', function (module, require, $) {
    module.initOnPjaxLoad = true;
    const init = function () {
        console.log('test');
    };
    module.export({
        init: init
    });
});
```

First time I open the modal box, "test" is displayed in the console.
But the second time, it is not displayed.

### Describe the solution you'd like

Having something similar to `module.initOnPjaxLoad = true;` for view rendered with ajax. E.g.
`module.initOnAjaxLoad = true;`

### Describe alternatives you've considered

`module.initOnPjaxLoad = true;` loads `init` function on after each `renderAjax()`.

yurabakhtin:
@marc-farre I am not sure some solution exists to detect JS function is called from script file which was loaded by AJAX.
But as I have understood your idea you want to call the JS function `module.init()` on each time when a php file is rendered which contains like `<script src="myModule.test.js" />` that run the JS function `humhub.module('myModule.test', ...)`, so if it is true then I would suggest to name new option as `initOnEachLoad`, i.e. module with such option will be initialized on each time when the function `humhub.module(...)` is called.
I hope the PR https://github.com/humhub/humhub/pull/6563 is what you need, please review and test, thanks.

marc-farre:
Thanks very much @yurabakhtin .
This is a great solution which can be used for many use cases, because it solves the problem with the `initOnPjaxLoad` which call the `init()` function on each pages, even the one where the JS is not loaded by PHP.

Your https://github.com/humhub/humhub/pull/6563 works perfectly, but the asset extending `AssetBundle` needs to have force copy on (which is for development environment):
```php
public $publishOptions = [
        'forceCopy' => true,
    ];
```

Would you have an idea to make this solution work with the default value `'forceCopy' => false,`? I've searched a little, it seams not to be easy... Perhaps with one of these events? https://github.com/defunkt/jquery-pjax#events
Thanks!

yurabakhtin:
> but the asset extending `AssetBundle` needs to have force copy on (which is for development environment):

[...]

PR #6979, opened 2024-04-30
Content Search: use AND operator by default and don't apply mask for phrases

yurabakhtin:

- [x] Bugfix
- [x] Feature

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
https://github.com/humhub/humhub-internal/issues/230


protected/humhub/modules/content/search/driver/MysqlDriver.php
```diff
@@ -129,10 +129,10 @@ private function createMysqlFullTextQuery(SearchQuery $query, array $matchFields
         $againstQuery = '';
 
         foreach ($query->andTerms as $keyword) {
-            $againstQuery .= '+' . rtrim($keyword, '*') . '* ';
```
luke-, line 132: @yurabakhtin Is there a reason why you removed the wildcards?

Basically, the search for “Apple” should also find “Applepie”.

luke- (changes requested):

protected/humhub/modules/content/search/driver/MysqlDriver.php
yurabakhtin: @luke- I have moved the wildcards from drivers(mysql, zend, solr) to `SearchQuery` code [here](https://github.com/humhub/humhub/pull/6979/files#diff-70f82fb3d09d4959fb9f44a526f8455144c112df88ec7299a2efe28a76171169R63-R66), because the chars `"` are removed on the `SearchQuery` side, so we cannot know what original term was really when we are on the driver side.

This is how it worked before the change:

 - `Apple` => `Apple*` (find `Applepie`)
 - `"Apple"` => `Apple*` (find `Applepie`)

After the change:

 - `Apple` => `Apple*` (find `Applepie`)
 - `"Apple"` => `Apple` (find only `Apple`, no `Applepie`)

As I understand it should work as it is described in the first table [here](https://github.com/humhub/humhub-internal/issues/159).

Tests for quoted cases are here:

- https://github.com/humhub/humhub/pull/6979/files#diff-097b244c3690843dabcdabfa0289d8a360bb013a8e05289e2e27d68301547962R51-R52
- https://github.com/humhub/humhub/pull/6979/files#diff-6d94b066e5467f549cd0cc6b834e574e7f82da4d61d4581aeaec7da9b3f9d572R19-R22
- https://github.com/humhub/humhub/pull/6979/files#diff-6d94b066e5467f549cd0cc6b834e574e7f82da4d61d4581aeaec7da9b3f9d572R108

I have done the change for the request "Phrases should actually work. Can you check what is causing the problem here and add tests if necessary?", please correct me if I misunderstood it.

luke-: @yurabakhtin My intention was for “SearchQuery” to be as neutral a representation of the search query as possible. Without any syntax like wildcards or terms in quotation marks.

e.g. Asterisk or Quotation marks  is more of a driver-specific syntax. (Even if the similar for MySQL and Lucene)

But I understand that, for example, with `SearchRequest::andTerms` we cannot distinguish whether it was originally quoted or not.

 Do you have any ideas?

- We could introduce new attributes such as `andParts` (with Wildcard) and `andTerms` (with Quotationmarks).

- We could define that as soon as a space is contained in e.g. `andTerms`, it is seen as a term, otherwise it is provided with a wildcard. (But then `“Apple”` always becomes `Apple*`. 

- ?

Or did I get something wrong here?

yurabakhtin: @luke- 

> My intention was for “SearchQuery” to be as neutral a representation of the search query as possible. Without any syntax like wildcards or terms in quotation marks.
>
> e.g. Asterisk or Quotation marks is more of a driver-specific syntax. (Even if the similar for MySQL and Lucene)

Ok, I have understood your intention.

For solve it we could implement new array as `andParts` or extend the existing array with new array item to mark each term like this:

```php
SearchQuery->andTerms = [
    ['Apple', 'mask' => false],
    ['Pine', 'mask' => true],
];
```

so on driver side we will know what term should be masked with `*`.

But I am thinking why should we keep an additional array `andParts` or use the complex structure if we can use a last char of each term as flag. I.e. if the last char is `*` then we can decide such term should be processed with specific way on a driver side.
For current drivers(mysql, zend, solr) we can keep the code as is, but for future drivers the last char `*` can be used as flag like this, if the driver doesn't support terms ended with `*`:

```php
foreach ($request->getSearchQuery()->andTerms as $term) {
    $keywordQuery->addSubquery(str_ends_with($term, '*') ? $this->prepareMaskedTerm($term) : $term);
}
```

luke-: @yurabakhtin Okay, I got your point.

Do I understand it correctly that now e.g. in `SearchQuery::andTerms`, can contain entries that are either quoted and contain an asterisk.  Or is only the asterisk included?

Your approach is probably the right one. Since in fact all current drivers support this and it also corresponds to the query language of `SearchQuery`.

yurabakhtin: > Do I understand it correctly that now e.g. in `SearchQuery::andTerms`, can contain entries that are either quoted and contain an asterisk. Or is only the asterisk included?

@luke- Yes, but a bit correction about "quoted":

- entries that are either **simple word without quotes** or contain an asterisk

Examples how a requested keyword is stored in terms:

- `Apple` => `Apple*`
- `"Apple"` => `Apple`
- `Foo bar` => `Foo bar*`
- `"Foo bar"` => `Foo bar`

marc-farre:
@yurabakhtin Very sorry, comment https://github.com/humhub/humhub/issues/6552#issuecomment-2092784285 and the followings are in the wrong issue...

However, thanks for the fix with the `@` char in the search field with your commit https://github.com/humhub/humhub/commit/308e40014a2c7014b9a21b92b2c15f3205fa7b6e

FYI, there is still an error when typing the `<` and `>`:

image

yurabakhtin:
> However, thanks for the fix with the `@` char in the search field with your commit [308e400](https://github.com/humhub/humhub/commit/308e40014a2c7014b9a21b92b2c15f3205fa7b6e)
> 
> FYI, there is still an error when typing the `<` and `>`:

@marc-farre Thanks, fixed in the commit https://github.com/humhub/humhub/pull/6979/commits/7aed7d6d9461ddb45f0f4780757e2ecdc6781ffc.

I have also tested all special chars from my keyboard and found the same problem with chars `~%()$` and also if char `-` or `+` is typed several time at the keyword beginning like `----keyword` or `+++keyword`.
Let me know if you will found others, thanks.

marc-farre:
@yurabakhtin Thanks.
@luke- On my side, everything's fine now. The meta-search engine is really powerful now!!

protected/humhub/modules/content/search/driver/MysqlDriver.php
luke-: @yurabakhtin Can you please add this example somewhere in the `SearchQuery` PHPDoc?

luke- (changes requested):
@yurabakhtin Thanks looks good. Can you please check my two comments? 

Are customization tests also necessary for the Solr class?

protected/humhub/modules/content/search/driver/MysqlDriver.php
```diff
@@ -145,6 +158,11 @@ private function createMysqlFullTextQuery(SearchQuery $query, array $matchFields
         );
     }
 
+    protected function prepareKeyword(string $keyword): string
```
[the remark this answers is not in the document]
yurabakhtin: @luke- Yes, correct, I have added a comment [here](https://github.com/humhub/humhub/pull/6979/commits/9cec0c6165cdc77a83a9336b8246f35a9a205dea#diff-15fad754585223b2ec6903d7e6c0bf67fdfa7e21721bc5bf64adde8713657418R163).

protected/humhub/modules/content/search/driver/MysqlDriver.php
yurabakhtin: > @yurabakhtin Can you please add this example somewhere in the `SearchQuery` PHPDoc?

@luke- https://github.com/humhub/humhub/pull/6979/commits/9cec0c6165cdc77a83a9336b8246f35a9a205dea#diff-70f82fb3d09d4959fb9f44a526f8455144c112df88ec7299a2efe28a76171169R28-R64

luke- (approved):

PR #6997, opened 2024-05-10
Fix migration on added duplicated container settings

yurabakhtin:

- [x] Bugfix

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**
https://github.com/humhub/humhub-internal/issues/246

### DIFF
--- a/protected/humhub/modules/notification/migrations/m240422_162959_new_is_untouched_settings.php
+++ b/protected/humhub/modules/notification/migrations/m240422_162959_new_is_untouched_settings.php
@@ -21,16 +21,23 @@ class m240422_162959_new_is_untouched_settings extends Migration
     {
         $rows = (new Query())
             ->select([
-                "module_id",
-                "contentcontainer_id",
-                new Expression("'is_touched_settings' as name"),
-                new Expression("'1' as value"),
+                'cs1.module_id',
+                'cs1.contentcontainer_id',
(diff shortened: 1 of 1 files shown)


luke-:
@marc-farre FYI

marc-farre:
Interesting, I wouldn't have thought about this eventuality, thanks for mentioning me.

@yurabakhtin Thanks for the fix!
Does it mean we should always check if the database is not already updated before applying a migration? E.g., if we add a column, we should check if it doesn't already exist?

yurabakhtin:
> @yurabakhtin Thanks for the fix!
> Does it mean we should always check if the database is not already updated before applying a migration? E.g., if we add a column, we should check if it doesn't already exist?

@marc-farre Your change is not related to a creating of new column, there was an issue related to duplicated records with unique index, really it shouldn't have happened for a normal migration call, it seems it happened because 2 migration processes were called twice ata the same time, that should be fixed from core side, I will review the code soon.

However for adding new column we should use `humhub\components\Migration` and methods like `safeAddColumn()`, `safeCreateTable()` and etc. which already check it for existing before a creating, instead of `yii\db\Migration` and `addColumn()`.

marc-farre:
@yurabakhtin Thanks. 

@luke- I've added @yurabakhtin recommendations in the doc via the PR https://github.com/humhub/documentation/pull/122

Issue #6962, opened 2024-04-22
Directory Filter - Remove Followers from Status if Following is disabled in Module


luke-:
![image](https://github.com/humhub/humhub/assets/4736168/68b16c70-f048-4927-8118-6d6bc448e631)

ArchBlood:
@yurabakhtin does this also fix the current issue when it comes to User Groups as well?
![Screenshot_1](https://github.com/humhub/humhub/assets/35392110/c467a0c9-4fd8-41c9-aebe-44648b43e5ea)

yurabakhtin:
@ArchBlood Thank you for the report!
@luke- Fixed in the PR https://github.com/humhub/humhub/pull/7004, sorry, it was my error, I forgot to add the filter `show_at_directory = 1` for case when people is filtered.

PR #6812, opened 2024-01-23
Reduce people filter options depending on what filtered users have

yurabakhtin:
**What kind of change does this PR introduce?** (check at least one)

- [x] Feature

**The PR fulfills these requirements:**

- [x] It's submitted to the `develop` branch, _not_ the `master` branch if no hotfix
- [ ] All tests are passing
- [ ] New/updated tests are included
- [x] Changelog was modified

**Other information:**
https://github.com/humhub/humhub-internal/issues/120#issuecomment-1892505437
https://github.com/humhub/humhub/pull/6806#issuecomment-1904682427

PR #7004, opened 2024-05-16
Fix people filter by group

yurabakhtin:

- [x] Bugfix

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**
https://github.com/humhub/humhub/issues/6962#issuecomment-2113740229


luke- (approved):

yurabakhtin:
> first thanks for the quick action, although I now see a different issue when it comes to the group filter, not sure if this was intended or not;

@ArchBlood I think it is not an issue, it is a feature which was implemented [here](https://github.com/humhub/humhub/pull/6812), it works like this: if some filter is active then values of all filters are reduced to what the filtered users have. For your case, the group "Team" is became hidden because the filtered users are not in the Team group, and such behaviour works for other filter as well.

PR #7006, opened 2024-05-17
 Disable mustChangePassword check for impersonated access tokens

gevorgmansuryan:
https://github.com/humhub/rest/issues/162


luke- (changes requested):
@gevorgmansuryan Thank you for the PR.

Generally, no pure module-specific code should be added to the core.

Couldn't it be solved more independently, e.g. with a new “setImpersonator” method. This could also be used by the core itself.

ArchBlood:
@luke- why not do something like this in the rest module itself?
```php
use yii\base\Security;

public function actionImpersonate()
{
    if (!Yii::$app->user->isAdmin()) {
        throw new ForbiddenHttpException();
    }

    $userId = Yii::$app->request->getBodyParam('userId');
    $user = User::findOne(['id' => $userId]);

    if ($user === null) {
        throw new NotFoundHttpException();
    }

    // Check if the user must change password
    if ($user->mustChangePassword) {
        // Generate a random password
        $security = new Security();
        $defaultPassword = $security->generateRandomString(16, 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789'); // Use both letters and numbers, increase length to 16
        
        // Hash the password
        $passwordHash = Yii::$app->security->generatePasswordHash($defaultPassword);

        // Set the hashed password
        $user->password = $passwordHash;

        // Update the user's mustChangePassword attribute
        $user->mustChangePassword = false;
        if (!$user->save()) {
            // Handle errors if necessary
            throw new ErrorException("Failed to update user's mustChangePassword attribute.");
        }
    }

    // Delete any existing impersonation token for the user
    if ($token = ImpersonateAuthToken::findOne(['user_id' => $user->id])) {
        $token->delete();
    }

    // Create a new impersonation token
    $token = new ImpersonateAuthToken();
    $token->user_id = $user->id;
    $token->save();

    return [
        'token' => $token->token,
        'expires' => $token->expiration,
    ];
}
```

luke-:
@ArchBlood Hmm, I don't know if we should really just automatically set a random password for the affected users.

luke- (approved):

Issue #7005, opened 2024-05-17
Receive 'New Content' Notifications for spaces in global settings is not taking in account in the space


marc-farre:
### What steps will reproduce the problem?

2. Account settings -> Notification -> click "Reset to defaults"
3. Administration -> Settings -> Notifications
4. Add a space in "Receive 'New Content' Notifications for the following spaces"
5. Go to the space

### What is the expected result?

![image](https://github.com/humhub/humhub/assets/23310825/2386dd25-a069-4983-9752-0b7fea9cd898)

### What do you get instead?

![image](https://github.com/humhub/humhub/assets/23310825/ed5a0754-cc42-43ca-a1af-b0bb17f52b68)

### Additional info

Now, if you go to Account settings -> Notification -> and click "Save", you have the expected result:

![image](https://github.com/humhub/humhub/assets/23310825/2386dd25-a069-4983-9752-0b7fea9cd898)

Original discussion: https://community.humhub.com/content/perma?id=277789

| Q                | A
| ---------------- | ---
| HumHub version   | 1.15.3 and 1.16-beta-2

PR #7014, opened 2024-05-20
Fix flag of new content notification space

yurabakhtin:

- [x] Bugfix

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
https://github.com/humhub/humhub/issues/7005


protected/humhub/modules/space/widgets/HeaderControlsMenu.php
```diff
@@ -91,7 +91,7 @@ public function init()
         if ($this->space->isMember()) {
             $membership = $this->space->getMembership();
 
-            if (!$membership->send_notifications) {
+            if (!$membership->send_notifications && !Yii::$app->notification->hasSpace($this->space)) {
```
luke-, line 94: Do we still need the `!$membership->send_notifications` check here?

luke- (changes requested):

protected/humhub/modules/space/widgets/HeaderControlsMenu.php
yurabakhtin: Yes, it may be deleted, If we will remove the `!$membership->send_notifications` from here the result is not changed because inside the method `hasSpace()` this checking is done too but there is an additional sql query to the table `space_memberhip` for all spaces.
I have decided to keep it here only to avoid the sql query when this setting is already enabled per current User and Space.
I.e. the removing will make the code cleaner but a bit slower.

PR #7047, opened 2024-06-05
Update visibility of activities according to the parent object

yurabakhtin:

- [x] Bugfix

- [x] All tests are passing
- [x] Changelog was modified

**Other information:**
https://github.com/humhub/humhub-internal/issues/263

### DIFF
--- a/protected/humhub/modules/content/models/Content.php
+++ b/protected/humhub/modules/content/models/Content.php
@@ -137,6 +137,11 @@ class Content extends ActiveRecord implements Movable, ContentOwner, SoftDeletab
     const EVENT_STATE_CHANGED = 'changedState';
 
     /**
+     * @event Event is used when a Content visibility is changed.
+     */
+    const EVENT_VISIBILITY_CHANGED = 'changedVisibility';
+
+    /**
      * @inheritdoc
      */
     public function behaviors()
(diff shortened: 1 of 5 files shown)


luke- (approved):

protected/humhub/modules/content/models/Content.php
```diff
@@ -273,6 +278,10 @@ public function afterSave($insert, $changedAttributes)
             }
         }
 
+        if (array_key_exists('visibility', $changedAttributes)) {
```
luke-, line 281: @yurabakhtin Do we really need to introduce a new event here? Can't we use the `AfterSaveEvent` and use the `changedAttributes` from that?

https://www.yiiframework.com/doc/api/2.0/yii-db-aftersaveevent#$changedAttributes-detail

luke- (changes requested):

protected/humhub/modules/content/models/Content.php
yurabakhtin: @luke- You are right, we don't need that, thanks.
Fixed in the commit https://github.com/humhub/humhub/pull/7047/commits/be7900e0c30fb0b0be5c8bc74f37b13abb69a71a.

Also I think we don't need the event `Content::EVENT_STATE_CHANGED` because it can be handled with the event `EVENT_AFTER_UPDATE` which has all required values:

- `content` => `$event->sender`
- `newState` => `$event->sender->state`
- `previousState` => `$event->changedAttributes['state']` 

It was added by you [here](https://github.com/humhub/humhub/pull/6135/files#diff-b3de8a7bb84547e2de7081f31795f491fb051a7df7042d02d991041ee0235a51R252), maybe you remember why we need that, if no we could remove that as well.

luke-: @yurabakhtin Thanks!

I've created a new issue for that: https://github.com/humhub/humhub/issues/7048

luke- (approved):

PR #7106, opened 2024-07-05
Enable option for non-member users of a space to create posts

dvn0:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**

I was trying to configure my HumHub instance so that non-members could post in into Spaces without first joining the space. This is vital for the way my instance is used. As it stood, it was only possible for instance Admins to create posts in Spaces they were not members of.

The change from this PR will make it so that in a Space's permission settings the option "Create Posts" will become available to configure, and setting it to "Allow" will result in letting non-members create posts in the space. This solution was proposed by @marc-farre . Thanks, Marc!

I tested this on my development instance and it works nicely.

This shouldn't create any unexpected behaviour because the setting is by default set to "Deny".

Please see this HumHub Community post for more info: https://community.humhub.com/content/perma?id=279216

### DIFF
--- a/protected/humhub/modules/post/permissions/CreatePost.php
+++ b/protected/humhub/modules/post/permissions/CreatePost.php
@@ -34,7 +34,6 @@ class CreatePost extends BasePermission
      * @inheritdoc
      */
     protected $fixedGroups = [
-        Space::USERGROUP_USER,
         Space::USERGROUP_GUEST,
         User::USERGROUP_SELF,
         User::USERGROUP_GUEST,


ArchBlood:
@dvn0 please click and sign the CLA so that the workflow can continue, I've highlighted the places where you can click to accept the CLA;

Screenshot_1

dvn0:
Done, @ArchBlood

protected/humhub/modules/post/permissions/CreatePost.php
```diff
@@ -34,7 +34,6 @@ class CreatePost extends BasePermission
      * @inheritdoc
      */
     protected $fixedGroups = [
-        Space::USERGROUP_USER,
```
ArchBlood, line 37: Wouldn't it be better to instead of removing this line just move to to `$defaultAllowedGroups` array?

dvn0: Sure, I'll try that. Not sure what the implications are, either way, but that seems more logical.

dvn0: Okay, I tested it on my development instance, and indeed that works equally well. I've made the change, squashed, and force-pushed.

dvn0: Oh, I just realized that moving it to this `$defaultAllowedGroups` array means that the default for the CreatePost option will be set to `Allow`, of course! So is that what we want? It might be unexpected behavior for users.

dvn0: Simply removing it and not adding it to the aforementioned array keeps the default setting as `Deny`, but it just makes the option visible.

dvn0: Thoughts on this point @ArchBlood @luke- ?

luke-:
@dvn0 From my side, we can provide the possibility that non-space members can also create posts. However, it should not be allowed by default.

Can you please change that?

dvn0:
> @dvn0 From my side, we can provide the possibility that non-space members can also create posts. However, it should not be allowed by default.

Okay, @luke-, that makes sense, and I've changed it back to how I had it before, so it will not be allowed by default.

luke- (approved):

luke-:
@dvn0 Thanks, looks good for me. I've added it also to the CHANGELOG.md

PR #7125, opened 2024-07-17
Add self test for new migrations

yurabakhtin:

- [x] Feature

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
https://github.com/humhub/humhub-internal/issues/304


luke- (approved):

ArchBlood:
@yurabakhtin @luke- I recently had some issues when it came to this P/R displaying migrations for already installed migrations, as a temporary fix I switched to using the following `getPendingMigrations()` method which corrects the SelfTest issue;

```php
    public function getPendingMigrations(): array
    {
        $lastMigrationOutput = $this->getLastMigrationOutput();

        if ($this->runAction() === self::DB_ACTION_PENDING && preg_match_all('/(^|\s)(m\d+.+)(\n|$)/', $lastMigrationOutput, $matches)) {
            return $matches[2];
        }

    return [];
    }
```

Then to fix the issue in the admin database view, I switched to using `MigrationService::create()->getPendingMigrations()` instead of `MigrationService::DB_ACTION_PENDING` which stops displaying already installed migrations.

https://github.com/humhub/humhub/blob/a8ca3746ae50c4f57efcd2e68a28616c949d30b5/protected/humhub/modules/admin/views/information/database.php#L40

If there is a better solution I'm all for it but I haven't found any.

luke-:
@ArchBlood Can you please provide steps or more details on the issue you have?

ArchBlood:
> @ArchBlood Can you please provide steps or more details on the issue you have?

The only thing that I did was update a hybrid mix of the pre-packaged community edition and git from the `master` or the `develop` branch. Once that is done I navigate to either Database or Prerequisites and see new migrations are available, but those migrations are already applied with then stops me from being able to run migrations completely, the above method fixes this and displays new migrations correctly.

luke-:
Hmm, I'll try to reproduce. Can you also reproduce this on pure GIT or pure Packaged Install?

ArchBlood:
> Hmm, I'll try to reproduce. Can you also reproduce this on pure GIT or pure Packaged Install?

Seems to happen with Softaculous packaged installs and updating via a mix between the updater and composer modules then running migrations manually, sometimes already applied migrations are rechecked due to session cache. I'm not sure it is install specific as it doesn't rely on this and could happen at any point after any of the installation/update options.

luke-:
Hmm, so if the problem is not entirely clear and the installation is also rather special, I wouldn't  make any changes here.
If you can reproduce the problem on a pure Git environment or packaged environment, we can create an issue for this.

PR #7153, opened 2024-08-02
Fix content visibility of disabled users

yurabakhtin:

- [x] Bugfix

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
https://github.com/humhub/humhub-internal/issues/315

### DIFF
--- a/protected/humhub/modules/content/jobs/ReindexUserContent.php
+++ b/protected/humhub/modules/content/jobs/ReindexUserContent.php
@@ -1,0 +1,74 @@
+<?php
+/**
+ * @link https://www.humhub.org/
+ * @copyright Copyright (c) HumHub GmbH & Co. KG
+ * @license https://www.humhub.com/licences
+ */
+
+namespace humhub\modules\content\jobs;
+
+use humhub\modules\content\models\Content;
+use humhub\modules\content\Module;
+use humhub\modules\content\search\driver\AbstractDriver;
+use humhub\modules\content\services\SearchJobService;
+use humhub\modules\queue\interfaces\ExclusiveJobInterface;
+use humhub\modules\queue\LongRunningActiveJob;
+use humhub\modules\user\models\User;
+use Yii;
+
+class ReindexUserContent extends LongRunningActiveJob implements ExclusiveJobInterface
+{
+    public $userId;
+
+    /**
+     * @inhertidoc
+     */
+    public function getExclusiveJobId()
+    {
+        return 'content-search.reindex-user.' . $this->userId;
+    }
+
+    /**
+     * @inhertidoc
+     */
+    public function run()
+    {
+        return $this->getService()->run(function () {
+            $user = User::findOne(['id' => $this->userId]);
+            if (!$user) {
+                return;
+            }
+
+            $contents = Content::find()
+                ->where(['created_by' => $user->id]);
+
+            foreach ($contents->each() as $content) {
+                if ($user->status === User::STATUS_ENABLED) {
+                    $this->getSearchDriver()->update($content);
+                } else {
+                    $this->getSearchDriver()->delete($content);
+                }
+            }
+        });
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function canRetry($attempt, $error)
+    {
+        return $this->getService()->canRetry($attempt);
+    }
+
+    public function getService(): SearchJobService
+    {
+        return new SearchJobService();
+    }
+
+    private function getSearchDriver(): AbstractDriver
+    {
+        /** @var Module $module */
+        $module = Yii::$app->getModule('content');
+        return $module->getSearchDriver();
+    }
+}
--- a/protected/humhub/modules/content/jobs/ReindexUserContent.php
+++ b/protected/humhub/modules/content/jobs/ReindexUserContent.php
@@ -8,13 +8,11 @@
 namespace humhub\modules\content\jobs;
 
 use humhub\modules\content\models\Content;
-use humhub\modules\content\Module;
-use humhub\modules\content\search\driver\AbstractDriver;
+use humhub\modules\content\services\ContentSearchService;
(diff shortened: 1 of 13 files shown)


protected/humhub/modules/stream/models/StreamQuery.php
```diff
@@ -525,6 +526,12 @@ protected function setupCriteria()
             ->joinWith('contentContainer')
             ->limit($this->limit);
 
+        // Filter out content created by not enabled users
+        $this->_query->andWhere(['OR',
+            ['IS', 'user.id', new Expression('NULL')],
+            ['user.status' => User::STATUS_ENABLED],
+        ]);
+
```
luke-, line 534: What about the condition below? Isn't it responsible for such users?

protected/humhub/modules/content/search/driver/MysqlDriver.php
```diff
@@ -172,10 +172,16 @@ protected function addQueryFilterVisibility(ActiveQuery $query): ActiveQuery
     {
         $query->andWhere(['content.state' => Content::STATE_PUBLISHED]);
 
-        $query->joinWith('contentContainer');
+        $query->joinWith(['contentContainer', 'createdBy']);
         $query->leftJoin('space', 'contentcontainer . pk = space . id and contentcontainer .class=:spaceClass', [':spaceClass' => Space::class]);
         $query->leftJoin('user cuser', 'contentcontainer . pk = cuser . id and contentcontainer .class=:userClass', [':userClass' => User::class]);
 
+        // Filter out content created by not enabled users
+        $query->andWhere(['OR',
```
luke-, line 180: I don't think we can realise this for other search engines, as a simple join is not possible here.
Would an ActiveJob (RemoveUserContent, AddUserContent) be a solution here?

luke- (changes requested):
@yurabakhtin Thanks! Please see my remarks.

protected/humhub/modules/stream/models/StreamQuery.php
yurabakhtin: @luke- Yes, you are right the below condition is same, but it is not by default, i.e. by default contents of disabled users are visible in stream.
Should we keep it as it was before of change the property `humhub\modules\stream\Module::$showDeactivatedUserContent` to `false` by default?

protected/humhub/modules/content/search/driver/MysqlDriver.php
yurabakhtin: @luke- Maybe we should not index content if author status is not enabled, we could update the method [ContentSearchService::isIndexable()](https://github.com/humhub/humhub/blob/fix/315-disabled-user-content-readable/protected/humhub/modules/content/services/ContentSearchService.php#L81-L84) to:

```php
    public function isIndexable(): bool
    {
        return $this->content->stream_channel === Content::STREAM_CHANNEL_DEFAULT &&
            $this->content->createdBy->status === User::STATUS_ENABLED;
    }
```

protected/humhub/modules/stream/models/StreamQuery.php
luke-: @yurabakhtin I've created another issue for this here: https://github.com/humhub/team_tasks/issues/351 Please keep the flag as it is.

yurabakhtin: @luke- Ok, then I will remove the current code as it is a duplication.

protected/humhub/modules/content/search/driver/MysqlDriver.php
yurabakhtin: @luke- 

> But what should we do with already indexed user content?

Nothing, I think only next rebuilding will solve it. It can be solved by Mysql driver, but not for ZendLucence Driver.

yurabakhtin: Commit https://github.com/humhub/humhub/pull/7153/commits/60e9723075583069768b5c493bed3cec21a1901d for new changes.

luke-: @yurabakhtin Looks good. Another point would be an ActiveJob "ReindexUserContent" which deletes indexed content using `created_by` and reindex it?

protected/humhub/modules/content/jobs/ReindexUserContent.php
```diff
@@ -0,0 +1,74 @@
...
+            if (!$user) {
+                return;
+            }
+
+            $contents = Content::find()
+                ->where(['created_by' => $user->id]);
+
+            foreach ($contents->each() as $content) {
+                if ($user->status === User::STATUS_ENABLED) {
+                    $this->getSearchDriver()->update($content);
```
luke-, line 47: Maybe it's better here to use `ContentSearchService::update()`?

luke- (changes requested):
@yurabakhtin Thanks looks good. Please see one point below.

protected/humhub/modules/content/jobs/ReindexUserContent.php
yurabakhtin: @luke- Commit https://github.com/humhub/humhub/pull/7153/commits/a3cf6251275455397e8bcbb27f58a6c4c7683010.

protected/humhub/modules/content/services/ContentSearchService.php
```diff
@@ -38,9 +39,9 @@ public function update($asActiveJob = true): void
         }
     }
 
-    public function delete($asActiveJob = true): void
+    public function delete(bool $asActiveJob = true, bool $forceDelete = false): void
```
luke-, line 42: @yurabakhtin Not sure about the `forceDelete` option. I would prefer to remove `isIndexable()` here for all cases.

What do you think about?

luke- (changes requested):

luke- (approved):

Issue #7113, opened 2024-07-11
Groups - Default permissions/Space creation


Semir1212:
When a new group is created, it is given the permission to create private/public spaces by default. We should change the default settings so that users do not get this permission by accident. 

Create Private Spaces --> Deny
Create Public Spaces --> Deny

luke-:
Should we allow this for the Default User Group? 
I would say yes, you can always make it more restrictive later on.

marc-farre:
I have often had to deny these default permissions in the config file, so I also think they should be denied by default.
For the Default User Group, I think it's better to deny them too, because:
- When we create a new group, we expect the default permissions to be the same as the User group (it might be confusing to have different default permissions).
- When installing a new HumHub instance, it's logical and safer to reserve Space creation for admins only by default.

Maybe a migration for existing HumHub instances could configure all existing Groups with the existing Space creation permissions?

Semir1212:
I agree with @marc-farre, it should also be denied for the default user group.

luke-:
Ok, we can do it that way too.

For me, the main pro reason was to demonstrate that HumHub can also be freely organised (anyone can create Spaces) without administration.

---

@marc-farre So you mean a migration to manually set the permissions to "Allow" for existing installations?

marc-farre:
> @marc-farre So you mean a migration to manually set the permissions to "Allow" for existing installations?

Yes. But it mustn't apply to the User Group when installing a new HumHub instance...

PR #7160, opened 2024-08-08
Enh/default group permissions change

gevorgmansuryan:
https://github.com/humhub/humhub/issues/7113

### DIFF
--- a/protected/humhub/migrations/m240807_230603_add_permissions_to_existing_user_groups.php
+++ b/protected/humhub/migrations/m240807_230603_add_permissions_to_existing_user_groups.php
@@ -20,16 +20,27 @@ class m240807_230603_add_permissions_to_existing_user_groups extends Migration
 
             /** @var Group $group */
             foreach ($groups->each() as $group) {
-                Yii::$app->user->permissionManager->setGroupState(
-                    $group->id,
-                    CreatePrivateSpace::class,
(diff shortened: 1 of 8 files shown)


protected/humhub/migrations/m240807_230603_add_permissions_to_existing_user_groups.php
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function safeUp()
+    {
+        if (Yii::$app->isInstalled()) {
+            $groups = Group::find();
+
+            /** @var Group $group */
+            foreach ($groups->each() as $group) {
```
luke-, line 22: @gevorgmansuryan Do you see a chance to handle this by raw db queries? 

For migrations, I would use as little business logic as possible, as this could change in the future and might have to be migrated as well.

luke- (changes requested):
@gevorgmansuryan Thanks, looks good for me. One small point :-)

protected/humhub/migrations/m240807_230603_add_permissions_to_existing_user_groups.php
gevorgmansuryan: @luke- got it. Will be done

luke- (approved):

PR #7167, opened 2024-08-12
Improve debug mode handling v2

gevorgmansuryan:
https://github.com/humhub/humhub-internal/issues/288


protected/config/.env.example
```diff
@@ -0,0 +1,3 @@
+DEBUG=1
```
luke-, line 1: I'm a little worried that the ‘.env’ file can simply be called up via the browser if the web server is configured incorrectly. Are there common approaches against this? e.g. an ‘env.php’ as you have already mentioned?

protected/config/common.php
```diff
@@ -4,5 +4,20 @@
  * @see http://www.yiiframework.com/doc-2.0/guide-concept-configurations.html
  * @see https://docs.humhub.org/docs/admin/advanced-configuration
  */
+
```
luke-, line 7: This file is for HumHub administrators and should remain empty by default.
Also, changes here quickly lead to conflicts in Git :-)

I'm sure we can handle the ENV Module Paths somewhere else.

luke- (changes requested):
@gevorgmansuryan Thanks, I like the current approach.


Please see idea regarding var naming here. https://github.com/humhub/humhub-internal/issues/288#issuecomment-2295227261

protected/config/.env.example
gevorgmansuryan: @luke- I placed it in protected directory, it's protected via htaccess, no one can open it via browser

protected/config/env.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+$env = [];
```
luke-, line 3: If I understand it correctly, this file should transfer ENV variables to the config file. I like this.

But lets move the file to `/protected/humhub/config`, because 
- /protected/config (local configuration)
- /protected/humhub/config (base configuration, not changeable)

protected/config/.env.example
luke-: Ok, in our standard configurations, this directory should be protected. However, I have also seen some installations where this is not the case.  

Anyway, this is more of a power user feature and even optional. Perhaps we can also add a clear header to the `.env.example`.

e.g. 

```env
# Warning: This file may contain sensitive information. Make sure that this file is not
# distributed via the web server and does not have file permissions that are too open.
```
---

I am currently thinking about whether we should put the `.env[.example]` in the project root directory. Here it is more visible and we could also adjust the documentation/example configurations so that this file is also protected.

What do you think?

luke- (changes requested):

protected/config/.env.example
gevorgmansuryan: @luke- good idea to put it to root, it will be visible to everyone. let's add this header. But anyway we can prevent opening this files via .htaccess too in same way that it's done for example for composer.json

protected/humhub/components/SettingsManager.php
```diff
@@ -134,6 +134,10 @@ public function isFixed(string $name): bool
      */
     public function getFixed(string $name)
     {
+        if ($value = $_ENV["{$this->moduleId}_{$name}"] ?? null) {
```
luke-, line 137: We can probably remove this with the automatic mapping between ENV and Config, right?

protected/humhub/config/env.php
```diff
@@ -0,0 +1,49 @@
...
+use yii\helpers\ArrayHelper;
+use yii\helpers\Inflector;
+use yii\helpers\Json;
+use yii\helpers\StringHelper;
+
+$env = [];
+
+$preparePath = function (string $key): array {
+    return ArrayHelper::getColumn(
+        explode('__', $key),
```
luke-, line 14: Somehow the syntax e.g. `HUMHUB_SETTINGS_BASE__MAILER_...`  looks like a typo for me.

Any idea for other variants?

- Maybe `HUMHUB_SETTINGS_BASE_MAILER-DSN`   (See `_` as new array node, and replace `.`   with `-`, since a `.` is not supported?

luke- (changes requested):
@gevorgmansuryan I really like the approach that you can set any config values via ENV. 

A few UNIT tests for the different variants would also be good here.

protected/humhub/config/env.php
gevorgmansuryan: @luke- replaced `__` with `.`, `-` is not supported. now it's:
`HUMHUB_PARAMS.MODULE_AUTOLOAD_PATHS`=["/app/modules/humhub","/app/modules/humhub-contrib"]
`HUMHUB_SETTINGS.BASE.MAILER.DSN`="smtp://...."
`HUMHUB_SETTINGS.BASE.MAILER.TRANSPORT_TYPE`="php"
`HUMHUB_SETTINGS.BASE.MAILER.SYSTEM_EMAIL_ADDRESS`="noreply@humhub.com"

gevorgmansuryan: Also moved env.php file functionality to separate helper for unit test. Tests not ready yet

luke-: @gevorgmansuryan Thank you, I like that.

But I hoped that we could now simply overwrite all config values via the ENV variables. e.g.
`HUMHUB_CONFIG_COMPONENTS_URLMANAGER_SHOWSCRIPTNAME=false`

But if I understand correctly, it's not possible because of Upper/LowerCase, is it?

If we use this syntax it should work right? 
`HUMHUB_CONFIG_components_urlManager_showScriptName=false`
`HUMHUB_CONFIG_params_fixedSettings_base_mailer.transportType=php`

gevorgmansuryan: @luke- I can enable this too. currently there is 2 namespaces
HUMHUB_PARAMS - it's config->params
HUMHUB_SETTINGS - fixed settings

i can add HUMHUB_COMPONENTS namespace too and this will work also
HUMHUB_COMPONENTS.URL_MANAGER.SHOW_SCRIPT_NAME=false

gevorgmansuryan: or we can have one namespace for all

`HUMHUB_CONFIG.PARAMS.MODULE_AUTOLOAD_PATHS`=["/app/modules/humhub","/app/modules/humhub-contrib"]
`HUMHUB_CONFIG.PARAMS.BASE.MAILER.SYSTEM_EMAIL_ADDRESS`="[noreply@humhub.com](mailto:noreply@humhub.com)"
`HUMHUB_CONFIG.COMPONENTS.URL_MANAGER.SHOW_SCRIPT_NAME`=false

luke-: @gevorgmansuryan One namespace is probably best.
But how does the conversation from UpperCase to camelCase work?

I'm not quite sure how you want to automatically convert `HUMHUB_CONFIG.COMPONENTS.URL_MANAGER.SHOW_SCRIPT_NAME` to `$config[‘components’][‘urlManager’][‘showScriptName’]`.

gevorgmansuryan: @luke- I'm using yii\helpers\Inflector::variablize to do this. but there is one problem, both env and variablize do not support `-` symbol. So with one namespace we can not handle 'fixed-settings' or any other config key with `-` symbol. Let's keep 2 namespaces HUMHUB_SETTINGS for fixed settings, and `HUMHUB_CONFIG` for everything

luke-: @gevorgmansuryan Ok, that looks like a good solution!

So we use:
- `.` for the configuration depth / array
- `_` to convert uppercase vars into camelcase

So we could cover the manipulation of the most config array values. Do you see config array elements that cannot be manipulated via ENV or other restrictions? That would be good for the documentation.

---



Please use `HUMHUB_FIXED_SETTINGS` as a second namespace. These FS are rarely used and should be clearly labelled.

index.php
```diff
@@ -6,28 +6,33 @@
  * @license https://www.humhub.com/licences
  */
 
-// comment out the following two lines when deployed to production
```
dreua, line 9: This needs to be updated in the documentation here: https://docs.humhub.org/docs/admin/security/#enable-production-mode

luke-: @dreua Thank you, we will adapt the documentation accordingly.
In future, debug mode will be automatically deactivated after installation.

protected/humhub/config/env.php
gevorgmansuryan: @luke- It covers almost evrything, Yii 100% keeps the `lowerCamelCase` in base configs. 
.env does not support space and dots. For example `['components']['db']['on afterOpen']`. But it can be done using json, like this:
HUMHUB_CONFIG.COMPONENTS.DB = {"on afterOpen": ["humhub\libs\Helpers", "SqlMode"]}

gevorgmansuryan: Finally. 
this:
image
will be converted to this:
image

luke-:
@gevorgmansuryan I like the current solution.

The following points are still open, right?

- Tests
- Documentation (Here we should remove old references to the debug mode, e.g. here https://docs.humhub.org/docs/admin/installation#enable-production-mode and create a paragraph here, enable debug mode: https://docs.humhub.org/docs/admin/troubleshooting 

Or did I missed something else?

PR #7187, opened 2024-08-27
Fix search reindexing after content deletion

yurabakhtin:

- [x] Bugfix

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
https://github.com/humhub/humhub-internal/issues/344

### DIFF
--- a/protected/humhub/modules/content/services/ContentSearchService.php
+++ b/protected/humhub/modules/content/services/ContentSearchService.php
@@ -15,21 +15,11 @@ use Yii;
 
 class ContentSearchService
 {
-    public ?Content $content;
-    public ?int $contentId;
+    public Content $content;
 
-    /**
-     * @param Content|int $content
-     */
-    public function __construct($content)
+    public function __construct(Content $content)
     {
-        if ($content instanceof Content) {
-            $this->content = $content;
-            $this->contentId = $content->id ?? null;
-        } elseif (is_int($content)) {
-            $this->content = Content::findOne(['id' => $content]);
-            $this->contentId = $content;
-        }
+        $this->content = $content;
     }
 
     public function update(bool $asActiveJob = true): void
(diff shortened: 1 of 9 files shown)


protected/humhub/modules/content/services/ContentSearchService.php
```diff
@@ -15,11 +15,21 @@
...
 {
-    public Content $content;
+    public ?Content $content;
+    public ?int $contentId;
 
-    public function __construct(Content $content)
+    /**
+     * @param Content|int $content
+     */
+    public function __construct($content)
```
luke-, line 24: @yurabakhtin I would prefer when the ContentSearch service always reliably has a non-null $content record.

What do you think of the following solution, we create a new static method `ContentSearchService::deleteContentById($id, $asActiveJob)`  
and also use this method in `ContentSearchService::delete($asActiveJob)`.

luke- (changes requested):

protected/humhub/modules/content/services/ContentSearchService.php
yurabakhtin: @luke- Done in the commit https://github.com/humhub/humhub/pull/7187/commits/5b8741fd3e237b7b3b6977f518f4231c2c764b53.

What do you think if we will implement new static method `instance()` for the service like this? :

```php
    public static function instance(Content $content): self
    {
        return new self($content);
    }
```

so then we will update all codes from:

- `(new ContentSearchService($content))->methodName()`

to:

- `ContentSearchService::instance($content)->methodName()`

I think it looks better without the brackets `()` around.

luke- (approved):

protected/humhub/modules/content/services/ContentSearchService.php
luke-: @yurabakhtin With a method `className::instance()` (getInstance) I always somehow think it is a singleton pattern. https://en.wikipedia.org/wiki/Singleton_pattern

---

Since this affects many other service classes too, we should try to find here a solution/style we use in the whole app. I've createda seperate issue for it here: https://github.com/humhub/humhub-internal/issues/347

Issue #7148, opened 2024-07-31
Improve UX on search popup


samuk:
### Is your feature request related to a problem? Please describe.

When searching for a term in the (new?) search popup I often need to scroll through many pages of content before finding the users who have that term in their /about content

### Describe the solution you'd like

1) Have an 'advanced search' link at the top of the popup/ modal that opens a full page to return more results

### Describe alternatives you've considered

2) Make the content types collapsible? In my example above if I could click on 'content' and that collapsed all the 'content' results I could more easily find my way to the people

cc @marc-farre

ArchBlood:
See the advanced search module;
https://marketplace.humhub.com/module/advanced-search/description

Eladnarlea:
@samuk are you referring to something like this?

<img width="416" alt="Screenshot 2024-08-01 at 5 49 36 AM" src="https://github.com/user-attachments/assets/1c24d36a-42bd-482c-a36c-72e110c7627a">

marc-farre:
The issues are with the Link Preview module.

1/ When adding a link, the text is not truncated anymore:

![image](https://github.com/user-attachments/assets/bd08c3cc-cd14-467d-a75b-3c6d723f6a90)

2/ We shouldn't display the "Allow content from external source" warning:

![image](https://github.com/user-attachments/assets/b67f6aa0-2172-4b43-89f9-954e4f39dbb0)

luke-:
@yurabakhtin Is the wrong richtext parser perhaps being used here?
I would rather expect a plain output here. This should not contain ‘Translate’, ‘LinkPreview’, etc., should it?

samuk:
> 2/ We shouldn't display the "Allow content from external source" warning:

Shouldn't we? I get that it's ugly from a UX perspective, but if we're taking that seriously we shouldn't leak user IP without consent?

marc-farre:
> > 2/ We shouldn't display the "Allow content from external source" warning:
> 
> Shouldn't we? I get that it's ugly from a UX perspective, but if we're taking that seriously we shouldn't leak user IP without consent?

In search results displayed in the dropdown box, we should display plain text (no preview or external content). So this warning is useless here.

yurabakhtin:
@luke- @marc-farre Currently it works like this on the version `1.16.2 (Unreleased)`:

![render-meta-search](https://github.com/user-attachments/assets/82b6ccd3-79e6-42d6-8d6d-cd18d21af2b6)

[...]

PR #7195, opened 2024-08-30
Improve rendering of meta search content

yurabakhtin:

- [ ] Bugfix
- [ ] Feature

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**
https://github.com/humhub/humhub/issues/7148#issuecomment-2320641330

### DIFF
--- a/protected/humhub/modules/content/search/SearchRecord.php
+++ b/protected/humhub/modules/content/search/SearchRecord.php
@@ -58,11 +58,11 @@ class SearchRecord implements MetaSearchResultInterface
      * @param int $maxWordNumberBeforeKeyword
      * @return string
      */
-    protected function cutStringToKeyword(string $string, int $maxWordNumberBeforeKeyword = 1): string
+    private function cutStringToKeyword(string $string, int $maxWordNumberBeforeKeyword = 1): string
     {
         $index = stripos($string, $this->keyword);
 
-        if ($index < 40) {
+        if ($index === false || $index < 40) {
             // Don't cut if the keyword is almost at the beginning
             return $string;
         }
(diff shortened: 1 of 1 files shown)


protected/humhub/modules/content/search/SearchRecord.php
```diff
@@ -47,19 +47,36 @@ public function getImage(): string
...
+     *
+     * @param string $string
+     * @param int $maxWordNumberBeforeKeyword
+     * @return string
+     */
+    protected function cutStringToKeyword(string $string, int $maxWordNumberBeforeKeyword = 1): string
+    {
+        $index = stripos($string, $this->keyword);
+
+        if ($index < 40) {
```
luke-, line 65: Does that make sense? Often there will be no hit here either.

```suggestion
        if ($index === false || $index < 40) {
```

protected/humhub/modules/content/search/SearchRecord.php
```diff
@@ -47,19 +47,36 @@ public function getImage(): string
...
+    }
+
+    /**
+     * Cut string to a word before first word contained the searched keyword
+     *
+     * @param string $string
+     * @param int $maxWordNumberBeforeKeyword
+     * @return string
+     */
+    protected function cutStringToKeyword(string $string, int $maxWordNumberBeforeKeyword = 1): string
```
luke-, line 61: If not used somewhere else, I would tend to make "private" until we need it.

luke- (changes requested):

protected/humhub/modules/content/search/SearchRecord.php
yurabakhtin: ok, done https://github.com/humhub/humhub/pull/7195/commits/78cfa3d44d8d2eb34beae55fbc32859c5802db51#diff-c4cb0fcf397d1272c6dd36f78b88a309cf38639b6947ca4258ca8eccd8d25e93R65 to avoid some restriction/error in future php versions, but now it works well because `(false < 40) === true`.

luke- (approved):

PR #7197, opened 2024-08-30
Increased version of `PHPOffice/PHPSpreadsheet` to v2.2+

gevorgmansuryan:
https://github.com/humhub/humhub-internal/issues/349

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
     },
     "minimum-stability": "stable",
     "require": {
-        "php": ">=8.0",
+        "php": ">=8.1",
         "ext-exif": "*",
         "ext-intl": "*",
         "ext-json": "*",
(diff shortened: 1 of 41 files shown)


gevorgmansuryan:
@luke- 
1. no BC for migrating from PhpSpreadsheet v2.1.0 to PhpSpreadsheet v2.2.2
2. we have to change required PHP minimum version from. 8.0 to 8.1 to update PhpSpreadsheet

ArchBlood (approved):

composer.json
```diff
@@ -13,7 +13,7 @@
     },
     "minimum-stability": "stable",
     "require": {
-        "php": ">=8.0",
+        "php": ">=8.1",
```
luke-, line 16: An increase in the minimum PHP version is a little more substantial and should also be coordinated first. 
See staff notes here: https://staff.humhub.org/s/development/wiki/146/increase-min-php-version

In this case, however, it works, as this can be planned in v.17.

luke- (approved):

PR #7227, opened 2024-09-24
Fix search reindexing after create new content

yurabakhtin:

- [x] Bugfix

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
https://github.com/humhub/custom-pages/issues/347

### DIFF
--- a/protected/humhub/modules/content/services/ContentSearchService.php
+++ b/protected/humhub/modules/content/services/ContentSearchService.php
@@ -44,8 +44,12 @@ class ContentSearchService
         self::deleteContentById($this->content->id, $asActiveJob);
     }
 
-    public static function deleteContentById(int $id, bool $asActiveJob = true): void
+    public static function deleteContentById(?int $id, bool $asActiveJob = true): void
     {
+        if (empty($id)) {
+            return;
+        }
+
         if ($asActiveJob) {
             Yii::$app->queue->push(new SearchDeleteDocument(['contentId' => $id]));
         } else {
--- a/protected/humhub/modules/content/services/ContentSearchService.php
+++ b/protected/humhub/modules/content/services/ContentSearchService.php
@@ -44,12 +44,8 @@ class ContentSearchService
         self::deleteContentById($this->content->id, $asActiveJob);
     }
 
-    public static function deleteContentById(?int $id, bool $asActiveJob = true): void
+    public static function deleteContentById(int $id, bool $asActiveJob = true): void
     {
-        if (empty($id)) {
-            return;
-        }
-
         if ($asActiveJob) {
             Yii::$app->queue->push(new SearchDeleteDocument(['contentId' => $id]));
         } else {
@@ -86,6 +82,10 @@ class ContentSearchService
 
     public function isIndexable(): bool
     {
+        if (empty($this->content->id)) {
+            return false;
+        }
+
         if ($this->content->stream_channel !== Content::STREAM_CHANNEL_DEFAULT) {
             return false;
         }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ HumHub Changelog
 - Fix #7174: In lists, when an item text is displayed on multiple lines, the lines below are not lining up with the first one
 - Fix #7222: Fix rendering of checkbox on MacOS and iOS
 - Fix #7225: Fix module JS config initialisation on AJAX request
+- Fix #7227: Fix search reindexing after create new content
 
 1.16.2 (September 5, 2024)
 --------------------------


protected/humhub/modules/content/services/ContentSearchService.php
```diff
@@ -44,8 +44,12 @@ public function delete(bool $asActiveJob = true): void
         self::deleteContentById($this->content->id, $asActiveJob);
     }
 
-    public static function deleteContentById(int $id, bool $asActiveJob = true): void
+    public static function deleteContentById(?int $id, bool $asActiveJob = true): void
```
luke-, line 47: @yurabakhtin 

In which situations the `deleteContentById` is called with `null` as first param?
Wouldn't it be better to prevent the method from being called without an ID?

luke- (changes requested):

protected/humhub/modules/content/services/ContentSearchService.php
yurabakhtin: @luke- I have investigated why the Custom Page was broken on a creating, and I have created the PR https://github.com/humhub/custom-pages/pull/348 in order to solve the error even without this current core PR, but I think it would be good to merge this PR to in order to avoid similar issues from other modules.

Why it happened:

- The Custom Page model had the following code order in the method [afterSave](https://github.com/humhub/custom-pages/pull/348/files#diff-3efc7636e2cf4923ac42d418dae03019e1d4d2ae6be379e37023c2e8f93ff3b8R520):
  - `RichText::postProcess()` => inside this method we call `ContentSearchService::deleteContentById($content->id)`
  - `parent::afterSave()` => but only here we insert the `Content` record into DB.

So in the Custom Page module's PR I have decided to run the method `parent::afterSave()` before the `RichText::postProcess()`. I find most of case we have such order, but in several case we have the wrong order, for example:

- [Meeting Item](https://github.com/humhub/meeting/blob/master/models/MeetingItem.php#L123) - but it doesn't produce the error because the model has `$streamChannel = null` so `ContentSearchService::isIndexable()` is false and the method `deleteContentById()` is not called,
- [AbstractMessageEntry](https://github.com/humhub/mail/blob/master/models/AbstractMessageEntry.php#L136) - is not Content Active Record, the `ContentSearchService` is not run at all here.

I don't think we should update the modules, but I think we should merge the current core PR to avoid such error, because no need to try find a record for new creating Content record which has no ID yet.

luke-: @yurabakhtin Thanks for the explanation.

When we merge the core PR, we would just silently ignore such bugs in the order.

Would the content appear in the search index if we simply skipped the error when creating it?

luke-: > RichText::postProcess() => inside this method we call ContentSearchService::deleteContentById($content->id)

Could we fix the issue in the "postProcess" method?

yurabakhtin: @luke- 

> When we merge the core PR, we would just silently ignore such bugs in the order.

You are right, ok we can keep it and fix only those module where we will detect it later.

> Would the content appear in the search index if we simply skipped the error when creating it?

Yes, it must appear in search index because we call `(new ContentSearchService($this))->update()` inside the `Content::afterSave()`.

But the problem was in that on run `RichText::postProcess()` we call `humhub\modules\file\components\FileManager::attach()` that calls `(new ContentSearchService($this->record->content))->update()` but the `$this->record->content` is not saved yet, so `id` is `null`. The code is [here](https://github.com/humhub/humhub/blob/master/protected/humhub/modules/file/components/FileManager.php#L86).

---

> > RichText::postProcess() => inside this method we call ContentSearchService::deleteContentById($content->id)
> 
> Could we fix the issue in the "postProcess" method?

I can suggest to don't run `(new ContentSearchService($this->record->content))->update()` in the method `FileManager::attach()` when `$this->record->content->isNewRecord`, however I prefer to do this inside the class `ContentSearchService`, I think better is to modify the method `ContentSearchService::isIndexable()` if `$this->content->id` is empty to avoid all wrong cases when it is called for `Content` record which is not stored in DB.
I think we should run all ContentSearchService's methods only when Content has a not empty ID, otherwise it is impossible to detect a record inside the indexed data by any search driver.

luke-: @yurabakhtin 

> > > RichText::postProcess() => inside this method we call ContentSearchService::deleteContentById($content->id)
> > 
> > 
> > Could we fix the issue in the "postProcess" method?
> 
> I can suggest to don't run `(new ContentSearchService($this->record->content))->update()` in the method `FileManager::attach()` when `$this->record->content->isNewRecord`, however I prefer to do this inside the class `ContentSearchService`, I think better is to modify the method `ContentSearchService::isIndexable()` if `$this->content->id` is empty to avoid all wrong cases when it is called for `Content` record which is not stored in DB. I think we should run all ContentSearchService's methods only when Content has a not empty ID, otherwise it is impossible to detect a record inside the indexed data by any search driver.

Sounds good to me. Can you adapt the PR accordingly?

yurabakhtin: @luke- Done https://github.com/humhub/humhub/pull/7227/commits/d00a61e1d74fb896fbdea85d825d1368a00de21b.

luke- (approved):

PR #7253, opened 2024-10-10
CSV export improvements

gevorgmansuryan:
https://github.com/humhub/humhub-internal/issues/393


protected/humhub/components/export/SpreadsheetExport.php
```diff
@@ -359,6 +360,32 @@ protected function composeBodyRow($spreadsheet, $model, $key, $index)
...
     }
 
+    /**
+     * Sanitize value to prevent injection.
+     */
+    private function sanitizeValue(?string $value): ?string
+    {
+        if (!empty($value) && $this->resultConfig['writerType'] === 'csv') {
+            // Extended list of risky starting characters
+            $riskyCharacters = ['=', '+', '-', '@', '|', '/', '\\', '\'', '!', '"', '^', '{', '}', '[', ']', '%', '&', ':', ';'];
```
ArchBlood, line 370: Personally speaking, I'm not sure if we should extend this even more like this or only do a more direct method;

```php
$riskyCharacters = ['=', '+', '-', '@', '|', '/', '\\', '\'', '"', '<', '>', '^', '{', '}', '[', ']', '%', '&', ':', ';', '(', ')', '`', '*', '$', '#', '.', ',', '?'];
```

I'm only  concerned that this method would block use of characters that could be used safely and are commonly used by the general user (i.e. `#`, `@` or `&`) if sanitizing is needed `Html::encode()` or `HtmlPurifier::process()` should be used regularly.

Another approach we could use is as follows which should be faster and more efficient for what's needed;

```php
/**
 * Sanitize value to prevent injection in CSV.
 */
private function sanitizeValue(?string $value): ?string
{
    // Early return if value is empty or not handling CSV
    if (empty($value) || $this->resultConfig['writerType'] !== 'csv') {
        return $value;
    }

    // Check for risky starting characters or formula-like values
    $riskyCharacters = '=+-@|/\\\'"^{}[]%&:;';
    if (strpbrk($value[0] ?? '', $riskyCharacters) !== false || preg_match('/^\d+[+\-*\/].+/', $value)) {
        $value = "'" . $value; // Prepend single quote to neutralize potential injection
    }

    // Sanitize values with special CSV characters (quotes, commas, newlines)
    if (strpbrk($value, "\"\n,") !== false) {
        $value = '"' . str_replace('"', '""', $value) . '"'; // Escape double quotes and wrap in quotes
    }

    return $value;
}
```

Original method: ~0.02624 seconds for 10,000 iterations
Updated method: ~0.02538 seconds for 10,000 iterations

gevorgmansuryan: @ArchBlood II think `Html::encode()` and `HtmlPurifier::process()` will not help us in CSV in any way, as we need plain text here.

Initial riskyCharacters list was extended version, we can shorten it to meet [OVASP requirments](https://owasp.org/www-community/attacks/CSV_Injection) only.
So it will: $riskyCharacters = ['=', '+', '-', '@', ',', ';', 0x09, 0x0D]

ArchBlood: Then we can use this in this case which should follow QWASP requirements;

```php
/**
 * Sanitize value to prevent injection in CSV.
 */
private function sanitizeValue(?string $value): ?string
{
    // Early return if value is empty or not handling CSV
    if (empty($value) || $this->resultConfig['writerType'] !== 'csv') {
        return $value;
    }

    // Check for risky starting characters or formula-like values
    $riskyCharacters = '=+-@,;'. "\t" . "\r"; // Include tab (0x09) and carriage return (0x0D)
    if (strpbrk($value[0] ?? '', $riskyCharacters) !== false || preg_match('/^\d+[+\-*\/].+/', $value)) {
        $value = "'" . $value; // Prepend single quote to neutralize potential injection
    }

    // Sanitize values with special CSV characters (quotes, commas, newlines)
    if (strpbrk($value, "\"\n,") !== false) {
        $value = '"' . str_replace('"', '""', $value) . '"'; // Escape double quotes and wrap in quotes
    }

    return $value;
}
```

CHANGELOG.md
```diff
@@ -46,6 +46,7 @@ HumHub Changelog
 - Enh #7244: Changed .env var depth separator from `.` to `__`
 - Fix #7248: Upgrade jQuery Highlight plugin
 - Fix #7254: Fix Login view HTML element ID from `user-auth-login-modal` to `user-auth-login`
+- Enh #7253: CSV export improvements
```
luke-, line 49: @gevorgmansuryan Does this only apply to CSV? Do we need some other improvements for XLSX?

luke- (changes requested):

CHANGELOG.md
gevorgmansuryan: @luke- While XLSX is more advanced and has protection, we can apply the same sanitizing for XLSX too, there will be no harm. Apply?

ArchBlood: @luke- if you think it is needed then the following could be used as an example;

```php
/**
 * Sanitize value to prevent injection in CSV and XLSX.
 */
private function sanitizeValue(?string $value): ?string
{
    if (empty($value) || !in_array($this->resultConfig['writerType'], ['csv', 'xlsx'])) {
        return $value;
    }

    // Check for risky starting characters or formula-like values and prepend single quote
    if (strpbrk($value[0], '=+-@,;' . "\t" . "\r") !== false || preg_match('/^\d+[+\-*\/].+/', $value)) {
        $value = "'" . $value;
    }

    // Sanitize escaping quotes, wrapping in double quotes if needed
    if (strpbrk($value, "\"\n,") !== false) {
        $value = '"' . str_replace('"', '""', $value) . '"';
    }

    return $value;
}
```

ArchBlood: @luke- @gevorgmansuryan I'd like to mention that adding XLSX in an array will increase the execution time but shouldn't be noticeable by the end user.

ArchBlood (approved):

PR #7298, opened 2024-11-07
Don't check email for existing on password recovery

yurabakhtin:

- [ ] Bugfix
- [ ] Feature

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
https://github.com/humhub/humhub-internal/issues/441#issuecomment-2462073750


luke- (changes requested):
- We also need to run this check, after the mail validation: https://github.com/humhub/humhub/blob/master/protected/humhub/modules/user/models/forms/AccountRecoverPassword.php#L70-L75 

- Can you please rebase to `develop`.

yurabakhtin:
> * We also need to run this check, after the mail validation: https://github.com/humhub/humhub/blob/master/protected/humhub/modules/user/models/forms/AccountRecoverPassword.php#L70-L75

@luke- Do you mean we should remove the code completely?
We will not display the error:

err-mes

We will display the message:

after-recover

and don't send any email to such users, right?

luke-:
@yurabakhtin We should also send an e-mail to the users. Only after they click the link in the e-mail should the error message appear.

yurabakhtin:
> @yurabakhtin We should also send an e-mail to the users. Only after they click the link in the e-mail should the error message appear.

@luke- Commit https://github.com/humhub/humhub/pull/7298/commits/13828d42cf8b02be81bb1c7285a49a8e298bd22b:

no-password-recovery

Please note I used the error message `'Password recovery disabled. Please contact your system administrator.` that was used before.

For users with password it works this way:

has-password-recovery

luke- (approved):

PR #7311, opened 2024-11-17
improve space picker focus

gevorgmansuryan:
https://github.com/humhub/humhub-internal/issues/446

### DIFF
--- a/protected/humhub/modules/space/widgets/SpacePickerField.php
+++ b/protected/humhub/modules/space/widgets/SpacePickerField.php
@@ -5,7 +5,6 @@ namespace humhub\modules\space\widgets;
 use humhub\modules\space\models\Space;
 use humhub\modules\ui\form\widgets\BasePicker;
 use Yii;
-use yii\helpers\Html;
 
 /**
  * Mutliselect input field for selecting space guids.
--- a/protected/humhub/modules/ui/form/widgets/BasePicker.php
+++ b/protected/humhub/modules/ui/form/widgets/BasePicker.php
@@ -422,6 +422,7 @@ abstract class BasePicker extends JsInputWidget
             'input-too-long' => Yii::t('UserModule.chooser', 'You reached the maximum number of allowed characters ({n}).', ['n' => $this->maxInput]),
             'default-results' => $this->getDefaultResultData(),
             'disabled' => $this->disabled,
+            'width' => '100%',
         ];
 
         if (!empty($this->disabledItems)) {


yurabakhtin:
@luke- I don't understand what problem you have there. For me even without the chages it works well, i.e. the green area is active for me to make the field focus on click, please review my video:

https://github.com/user-attachments/assets/453dee1d-b1e2-4988-9518-c4a9204b5351

I have tested on Windows Chrome browser without the changes from this PR.

luke-:
@yurabakhtin It's strange, it also/sometimes always works for me after the first focus. Even after a reload.  But not reliable...

@gevorgmansuryan Could you reproduce my problem?

gevorgmansuryan:
@luke- it can be reproduced only when page is loading with ajax

gevorgmansuryan:
@yurabakhtin here is it, entire page should be loaded via ajax (pjax)
https://github.com/user-attachments/assets/f27ac0ea-4b14-4487-b4b1-a7db79ed7923

yurabakhtin:
@gevorgmansuryan Thank you for the video, now I have understood the problem is in the short width of the input so you see the placeholder as cropped. I cannot reproduce it on my side as you show, however I agree we should fix the short input because it may be cropped on some language with long translated text.

But please see how the input looks when I select at least one item there:

new-line

inspect-code

I think it can be fixed if we use `display: flex` for the `ul`(wrapper of the elements) and then `flex-grow: 1` for the input area, but I didn't test this, so let me know if you need help.

And it is important to do CSS changes in the LESS files and then run `grunt build-theme` as I wrote above.
Probably we should add the changes in the file [/static/css/select2Theme/select2-humhub.less](https://github.com/humhub/humhub/blob/master/static/css/select2Theme/select2-humhub.less) or some other from the folder https://github.com/humhub/humhub/tree/master/static/less.

luke-:
@yurabakhtin Thanks!
@gevorgmansuryan Let me know when there is an update here

gevorgmansuryan:
@yurabakhtin flex did not helped :(  @luke- I've managed to fix it via js, but autofocus is not working. What do you think is it worth to research more?

gevorgmansuryan:
@luke- initially there was 2 issues, invalid click zone and and non working autofocus when opening modal. now click zone works

luke-:
@gevorgmansuryan Ok, should we merge this and create a new PR when we find a solution for the autofocus?

gevorgmansuryan:
@luke- yes we can merge this for now. Then we can find solution for autofocus if needed

luke-:
@yurabakhtin Can you please review? This PR should fixes the click area issue.

I'll create a new issue for the autofocus after click. https://github.com/humhub/humhub-internal/issues/458

PR #7322, opened 2024-11-28
Always allow invitation by link

yurabakhtin:

- [x] Bugfix

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
https://github.com/humhub/humhub-internal/issues/460

### DIFF
--- a/protected/humhub/modules/user/services/LinkRegistrationService.php
+++ b/protected/humhub/modules/user/services/LinkRegistrationService.php
@@ -70,12 +70,13 @@ final class LinkRegistrationService
 
     public function isValid(): bool
     {
-        return $this->getStoredToken() === $this->token;
+        return $this->isEnabled() && $this->getStoredToken() === $this->token;
     }
 
     public function isEnabled(): bool
     {
         if ($this->target === self::TARGET_ADMIN) {
+            // The link registration with token from Administration is always enabled
             return true;
         }
 
--- a/protected/humhub/modules/user/services/LinkRegistrationService.php
+++ b/protected/humhub/modules/user/services/LinkRegistrationService.php
@@ -8,6 +8,7 @@
 
 namespace humhub\modules\user\services;
 
+use humhub\components\SettingsManager;
 use humhub\modules\space\models\Space;
 use humhub\modules\user\models\Invite;
 use humhub\modules\user\models\User;
@@ -24,10 +25,17 @@ use yii\db\StaleObjectException;
  */
 final class LinkRegistrationService
 {
+    public const SETTING_VAR_ENABLED = 'auth.internalUsersCanInviteByLink';
     public const SETTING_VAR_SPACE_TOKEN = 'inviteToken';
-    public const SETTING_VAR_TOKEN = 'registration.inviteToken';
+    public const SETTING_VAR_ADMIN_TOKEN = 'registration.inviteToken';
+    public const SETTING_VAR_PEOPLE_TOKEN = 'people.inviteToken';
+
+    public const TARGET_ADMIN = 'admin';
+    public const TARGET_PEOPLE = 'people';
+
     private ?Space $space;
     private ?string $token;
+    public ?string $target = null;
 
     public static function createFromRequest(): LinkRegistrationService
     {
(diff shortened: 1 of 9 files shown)


protected/humhub/modules/user/services/LinkRegistrationService.php
```diff
@@ -57,10 +56,10 @@ public function isValid(): bool
 
     public function isEnabled(): bool
```
luke-, line 57: Can't we remove the method then?

luke- (changes requested):
@yurabakhtin Thanks. Please see my remark.

luke- (changes requested):
@yurabakhtin  Would it now still be possible to retrospectively deactivate link invitations, e.g. also via spaces?
e.g. the following scenario.

- Link registration is allowed
- In a space, this is used to an extreme extent. 
- Super Admin wants to deactivate it and ends link registration
---
Would registration with the space registration links still be possible?

yurabakhtin:
@luke- I have tested the Space case.

The feature "Invite by link" is visible on Space only when the general setting "Members can invite external users by link" is enabled:

space-invite

However if I disable the general setting I see the registration is still allowed if user has a link with a correct token.

Admin can reset token to new one by click "Create new link":
reset-token

We could reset ALL tokens(space and general) on each case when the setting "Members can invite external users by link" is updated from "enalbed" to "disabled". But it may be unclear for users because they will see the message:

invalid-token

So it seems we should revert the method `isEnabled()` but use it only for Space case, right?

marc-farre:
@yurabakhtin @luke- I would say:
- no automatic token reset (because we might want to disable the feature for a day and enable it back later, the same links should still be working)
- When "Members can invite external users by link", the full feature is disabled (event previous links are not working anymore)
- Exception: System admins can still invite globally when the feature is disabled

Same behaviour with Email invitations.

But we have an issue with this: the invite link in the People page is the same as the one in the global admin page (for sys-admins).
So it is currently not possible to allow sys-admin links and not links from the People page:

image

image

So I think that ideally, the links should be different...

luke-:
@marc-farre Thanks, that sounds good to me. 

@yurabakhtin Can we implement it like that? Ideally +tests :-)

yurabakhtin:
@marc-farre To be sure what is required I would like to describe how it works now and what I have understood to implement:

Now:

1. We have 2 link token types: Global and Space. The Global is used from Administration and People pages. The Space token is generated per each space.
2. Registration is not blocked from the Global link now even when the setting "Members can invite external users by link" is disabled because Admins always have an access there.

New requirement:
1. Create a new link type: People.
2. Such link should have a different token than Global. I think to add a new param like `target=people` => `/user/registration/by-link?token=ygL9JR8erHfJlq&target=people`
3. Registration should be blocked from such link when "Members can invite external users by link" is disabled.
4. Registration from Global(Admin) link will never be blocked.

marc-farre:
@yurabakhtin Yes, it's exactly that, except that Now:
- Registration IS BLOCKED from the Global link now when the setting "Members can invite external users by link" is disabled despite that the Admins should always have an access there.

yurabakhtin:
> * Registration IS BLOCKED from the Global link now when the setting "Members can invite external users by link" is disabled despite that the Admins should always have an access there.

@marc-farre Sorry, I am confused in this point.
Do you want to block a registration when the setting "Members can invite external users by link" even that Admins have an access to generate such global links?
I thought it was a bug and we should make all global links enabled even when the setting is disabled, because Admins can generate such links and they may don't know the setting is disabled currently, so it will be strange if admin sends such global link to user and he will see error page instead of expected registration form.

Or when you wrote "Now" you mean how it works now without new changes from this PR.
When I wrote "Now" above I meant how it works with new changes from this PR.

marc-farre:
> Or when you wrote "Now" you mean how it works now without new changes from this PR.

I see. I meant currently (before PR). I miss understood "Now". 
Si everything is fine with your summary.

yurabakhtin:
@luke- @marc-farre Done, how it works now in this PR:

- New separate token has been inmplemented for the People page, it is stored as general setting `people.inviteToken`,
- Previous token is stored in the same setting `registration.inviteToken`, it is used for Administration part.
- I didn't add a new param like `target=people` as I wrote above, instead I have decided to detect a target by token value, so the links are different only by token value:
diff_tokens

When the setting is **enabled**:
link-enabled

When **disabled**:
link-disabled

marc-farre:
@yurabakhtin That's really perfect, thanks!

luke- (approved):

luke-:
@yurabakhtin @marc-farre Thanks, looks fine for me too!

Issue #5458, opened 2021-12-03
Reduce `dynamic.php` content


luke-:
The configuration file `dynamic.php` should ideally contain only the database access data. Everything else should be read from the database & cache.

aronbos:
Please think about an option to add multiple databases. I am using two extra database in my own modules that have no direct connection to the Humhub main database. In dynamic.php adding the databases is simple. Just add the name en parameters to the db-array.

luke-:
@gevorgmansuryan Ready to review?

gevorgmansuryan:
@luke- yes

gevorgmansuryan:
@luke- tests are fixed

PR #7328, opened 2024-12-04
reduce dynamic.php

gevorgmansuryan:
https://github.com/humhub/humhub/issues/5458


luke-:
Can you please update the `develop` branch, to get rid of the many PHPFixer changes?

protected/humhub/libs/DynamicConfig.php
```diff
@@ -91,37 +90,10 @@ public static function rewrite()
...
-        if (Yii::$app->getModule('user')->settings->get('auth.defaultUserIdleTimeoutSec')) {
-            $config['components']['user']['authTimeout'] = Yii::$app->getModule('user')->settings->get('auth.defaultUserIdleTimeoutSec');
-        }
 
-        // Install Mail Component
-        $config['components']['mailer'] = self::getMailerConfig();
 
         // Remove old theme/view stuff
         unset($config['components']['view']);
-        unset($config['components']['mailer']['view']);
```
luke-, line 124: We should `unset` all old `dynamic.php` vars, otherwise it would not be removed on the next save.

protected/humhub/components/bootstrap/SettingsLoader.php
```diff
@@ -0,0 +1,94 @@
...
+        if (!$app) {
+            return;
+        }
+
+        $this->setMailerConfig($app);
+        $this->setUserConfig($app);
+        $this->setCacheConfig($app);
+    }
+
+    protected function setMailerConfig($app): void
```
luke-, line 22: Please mark such methods as 'private'. No inheritance is planned here.

luke- (changes requested):
@gevorgmansuryan Thanks for the PR.

I like the approach with the 'SettingsLoader'.

- What happens if a 'Mailer' or 'Cache', for example, is still specified in the configuration file? Would this still be possible?

- I think that now the `mailer` (and `user`) component seems to be always instantiated. Previously, this was only the case when a mail was sent. Do you see a change to avoid this?

protected/humhub/libs/DynamicConfig.php
gevorgmansuryan: this was done in second PR

gevorgmansuryan:
@luke- 

> What happens if a 'Mailer' or 'Cache', for example, is still specified in the configuration file? Would this still be possible?
This was done in second OR

> I think that now the mailer (and user) component seems to be always instantiated. Previously, this was only the case when a mail was sent. Do you see a change to avoid this?
Right, should be fixed now

luke-:
@gevorgmansuryan Thanks for the adjustments.

Somehow, I'm not so happy with the code, it looks duplicated and complex, since it handles the instanciated and config state.

Any idea how to improve this?

Maybe we could here adjust the config with default values, if not provided?
https://github.com/humhub/humhub/blob/develop/protected/humhub/components/ApplicationTrait.php#L46

gevorgmansuryan:
@luke- we can just remove duplications and just use app->set as all components are not instanciated at this step.

luke-:
> @luke- we can just remove duplications and just use app->set as all components are not instanciated at this step.

@gevorgmansuryan That would be perfect. Perhaps we can log a warning if a component has already been initialized.

protected/humhub/components/bootstrap/SettingsLoader.php
```diff
@@ -29,19 +30,17 @@ private function updateComponentDefinition($app, $component, $definition)
 
     private function setMailerConfig($app): void
     {
-        $instantiated = $app->has('mailer', true);
+        if ($app->has('mailer', true)) {
+            $app->log->logger->log('`mailer` component should not be instantiated before settings are loaded.', Logger::LEVEL_WARNING);
```
luke-, line 44: Should we return here?

luke- (changes requested):
@gevorgmansuryan Thanks, looks good. one small thing:

luke- (approved):

PR #7332, opened 2024-12-05
Reduce dynamic.php content params

gevorgmansuryan:
https://github.com/humhub/humhub/issues/5458


protected/humhub/libs/DynamicConfig.php
```diff
@@ -19,17 +19,6 @@
  */
 class DynamicConfig extends BaseObject
```
luke-, line 20: @gevorgmansuryan We should probably also rename the file to e.g. “DatabaseCredConfig” in a next PR and further simplify it.

luke-: not sure we need `load()` here. Maybe just a `saveCurrent()`.

luke- (approved):

PR #7344, opened 2024-12-13
disable fixed baseurl

gevorgmansuryan:
https://github.com/humhub/humhub-internal/issues/475

### DIFF
--- a/protected/humhub/modules/admin/models/forms/BasicSettingsForm.php
+++ b/protected/humhub/modules/admin/models/forms/BasicSettingsForm.php
@@ -102,9 +102,7 @@ class BasicSettingsForm extends Model
     public function save()
     {
         Yii::$app->settings->set('name', $this->name);
-        if (!$this->isBaseUrlFixed()) {
-            Yii::$app->settings->set('baseUrl', $this->baseUrl);
-        }
+        Yii::$app->settings->set('baseUrl', $this->baseUrl);
         Yii::$app->settings->set('defaultLanguage', $this->defaultLanguage);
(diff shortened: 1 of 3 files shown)


protected/humhub/modules/admin/views/setting/basic.php
```diff
@@ -19,7 +19,7 @@
     <?php $form = ActiveForm::begin(['acknowledge' => true]); ?>
 
     <?= $form->field($model, 'name') ?>
-    <?= $form->field($model, 'baseUrl') ?>
+    <?= $form->field($model, 'baseUrl')->textInput(['disabled' => $model->isBaseUrlFixed()]) ?>
```
luke-, line 22: Better use the same style as here below:

```php
<?= $form->field($model, 'defaultTimeZone')->dropDownList(TimezoneHelper::generateList(true), ['disabled' => Yii::$app->settings->isFixed('defaultTimeZone')]) ?>
```

protected/humhub/modules/admin/models/forms/BasicSettingsForm.php
```diff
@@ -104,7 +102,9 @@ public function attributeHints()
     public function save()
     {
         Yii::$app->settings->set('name', $this->name);
-        Yii::$app->settings->set('baseUrl', $this->baseUrl);
+        if (!$this->isBaseUrlFixed()) {
```
luke-, line 105: We don't need this condition. Like for `defaultTimeZone` which may be also fixed, the settings manager always returns the fixed value.

luke- (changes requested):

luke- (approved):

PR #7346, opened 2024-12-14
Rename cahce config keys

gevorgmansuryan:
https://github.com/humhub/humhub-internal/issues/479


protected/humhub/migrations/m160508_005740_settings_cleanup.php
```diff
@@ -49,8 +49,8 @@ public function up()
         $this->update('setting', ['name' => 'auth.ldap.emailAttribute', 'module_id' => 'user'], ['module_id' => 'authentication_ldap', 'name' => 'emailAttribute']);
 
         // Fix cache settings module id
-        $this->update('setting', ['name' => 'cache.class', 'module_id' => 'base'], ['module_id' => 'cache', 'name' => 'type']);
-        $this->update('setting', ['name' => 'cache.expireTime', 'module_id' => 'base'], ['module_id' => 'cache', 'name' => 'expireTime']);
+        $this->update('setting', ['name' => 'cacheClass', 'module_id' => 'base'], ['module_id' => 'cache', 'name' => 'type']);
```
luke-, line 52: Not sure we should change it here. This only affects older DBs.

luke- (changes requested):

protected/humhub/migrations/m160508_005740_settings_cleanup.php
gevorgmansuryan: @luke- but it will run when installing fresh humhub. Undo?

luke-: @gevorgmansuryan For new installations, the correct value should already be set via “InitialData” here: https://github.com/humhub/humhub/pull/7346/files#diff-ad3c9b4512c98bc7307be8b2d80a598fe0b59587f3948400819a827a0b408a5dR66

Only existing installations should be affected by the migration.

So I would undo this migration change.

luke- (approved):

PR #7347, opened 2024-12-16
Check Codeceptions Tests

gevorgmansuryan:
https://github.com/humhub/humhub-internal/issues/473


gevorgmansuryan:
There was issue related application installed state in test environment. Still in progress

gevorgmansuryan:
The problem is that application installed state now works based on database, but before starting tests the database is being cleared. Still in progress

luke-:
@gevorgmansuryan Thanks for the update.

gevorgmansuryan:
Will be fixed by https://github.com/humhub/humhub/pull/7361, as it's related with applicationState

PR #7349, opened 2024-12-16
Enh: Add body classes about the current device and methods to the Dev…

marc-farre:
…iceDetectorHelper class

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes (some JS fixes for mobile won't work correctly on themes not adding the body classes)
- [ ] No

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**

Issue https://github.com/humhub/humhub-internal/issues/472


protected/humhub/views/layouts/main.php
```diff
@@ -25,7 +26,7 @@
         <?= $this->render('head') ?>
     </head>
 
-    <body>
+    <body class="<?= implode(' ', DeviceDetectorHelper::getBodyClasses()) ?>">
```
luke-, line 29: Small point: maybe we should return a string in `getBodyClasses()` or create another method.
This would keep the View cleaner?

protected/humhub/widgets/views/richTextEditor.php
```diff
@@ -98,7 +99,7 @@
         $editableContent = $('#<?php echo $id; ?>_contenteditable');
 
         //This is a workaround for mobile browsers especially for Android Chrome which is not able to remove contenteditable="false" nodes.
-        if (/Android|webOS|iPhone|iPad|iPod|BlackBerry|BB|PlayBook|IEMobile|Windows Phone|Kindle|Silk|Opera Mini/i.test(navigator.userAgent)) {
+        if ($('body').hasClass('device-mobile')) {
```
luke-, line 102: Also device-tablet?

luke- (changes requested):

protected/humhub/views/layouts/main.php
marc-farre: I had the same thought. But I like to keep array, because I think there are less bugs when we want to add/remove classes from the result, e.g. in a custom theme, but mainly in general, I think it's a better practice.

Perhaps later we will want to add an event in this method too.

So 2 options:
- Adding the method `getBodyClassesAsString()`
- Using `<?= Html::beginTag('body', ['class' => DeviceDetectorHelper::getBodyClasses()]) ?>`
- 
What do you prefer?

luke-: Hmm I kike the second option. What do you think?

marc-farre: Yes. I also prefer to avoid too many methods. Done in latest commit.

PR #7360, opened 2024-12-19
FormModal example

luke-:


luke-:
@marc-farre A quick idea. What do you think?
 
We may could also add the "[begin|end]FormModal" methods directly to `humhub\widgets\modals\Modal`.

marc-farre:
@luke- Thanks!
Yes, that will help.
But as the new widget only has 2 methods, I think I we could avoid adding a new one (we already have lots of widgets) and add these 2 methods directly into `humhub\widgets\modals\Modal` as you also suggest.
I think it also makes things clearer, because without reading the code, we don't know if `FormModal` extends `Modal` or `ActiveForm` widget.

luke-:
@marc-farre Good point.

I've adjusted the PR. Should we want to introduce this or stay with the current way?

marc-farre:
Thanks! Yes, I think it's better to use these 2 new methods.
I will take care of it to finish migrating all existing Modals with forms, and updating the doc.
I will then merge the PR.

PR #7368, opened 2025-01-03
Rename dynamic.php and DynamicConfig.php

gevorgmansuryan:
Rename dynamic.php and DynamicConfig.php


luke-:
@gevorgmansuryan Thanks for the separate PR. I would like to put it on hold for a bit and schedule it again later in v18 or v19.

Issue #7134, opened 2024-07-22
Choice between Instant or Background Jobs


ArchBlood:
### Is your feature request related to a problem? Please describe.
It would be nice not just having the module disabling and uninstalling in the background, as currently this is how the process goes, but it should be configurable at the very least, the old way was instant and cleared the cache correctly, having them as background jobs depend on the Cron Jobs which do not clear the module cache after disabling or uninstalling and force using ACP > Settings > Advanced > Caching.

### Describe the solution you'd like
I'd like it so that the user could have either the background job or an instant method for disabling & uninstalling a module.

### Describe alternatives you've considered
Revert away from using background jobs or properly handle cache for jobs.

https://community.humhub.com/content/perma?id=279751

ArchBlood:
I say this because some server providers use a cache management system that could prevent HumHub from clearing cache in the background jobs, one system that I can note that does this is Cachewall which normally the user does not have permissions to manage.

luke-:
You are right, the APC cache cannot always be flushed outside of the WebServer process (e.g. Cron). I would therefore remove this option soon. At least in the ACP. Files/Redis can be reliable flushed.

---

In the current case, the user seems to have a problem with the queue processing/cron jobs. It is important that the user fixes this problem instead of us introducing an option to bypass it for that case.

Robkne:
After a lot of testing the problem with the queue processing cron jobs is resolved.

Whenever the job is running I get this error message:

yii\base\ErrorException: Undefined array key "SCRIPT_FILENAME" in /home2/organxom/sandbox.8045.mx/protected/vendor/yiisoft/yii2-queue/src/cli/Command.php:168
Stack trace:
#0 /home2/organxom/sandbox.8045.mx/protected/vendor/yiisoft/yii2-queue/src/cli/Command.php(168): yii\base\ErrorHandler->handleError(2, 'Undefined array...', '/home2/organxom...', 168)
#1 /home2/organxom/sandbox.8045.mx/protected/vendor/yiisoft/yii2-queue/src/cli/Command.php(126): yii\queue\cli\Command->handleMessage(3, 'O:48:"humhub\\mo...', 3600, 1)
#2 [internal function]: yii\queue\cli\Command->yii\queue\cli\{closure}(3, 'O:48:"humhub\\mo...', 3600, 1)
[...]

PR #7413, opened 2025-01-27
Fix: APCu Cache Compatibility in IsOnlineService

ArchBlood:

- [x] Bugfix
- [ ] Feature
- [x] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] All tests are passing
- [x] Changelog was modified

**Other information:**
This P/R should help fix some of the issues like this one;
https://community.humhub.com/content/perma?id=283871

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ HumHub Changelog
 - Fix #7400: Fixed `Default user profile visibility` field visibility in the user settings
 - Fix #7404: Marketplace - Allow symlinked `@app/modules` directory
 - Fix: Always allow admins to edit/delete content in the other Profile stream
+- Fix: APCu Cache Compatibility in IsOnlineService
 
 1.17.0 (January 13, 2025)
 -------------------------
--- a/protected/humhub/modules/user/services/IsOnlineService.php
+++ b/protected/humhub/modules/user/services/IsOnlineService.php
@@ -30,16 +30,19 @@ class IsOnlineService
 
     public function updateStatus(): void
     {
-        if ($this->isEnabled() && !Yii::$app->cache->exists($this->getCacheKey())) {
+        if ($this->isEnabled() && $this->getStatus() === false) {
             Yii::$app->cache->set($this->getCacheKey(), true, 60); // Expires in 60 seconds
         }
     }
 
     public function getStatus(): bool
     {
-        return
-            $this->isEnabled()
-            && Yii::$app->cache->exists($this->getCacheKey());
+        if (!$this->isEnabled()) {
+            return false;
+        }
+
+        $cacheValue = Yii::$app->cache->get($this->getCacheKey());
+        return $cacheValue !== false; // Return true if key exists and is not `false`
     }
 
     public function isEnabled(): bool


marc-farre:
@ArchBlood Thanks for your PR.
So it means that replacing `Yii::$app->cache->exists()` with `Yii::$app->cache->get()` fixes the issue with APCu cache?
If yes, it means that with APCu the cached value don't expire after 60 seconds, despite the parameter `Yii::$app->cache->set($this->getCacheKey(), true, 60)`?
Have you tested successfully this PR, or it's a non-tested suggestion?

ArchBlood:
> @ArchBlood Thanks for your PR. So it means that replacing `Yii::$app->cache->exists()` with `Yii::$app->cache->get()` fixes the issue with APCu cache? If yes, it means that with APCu the cached value don't expire after 60 seconds, despite the parameter `Yii::$app->cache->set($this->getCacheKey(), true, 60)`? Have you tested successfully this PR, or it's a non-tested suggestion?

I've tested and it works as it should with no errors for File or APCu caching enabled;

### 1 Online User
Screenshot_11

### More Than 1 Online User
Screenshot_12

marc-farre:
@ArchBlood OK, thanks. But on my side, the current code also works with APCu.
Do you mean that on your side, before this PR, the online indicator doesn't work, and after, it works?
Because I agree that `Yii::$app->cache->get()` would work as good, but it takes more CPU resources than `Yii::$app->cache->exists()`, and I don't understand why it would work better.

ArchBlood:
> @ArchBlood OK, thanks. But on my side, the current code also works with APCu. Do you mean that on your side, before this PR, the online indicator doesn't work, and after, it works? Because I agree that `Yii::$app->cache->get()` would work as good, but it takes more CPU resources than `Yii::$app->cache->exists()`, and I don't understand why it would work better.

It doesn't work correctly when cache is flushed, same as for when APCu is used for console commands for flushing cache, this is because `exists()` works differently.

marc-farre:
@ArchBlood these are the 2 methods:

```php
    public function get($key)
    {
        $key = $this->buildKey($key);
        $value = $this->getValue($key);
        if ($value === false || $this->serializer === false) {
            return $value;
        } elseif ($this->serializer === null) {
            $value = unserialize((string)$value);
        } else {
            $value = call_user_func($this->serializer[1], $value);
        }
        if (is_array($value) && !($value[1] instanceof Dependency && $value[1]->isChanged($this))) {
            return $value[0];
        }

        return false;
    }

    /**
     * Checks whether a specified key exists in the cache.
     * This can be faster than getting the value from the cache if the data is big.
     * In case a cache does not support this feature natively, this method will try to simulate it
     * but has no performance improvement over getting it.
     * Note that this method does not check whether the dependency associated
     * with the cached data, if there is any, has changed. So a call to [[get]]
     * may return false while exists returns true.
     * @param mixed $key a key identifying the cached value. This can be a simple string or
     * a complex data structure consisting of factors representing the key.
     * @return bool true if a value exists in cache, false if the value is not in the cache or expired.
     */
    public function exists($key)
    {
        $key = $this->buildKey($key);
        $value = $this->getValue($key);

        return $value !== false;
    }
```

As you can see, `exists()` is the same as `get()`, except that `get()` unserialize the value and returns an array.

But your PR compares the value returned by get to `false`, exactly what `exists()` already does: https://github.com/humhub/humhub/pull/7413/files#diff-4480fb0474cadf31b48950c2fe74ee0163eba5f2f34bc83607f81c7c0a4ded85R45

> It doesn't work correctly when cache is flushed

If cache is flushed, of course, it needs a minute to have all online users again.
But why would `get()` work better?

ArchBlood:
> @ArchBlood these are the 2 methods:
> 
> ```
>     public function get($key)
>     {
>         $key = $this->buildKey($key);
>         $value = $this->getValue($key);
>         if ($value === false || $this->serializer === false) {
>             return $value;
>         } elseif ($this->serializer === null) {
>             $value = unserialize((string)$value);
>         } else {
>             $value = call_user_func($this->serializer[1], $value);
>         }
>         if (is_array($value) && !($value[1] instanceof Dependency && $value[1]->isChanged($this))) {
>             return $value[0];
>         }
> 
>         return false;
>     }
> 
>     /**
>      * Checks whether a specified key exists in the cache.
>      * This can be faster than getting the value from the cache if the data is big.
>      * In case a cache does not support this feature natively, this method will try to simulate it
>      * but has no performance improvement over getting it.
>      * Note that this method does not check whether the dependency associated
>      * with the cached data, if there is any, has changed. So a call to [[get]]
>      * may return false while exists returns true.
>      * @param mixed $key a key identifying the cached value. This can be a simple string or
>      * a complex data structure consisting of factors representing the key.
>      * @return bool true if a value exists in cache, false if the value is not in the cache or expired.
>      */
>     public function exists($key)
>     {
>         $key = $this->buildKey($key);
>         $value = $this->getValue($key);
> 
>         return $value !== false;
>     }
> ```
> 
> As you can see, `exists()` is the same as `get()`, except that `get()` unserialize the value and returns an array.
> 
> But your PR compares the value returned by get to `false`, exactly what `exists()` already does: https://github.com/humhub/humhub/pull/7413/files#diff-4480fb0474cadf31b48950c2fe74ee0163eba5f2f34bc83607f81c7c0a4ded85R45
> 
> > It doesn't work correctly when cache is flushed
> 
> If cache is flushed, of course, it needs a minute to have all online users again. But why would `get()` work better?

Using `get()` instead of `exists()` ensures deserialization and dependency validation, which `exists()` skips. If the cache value is corrupted, invalid, or affected by a dependency, `get()` will correctly return `false`, while `exists()` might incorrectly return `true`. This makes `get()` more robust in situations where cached data integrity matters, even though both return `false` when the cache is flushed.

marc-farre:
I'm sorry, but I don't understand why exists() may incorrectly return true.
And if it was the case, there would be more users having the online status green, but the issue here is reverse: https://community.humhub.com/s/general-discussion/post/post/view?id=17192

Here, we just want to know if something has been set in the cache for a user, do matter what it is. So unserialization is useless here.
It expires after 60 seconds, which means the cached value doesn't exist any more (serialized or not), which is what we want to test.

This value is set each time the browser makes a request via JS, every 15/20 seconds, via the live module (`/live/poll` URL).
Maybe the issue is that sometimes the poll request is not properly sent by the browser (network issue, browser extension that blocks it, JS issue, etc).

ArchBlood:
> I'm sorry, but I don't understand why exists() may incorrectly return true. And if it was the case, there would be more users having the online status green, but the issue here is reverse: https://community.humhub.com/s/general-discussion/post/post/view?id=17192
> 
> Here, we just want to know if something has been set in the cache for a user, do matter what it is. So unserialization is useless here. It expires after 60 seconds, which means the cached value doesn't exist any more (serialized or not), which is what we want to test.
> 
> This value is set each time the browser makes a request via JS, every 15/20 seconds, via the live module (`/live/poll` URL). Maybe the issue is that sometimes the poll request is not properly sent by the browser (network issue, browser extension that blocks it, JS issue, etc).

That is mainly the issue when it comes to using APCu cache, sometimes it does take a few minutes other times it just doesn't process at all and a manual flush is required by webserver which one case in point is #7134.

marc-farre:
@ArchBlood Thanks for the info. So the issue seams to be APCu, but I cannot see how this PR can solve it.

@luke- What do you think?
Perhaps we should remove APCu caching support (see https://github.com/humhub/humhub-internal/issues/262)?

ArchBlood:
> @ArchBlood Thanks for the info. So the issue seams to be APCu, but I cannot see how this PR can solve it.
> 
> @luke- What do you think? Perhaps we should remove APCu caching support (see https://github.com/humhub/humhub-internal/issues/262)?

To be honest, you are correct, this does not fix all of the issues when it comes to APCu cache, I'd like to think of it more along the lines of starting actual support for ACPu, it is on the other hand slightly more difficult when it comes to CLI though which is why the issue was reported in the first place and even mentioned here.

ArchBlood:
> Hmm, I just briefly looked over the issue, but if `cache->exists()` isn't working correctly, we should rather fix this bug in Yii2?

That I am not completely sure of myself, as the Yii2 team may have done this intentionally at the time when it was needed.

> Besides, we haven't planned to officially support APC for now anyway, so I would avoid workarounds like this.

If workarounds like this aren't wanted then dropping support, even if not officially supported, should be done. As it stands allowing APC(u) unofficially still harms the platform due to this cache issue when it is enabled, and is even more noticeable when it comes to CLI and background jobs.

Issue #7414, opened 2025-01-29
1.17.0 User Title is Mangled


miqrogroove:
After updating from v1.16.2 to v1.17.0 the encoding of special characters in the header bar is incorrect.

My user title contains a simple apostrophe `'` and still appears correctly in the form at `/user/account/edit`.  However, in the header bar below my name, the apostrophe now displays as `&#039;` and the source encoding has changed to `&amp;#039;`

The same bug appears in the user list at `/admin/user/list` as well as in the spaces list at `/admin/space/index` and in the title on my user profile page.

### Additional info

| Q                | A
| ---------------- | ---
| HumHub version   | 1.17.0
| PHP version      | 8.3
| Operating system | Ubuntu

ArchBlood:
@luke- I believe the issue is because it is being encoded twice somewhere, once because of #6955 and then again in accountTopMenu view and maybe elsewhere, what's weird is this issue only happens for `$userModel->displayNameSub` and not for `$userModel->displayName`.

https://github.com/humhub/humhub/blob/e6d72bd182d8cd654abb30d673a78f0a404b6ffb/protected/humhub/modules/user/widgets/views/accountTopMenu.php#L40-L41

miqrogroove:
Just a thought... First name and displayName are different values, so need to dig deeper.

https://github.com/humhub/humhub/blob/e6d72bd182d8cd654abb30d673a78f0a404b6ffb/protected/humhub/modules/user/models/User.php#L674

https://github.com/humhub/humhub/blob/e6d72bd182d8cd654abb30d673a78f0a404b6ffb/protected/humhub/modules/user/models/User.php#L702

https://github.com/humhub/humhub/blob/e6d72bd182d8cd654abb30d673a78f0a404b6ffb/protected/humhub/modules/user/models/Profile.php#L426

https://github.com/humhub/humhub/blob/e6d72bd182d8cd654abb30d673a78f0a404b6ffb/protected/humhub/modules/user/models/fieldtype/Text.php#L217

Voila.  There's the second call to `encode`.

ArchBlood:
> Just a thought... First name and displayName are different values, so need to dig deeper.
> 
> displayName:
> 
> [humhub/protected/humhub/modules/user/models/User.php](https://github.com/humhub/humhub/blob/e6d72bd182d8cd654abb30d673a78f0a404b6ffb/protected/humhub/modules/user/models/User.php#L674)
> 
[...]

PR #7420, opened 2025-02-06
Fix profile field encoding

yurabakhtin:

- [x] Bugfix

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
https://github.com/humhub/humhub/issues/7414

### DIFF
--- a/protected/humhub/modules/user/models/Profile.php
+++ b/protected/humhub/modules/user/models/Profile.php
@@ -413,9 +413,10 @@ class Profile extends ActiveRecord
      *
      * @param string $field
      * @param bool $raw
+     * @param bool $encode
      * @return string|null
      */
-    public function getFieldValue(string $field, bool $raw = false): ?string
+    public function getFieldValue(string $field, bool $raw = false, bool $encode = false): ?string
     {
         if (!$this->hasAttribute($field) || !$this->user) {
             return null;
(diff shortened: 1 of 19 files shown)


protected/humhub/modules/user/models/Profile.php
```diff
@@ -413,16 +413,17 @@ public function softDelete()
      *
      * @param string $field
      * @param bool $raw
+     * @param bool $encode
      * @return string|null
      */
-    public function getFieldValue(string $field, bool $raw = false): ?string
+    public function getFieldValue(string $field, bool $raw = false, bool $encode = false): ?string
```
luke-, line 419: @yurabakhtin Wouldn't it be better to also use `true` as the default value here? If all returns were previously encoded?

luke- (changes requested):

protected/humhub/modules/user/models/Profile.php
yurabakhtin: @luke- I find the method is used only in the single place (inside method `User->getDisplayNameSub()`), and there we need `$encode = false` to fix the issue.
We can set the default value to `true` but then we need to change [the code](https://github.com/humhub/humhub/blob/fix/7414-profile-field-encoding/protected/humhub/modules/user/models/User.php#L702) to:

```php
return $this->profile->getFieldValue(Yii::$app->settings->get('displayNameSubFormat', ''), false, false) ?? '';
```

Do you prefer this?

yurabakhtin: @luke- done https://github.com/humhub/humhub/pull/7420/commits/1732d88896268445d7822299283f3d8e10f69b45.

luke- (approved):

PR #7438, opened 2025-02-23
Profile Topic Issue

gevorgmansuryan:
https://github.com/humhub/humhub-internal/issues/527


protected/humhub/modules/content/widgets/ContentTagPicker.php
```diff
@@ -53,7 +53,10 @@ protected function findDefaults()
     {
         /* @var ContentTagActiveQuery $query */
         $query = call_user_func([$this->itemClass, 'findByContainer'], $this->contentContainer, true);
-        $query->readable()->limit($this->limit);
+        $query->limit($this->limit);
+        if (!Yii::$app->user->identity->canManageAllContent()) {
+            $query->readable();
```
luke-, line 58: @gevorgmansuryan Hmm, it would probably be good to solve this within the `readable()` method. Or is there something against it here?

protected/humhub/modules/topic/widgets/TopicPicker.php
```diff
@@ -97,11 +97,7 @@ private static function canAddTopic(ContentContainerActiveRecord $container = nu
      */
     private static function hasTopics(ContentContainerActiveRecord $container = null)
     {
-        if (!$container) {
-            return (bool)Topic::find()->count();
-        }
-
-        return (bool)Topic::findByContainer($container)->count();
+        return ($container ? Topic::findByContainer($container)->count() : 0) + Topic::find()->count();
```
luke-, line 100: Can we perhaps rename the method to count Topics and fix the PHPDoc if we return the count here and not a bool?

luke- (changes requested):

protected/humhub/modules/topic/widgets/TopicPicker.php
gevorgmansuryan: @luke- better bool

protected/humhub/modules/topic/Module.php
```diff
@@ -9,10 +9,9 @@
 
 namespace humhub\modules\topic;
 
-use humhub\modules\topic\permissions\ManageTopics;
-use Yii;
-use humhub\modules\topic\permissions\AddTopic;
 use humhub\modules\space\models\Space;
+use humhub\modules\topic\permissions\AddTopic;
+use humhub\modules\topic\permissions\ManageTopics;
```
luke-, line 14: Why do we need this imports?

luke- (changes requested):

protected/humhub/modules/topic/Module.php
gevorgmansuryan: @luke- there is no new import, just removed unused `use Yii;` and it's just my IDE fixed order of imports.

luke-: @gevorgmansuryan Oh thanks, missed that!

Issue #7388, opened 2025-01-16
Email "Reset Password" link is wrong when pretty URL is disabled and HumHub not in the URL root


marc-farre:
### What steps will reproduce the problem?

1. Install a HumHub instance in `https://domain.tld/folder`
2. Disable Pretty URL
3. Recover password by email
4. Click on the "Reset Password" button in the email:

![image](https://github.com/user-attachments/assets/36d41659-e55a-4d96-8b19-c5e125e1f765)

### What is the expected result?

The URL is `https://domain.tld/folder/index.php?...`

### What do you get instead?

The URL is `https://domain.tld/folder/folder/index.php?...`

### Additional info

Original post: https://community.humhub.com/s/french/post/post/view?id=17160

| Q                | A
| ---------------- | ---
| HumHub version   | 1.17.0

luke-:
@gevorgmansuryan Please see my review

gevorgmansuryan:
@luke – The initial problem was that SafeBaseUrl was being called from a web application, but it was internally switching the urlManager to the console UrlManager. As a result, the console UrlManager was being used within the web app, and UrlManager::createUrl() did not work correctly.

I’ve fixed it so that the web UrlManager is used in SafeBaseUrl when running in a web application context. I also added an exception in the console UrlManager to prevent it from being used in a web application.

luke-:
@marc-farre If you have time, can you please confirm it's fixed now?

marc-farre:
@luke- @gevorgmansuryan I confirm the issue is fixed. Thanks!

PR #7448, opened 2025-02-28
Email "Reset Password" link is wrong when pretty URL is disabled and HumHub not in the URL root

gevorgmansuryan:
https://github.com/humhub/humhub/issues/7388

### DIFF
--- a/protected/humhub/components/console/UrlManager.php
+++ b/protected/humhub/components/console/UrlManager.php
@@ -29,7 +29,7 @@ class UrlManager extends \humhub\components\UrlManager
             $hostInfo .= ':' . $urlParts['port'];
         }
         $this->setHostInfo($hostInfo);
-        $this->setScriptUrl($this->getBaseUrl() . ($this->getScriptUrl() ?: '/index.php'));
+        $this->setScriptUrl($this->getBaseUrl() . (str_replace($this->getBaseUrl(), '', $this->getScriptUrl()) ?: '/index.php'));
 
         parent::init();
     }


protected/humhub/components/console/UrlManager.php
```diff
@@ -29,7 +29,7 @@ public function init()
             $hostInfo .= ':' . $urlParts['port'];
         }
         $this->setHostInfo($hostInfo);
-        $this->setScriptUrl($this->getBaseUrl() . ($this->getScriptUrl() ?: '/index.php'));
+        $this->setScriptUrl($this->getBaseUrl() . (str_replace($this->getBaseUrl(), '', $this->getScriptUrl()) ?: '/index.php'));
```
luke-, line 32: As a hotfix for the `master` branch, this change is fine.

But we should create a second PR and refactor this code.

According to the Yii2 documentation, the scriptUrl should contain the relative URL.
https://www.yiiframework.com/doc/api/2.0/yii-web-request#setScriptUrl()-detail

We should therefore not force the BaseURL in front of it.

Can you create another PR for `develop` here and this one for `master`?

luke- (changes requested):

protected/humhub/components/console/UrlManager.php
gevorgmansuryan: @luke- done https://github.com/humhub/humhub/pull/7560

gevorgmansuryan:
@luke- The initial problem was that SafeBaseUrl was being called from a web application, but it was internally switching the urlManager to the console UrlManager. As a result, the console UrlManager was being used within the web app, and UrlManager::createUrl() did not work correctly.

I did some tests and tried several solutions, but the easiest fix is to prevent the console UrlManager from resetting scriptUrl when called from a web application context, while still allowing hostInfo to be reset based on configuredBaseUrl.

luke- (changes requested):
@gevorgmansuryan It is intended that the Console UrlManager (with fixed URLs) will also be used in some places on the web, e.g. for critical processes such as password reset.

It still would be good to have Unit Tests here. Or do we have already?

protected/humhub/components/console/UrlManager.php
```diff
@@ -30,12 +30,15 @@ public function init()
             $hostInfo .= ':' . $urlParts['port'];
         }
         $this->setHostInfo($hostInfo);
-        $this->setScriptUrl($this->getBaseUrl() . ($this->getScriptUrl() ?: '/index.php'));
+
+        if (!Yii::$app instanceof \yii\web\Application) {
```
luke-, line 34: @gevorgmansuryan Why does the Console URL Manager behave differently depending on the web/console application?
Can't we always ensure that the configured values are used, regardless of the application type?

gevorgmansuryan: @luke- because in web ScriptUrl is automatically detected, we do not need to set it

luke-: @gevorgmansuryan But in the SafeBaseUrl context, there should be no automatic detection; instead, it should be explicitly set.

gevorgmansuryan: @luke – The problem originated from the initial issue where `setScriptUrl` was not correctly set in the web application. In the console application, we need to manually specify the script URL (hardcoded as `/index.php`) to make `createUrl` work properly.

Assuming the base URL is `x` and the script URL is `y` in the web context, it will reset the script URL to `xxy` because `getScriptUrl` is not empty in the web app. However, it works fine in the console app since `getScriptUrl` (returns empty there, so it correctly resets it to `xy`.

luke-: @gevorgmansuryan Sorry, I don't understand why we can't always set the correct values here in `UrlManager` during initialization, regardless of whether it's a WebApplication or ConsoleApplication.

- Script URL: relative URL of the entry script
- Base URL:  relative URL for the application. This is similar to [[scriptUrl]] except that it does not include the script file name, and the ending slashes are removed.

If we do that, the `UrlManager` should always returns correct URLs.

(The current PR solution, where we append the `baseUrl` to the `scriptUrl`, looks wrong because the “UrlManager” expects different values. See above.)

gevorgmansuryan:
@luke- we can not detect correct value here. For example we have 

in web app we have:
- baseUrl set to `https://humhub.local/folder1`
- and the current route is /user/profile, so the scriptUrl will be: `/folder1//user/profile`

code ` $this->setScriptUrl($this->getBaseUrl() . ($this->getScriptUrl() ?: '/index.php'));` which was before my changes

will reset script url to: baseurl + script url, which is `https://humhub.local/folder1` + `/folder1/user/profile`, the result will

`https://humhub.local/folder1/folder1/user/profile`, Yii will consider `https://humhub.local/folder1/folder1` as baseUrl and will create new urls based on this.

**we can not change/set script url in web based on config as script url works based on current url**

for console we dont have urls/pages so we can hardcode scriptUrl to make urlManager url creation work. So in console app scriptUrl is empty by default so the code code ` $this->setScriptUrl($this->getBaseUrl() . ($this->getScriptUrl() ?: '/index.php'));` will set scriptUrl to hardcoded baseurl + '/index.php' which will be correct.

marc-farre:
@luke- I see that https://github.com/humhub/humhub/pull/7560, which is for the same issue, is merged. Should we close this one?

luke-:
@marc-farre Yes, please open if it's still a problem.

PR #7449, opened 2025-02-28
Show More - Configuration

gevorgmansuryan:
https://github.com/humhub/humhub-internal/issues/574

### DIFF
--- a/protected/humhub/modules/stream/assets/StreamAsset.php
+++ b/protected/humhub/modules/stream/assets/StreamAsset.php
@@ -14,6 +14,7 @@ use humhub\modules\content\assets\ContentAsset;
 use humhub\modules\content\assets\ContentContainerAsset;
 use humhub\modules\ui\filter\assets\FilterAsset;
 use humhub\modules\user\assets\UserAsset;
+use Yii;
 
 /**
  * Stream related assets.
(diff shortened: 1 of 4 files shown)


protected/humhub/modules/stream/assets/StreamAsset.php
```diff
@@ -48,5 +49,12 @@ class StreamAsset extends AssetBundle
         CoreExtensionAsset::class,
     ];
 
+    public static function register($view)
+    {
+        $view->registerJsConfig('ui.showMore', [
```
luke-, line 54: Maybe we should better this in the `CoreJsConfig`. Because it should not be only set for Streams.

luke- (changes requested):

protected/humhub/modules/stream/assets/StreamAsset.php
gevorgmansuryan: `ui.showMore` is used from all the places anyway. Change?

luke-: @gevorgmansuryan Isn't the Stream asset only used in Stream contexts? 

Yes please change to more global core.

static/js/humhub/humhub.ui.showMore.js
```diff
@@ -1,6 +1,8 @@
 humhub.module('ui.showMore', function (module, require, $) {
     var additions = require('ui.additions');
 
+    alert(module.config.collapseAt);
```
luke-, line 4: @gevorgmansuryan Can you please remove?

luke- (changes requested):

PR #7450, opened 2025-03-01
Stream will break if Wiki Entries are archived

gevorgmansuryan:
https://github.com/humhub/wiki/issues/364


luke-:
@gevorgmansuryan Thanks for the fix!

gevorgmansuryan:
@luke- @yurabakhtin The problem was that “load more” used data-key for loading and insertAfter, which is the latest content element of the group. When the latest element was deleted, it stopped working. Now, it checks all content elements, if previous ones are deleted, it will insert after the last available content element. When all contents are deleted, it will insert after the “load more” button.

PR #7504, opened 2025-04-21
Integrate Optimistic Locking for Post and Comment

gevorgmansuryan:
https://github.com/humhub/rest/issues/196


luke-:
@gevorgmansuryan Thank you for the implementation.

It would probably be better to use the existing `updated_at` to recognize whether the content has been updated in the meantime. (Instead of a new `version` attribute).

In principle, I like such check here, but would put this feature/PR on hold for now. Maybe we could also use the `live` module in future, to inform the user of changed content before `save`.

gevorgmansuryan:
@luke- requested changes are done

PR #7509, opened 2025-04-30
Move mail link service to core

yurabakhtin:

- [x] Feature

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**
https://github.com/humhub/humhub-internal/issues/617

### DIFF
--- a/protected/humhub/modules/admin/views/authentication/authentication.php
+++ b/protected/humhub/modules/admin/views/authentication/authentication.php
@@ -39,7 +39,6 @@ $this->beginContent('@admin/views/authentication/_authenticationLayout.php');
         <?= $form->field($model, 'blockUsers')->checkbox() ?>
         <?= $form->field($model, 'hideOnlineStatus')->checkbox() ?>
         <?= $form->field($model, 'allowUserTopics')->checkbox(['data' => ['action-change' => 'admin.space.restrictTopicCreation']]) ?>
-        <?= $form->field($model, 'disableChoicesIos')->checkbox() ?>
         <?= $form->field($model, 'defaultUserIdleTimeoutSec')->textInput(['readonly' => $userModule->settings->isFixed('auth.defaultUserIdleTimeoutSec')]) ?>
         <p class="help-block"><?= Yii::t('AdminModule.user', 'The default user idle timeout is used when user session is idle for a certain time. The user is automatically logged out after this time.') ?></p>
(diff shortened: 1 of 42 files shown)


protected/humhub/modules/admin/views/authentication/authentication.php
```diff
@@ -39,6 +39,7 @@
         <?= $form->field($model, 'blockUsers')->checkbox() ?>
         <?= $form->field($model, 'hideOnlineStatus')->checkbox() ?>
         <?= $form->field($model, 'allowUserTopics')->checkbox(['data' => ['action-change' => 'admin.space.restrictTopicCreation']]) ?>
+        <?= $form->field($model, 'disableChoicesIos')->checkbox() ?>
```
luke-, line 42: @marc-farre Do we still need this option? At least we should move it somewhere into the config file?

marc-farre: Yes, we do need it, until we provide Apple authentication.
But I agree, we could move it to the config file.

yurabakhtin: @luke- Should we put new config var in the `params/humhub` [here](https://github.com/humhub/humhub/blob/master/protected/humhub/config/common.php#L270), or maybe create some new block like `params/mobile`?

Please note previous value from the module will be lost, or do you know soma way to migrate a setting value from DB to config file?

marc-farre: @yurabakhtin No need to migrate, because I know the HumHub instance that is using it (it's for white label mobile app only).

ArchBlood: > Yes, we do need it, until we provide Apple authentication. But I agree, we could move it to the config file.

If it helps HumHub you can always improve @GreenMeteor's OAuth clients as needed as we do provide a number of different ones;
https://github.com/GreenMeteor/humhub-oauth/blob/master/oauth/Apple/Apple.php

marc-farre: @ArchBlood very interesting. Have you successfully tested it?

yurabakhtin: > A simple param in the humhub block is fine for now.

@luke- Commit https://github.com/humhub/humhub/pull/7509/commits/9bbfd7c210f84ad360a686610a0ad42bf1421ad4.

luke- (approved):

PR #7519, opened 2025-05-12
Fix duplicated notifications on timeout

yurabakhtin:

- [x] Bugfix

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
https://github.com/humhub/calendar/issues/547

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ HumHub Changelog
 - Fix #7503: Don't run migration on updating of disabled module
 - Fix #7513: Fix priority of `dynamic.php`
 - Enh #7514: Update Redis example in `env.example`
+- Fix #7519: Fix duplicated notifications on timeout
 
 1.17.2 (April 7, 2025)
 ----------------------
--- a/protected/humhub/modules/notification/components/NotificationManager.php
+++ b/protected/humhub/modules/notification/components/NotificationManager.php
@@ -13,6 +13,7 @@ use humhub\components\Module;
 use humhub\modules\content\components\ContentContainerActiveRecord;
 use humhub\modules\content\models\Content;
 use humhub\modules\content\models\ContentContainerSetting;
+use humhub\modules\notification\jobs\SendBulkNotification;
 use humhub\modules\notification\targets\BaseTarget;
 use humhub\modules\space\models\Membership;
 use humhub\modules\space\models\Space;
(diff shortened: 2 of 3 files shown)


luke- (changes requested):
@yurabakhtin 

- The ActiveJob is currently set to no retry https://github.com/humhub/humhub/blob/dfa3f28d75e38523245811cf254a5122f1fc81ab/protected/humhub/modules/queue/LongRunningActiveJob.php#L26 I find it strange that the timeout still triggers a retry. I think we should avoid that somehow?
- Can you explain, why the retry is after 1 hour? Maybe the job runs about 1 hour before retry?
- In principle, other queue drivers such as Redis should also be supported? I'm not sure if other queue features like “ExclusiveJob” work with it, but I think updating a task will be difficult to realize with other queue drivers.
- I do like that we store the processed users, but a more lightweight solution would probably be nice here instead of serializing & save the whole object after each loop. Either cache or a lightweight DB table?

yurabakhtin:
> * The ActiveJob is currently set to no retry https://github.com/humhub/humhub/blob/dfa3f28d75e38523245811cf254a5122f1fc81ab/protected/humhub/modules/queue/LongRunningActiveJob.php#L26
>    I find it strange that the timeout still triggers a retry. I think we should avoid that somehow?

@luke- Yes, it is very strange, it is why I could not reproduce the bug long time, and I found only this one way(set `true` there) to make the job repeatable.

> * Can you explain, why the retry is after 1 hour? Maybe the job runs about 1 hour before retry?

As I tested only with `canRetry === true` then the retry is fired after 1 hour because [here](https://github.com/humhub/humhub/blob/master/protected/humhub/modules/queue/Module.php#L26) `$longRunningJobTtr = 60 * 60;` (1 hour). I have only this explnation at this time, i.e. the failed job is repeated after 1 hour when the command `yii queue/listen` is running.

> * In principle, other queue drivers such as Redis should also be supported? I'm not sure if other queue features like “ExclusiveJob” work with it, but I think updating a task will be difficult to realize with other queue drivers.

I have implemented this only for queue drivers extended from database drivers `yii\queue\db\Queue`.
HumHub by default uses `\humhub\modules\queue\driver\MySQL` extended from `yii\queue\db\Queue`, but tests use `humhub\modules\queue\driver\Instant` extended from not db driver `yii\queue\Queue`, so there the fix is ignored.
I see Redis `yii\queue\redis\Queue` is extended from not db driver `yii\queue\cli\Queue`, then there it will not work. Should I investigate how it can be fixed for the driver?

> * I do like that we store the processed users, but a more lightweight solution would probably be nice here instead of serializing & save the whole object after each loop. Either cache or a lightweight DB table?

Hm, ok I will review again what we can do there, I thought it is not hard to update the queue record in db...

PR #7526, opened 2025-05-15
InstallationState issues fix

gevorgmansuryan:
https://github.com/humhub/humhub-internal/issues/641


protected/humhub/components/ModuleManager.php
```diff
@@ -176,7 +169,13 @@ public function register(string $basePath, ?array $config = null)
         ) {
             $errorMessage = Yii::t('error', 'The module {moduleId} is present in the HumHub configuration file even though this module is disabled. Please remove it from the configuration.', ['moduleId' => '“' . $config['id'] . '”']);
             Yii::error($errorMessage);
-            Yii::$app->view?->error($errorMessage);
+
+            if (Yii::$app instanceof \yii\console\Application) {
```
luke-, line 173: Why do we need the change here?

luke- (changes requested):

protected/humhub/components/ModuleManager.php
gevorgmansuryan: @luke- it triggered an error in console app, because `Yii::$app->view?->error` uses Yii::$app->session->setFlesh... and console application does not have session

PR #7540, opened 2025-05-27
BS5 - Implement wrappers for jQuery show/hide methods to work with bootstrap style `d-none` instead of `display:none`

yurabakhtin:
@marc-farre I have a suggestion for the jQuery show/hide methods https://github.com/humhub/humhub/blob/bs5/MIGRATE-DEV-BS5.md#javascript-with-d-none, please review how we can implement wrappers to use `d-none` style class instead of the inline `display: none`.

Do you find some problems if we will use this?

I would like to use the auto wrapper because:
- Many js code should be updated,
- External libaries cannot be edited.

PS: If you know better name for `jquery.fix.js` please change that, maybe `jquery.bootstrap.js` or `jquery.bs5.js`.


marc-farre:
@yurabakhtin thanks for your solution!

What do you think about naming it `jquery.bootstrap3.js`, the same way we have creating this CSS compatibility code?
https://github.com/humhub/humhub/blob/bs5/static/scss/_bootstrap3.scss

We could mark it as deprecated, and then try slowly (in future HH versions) to migrate the code to BS5 recommended practice (even jQuery has been dropped in BS5).

marc-farre:
@yurabakhtin Done.
Please check my commit if it was OK to replace `original` with `bs3`:
https://github.com/humhub/humhub/pull/7540/commits/464d4a53e755aca0aa25a9625e2c01c90b862032#diff-e64b19c4e1f7fc90f6c9156cd375aec9b04b96bd022770efae7d529965e48132L3

yurabakhtin:
@marc-farre Maybe `bs3` is ok, but I don't unserstand this `TODO: Remove when Bootstrap 3 is no longer supported`.
Does it mean when we will remove Bootstrap 3 we also delete jQuery and Bootstrap 5 has functions with same names as jQuery?
But if Bootstrap 5 doesn't have jQuery functions then does it mean we should rewrite all JS code to new Bootstrap 5 functions?

marc-farre:
@luke- Sorry, my "Examples of replacements:" section was unclear.
This section is not about this PR.
It's just to show examples for current jQuery code that can be replaced in the long term with vanilla JS (or BS5 methods).
But the current code work, because jQuery can still be used together with BS5: https://getbootstrap.com/docs/5.3/getting-started/javascript/#optionally-using-jquery

However, there is a conflict with the new BS5 `d-none` class which has a `display:none !important` style.
So jQuery `show()` and others don't work anymore because of this `!important`.
@yurabakhtin made this PR to still allow using these jQuery methods, but I'll leave @yurabakhtin explain more about it.

yurabakhtin:
@marc-farre I have understood the idea of replacing jQuery with Vanilla JS in future, thanks for the explanation.

> However, there is a conflict with the new BS5 `d-none` class which has a `display:none !important` style.
> So jQuery `show()` and others don't work anymore because of this `!important`.
> @yurabakhtin made this PR to still allow using these jQuery methods, but I'll leave @yurabakhtin explain more about it.

@luke- Yes, Marc is right, I made this PR in order to don't do many replacements like from `.hide()` to `.addClass('d-none')` as I had to do it for example [here](https://github.com/humhub/calendar/pull/549/files#diff-92984e50c9b6d70bb3fc7a5996c0f7bfcd1e1a67f5d504ca0da8f0574c6507a6L180-L181).

marc-farre:
@yurabakhtin commit https://github.com/humhub/humhub/pull/7540/commits/9260876cae7cd31cb5dd5b952a4106fd89223c2c

@luke- the PR is ready (and rebased), you can merge.

PR #7552, opened 2025-06-06
Downgrade composer package `behat/gherkin` to fix tests

yurabakhtin:

- [x] Bugfix

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**
https://github.com/humhub/humhub-internal/issues/672


protected/humhub/config/assets-prod.php
```diff
@@ -51,6 +51,14 @@
             'app',
         ],
     ],
+    'humhub\\assets\\JqueryBs5FixAsset' => [
```
marc-farre, line 54: @yurabakhtin Why do we have this in this PR?

yurabakhtin: @marc-farre The file was updated automatically when I run `grunt build-assets`.
I think it is because we recently added `JqueryBs5FixAsset::class` to the class `humhub\assets\AppAsset`.

PR #7561, opened 2025-06-18
Cache issue with develop branch / db error

gevorgmansuryan:
https://github.com/humhub/humhub-internal/issues/676


protected/humhub/modules/admin/views/setting/caching.php
```diff
@@ -13,8 +14,15 @@
...
 
-<?php if (!Yii::$app->settings->isFixed('cacheClass')): ?>
-    <?= $form->field($model, 'type')->dropDownList($cacheTypes, ['readonly' => Yii::$app->settings->isFixed('cacheClass')]) ?>
+<?php if (!$model->isTypeFixed): ?>
+    <?= $form->field($model, 'type')->dropDownList($cacheTypes) ?>
+<?php else: ?>
+    <?= $form->field($model, 'type')->textInput([
+        'value' => ArrayHelper::getValue($cacheTypes, $model->fixedTypeValue, $model->fixedTypeValue),
+        'readonly' => true,
+        'title' => Yii::t('base', 'Specified in the configuration file'),
```
luke-: Please use Message Category "AdminModule.settings"

luke- (changes requested):

luke-:
@gevorgmansuryan Wouldn't it be better to simply not define any cache components in the standard configuration instead of an “InitialCache”?

This would result in an error if an attempt was made to use the cache before it was set via the `ComponentLoader`.

Then we can fix such cases. Currently, such early cache requests would simply be directed to the “DummyCache/InitialCache” and not cached at all. Even if it appears to be the case.

gevorgmansuryan:
@luke- `ComponentLoader` relies on `SettingsManager` which uses `cache` in most places. We need to refactor `SettingsManager` (which is very risky as all application/modules relies on this) to make it work without cache first.

luke-:
@gevorgmansuryan Okay, that's definitely a problem if we need the cache before we can set the cache component.

The cleanest and most reliable solution here would probably be to completely remove all cache settings from the UI and also the component loader. We can then specify “FileCache” as the fixed default in the default configuration file.

Changes to the cache must then be defined via the config file or ENV.

Can you please provide a PR (Maybe a new one), and remove all Cache Config Stuff (except config file). Please also create a migration to delete cache releated settings.

---

However, we still need a prominent button to clear all caches. Maybe we add it here for now:

image

@mbumpalumpa Or any other idea for the button?

gevorgmansuryan:
@luke- done

luke-:
@gevorgmansuryan Thanks. Can you please also remove the tab and page here:

image

For now please simply add a button "Clear Cache" on `Settings` -> `General`. For that use a new Action `SettingController::actionClearCache`.

luke-:
> @gevorgmansuryan Thanks. Can you please also remove the tab and page here:
> 
> image
> For now please simply add a button "Clear Cache" on `Settings` -> `General`. For that use a new Action `SettingController::actionClearCache`.

---

Can you please scan if any module uses `CacheSettingsForm` for flush caches. (e.g. Updater) module. If so, we should provide a PR for these modules and add this change to `Migrate-DEV.md`  file.

gevorgmansuryan:
@luke- module scan was already done, there is no any usage of `CacheSettingsForm` in modules. Cache tab removed

luke- (approved):

luke-:
@gevorgmansuryan Thanks, can you please fix the tests?

gevorgmansuryan:
@luke- tests are ok now

protected/humhub/modules/activity/tests/codeception/acceptance.suite.yml
```diff
@@ -20,6 +20,7 @@ modules:
             window_size: maximize
             browser: chrome
             port: 4444
+            restart: true
```
luke-, line 23: @gevorgmansuryan Can you briefly explain what the function does and why we need it?

gevorgmansuryan: Its restarts the browser session before running suite, if browser was opened before

luke-: @gevorgmansuryan Is this required now or does it also works without?

gevorgmansuryan: @luke- not required, but with this tests will work more stable after switching modules. all other modules has this config enabled

luke- (approved):

PR #7564, opened 2025-06-23
Support Login BGs

luke-:
Support Login Background in BS5 as well

### DIFF
--- a/protected/humhub/modules/user/helpers/LoginBackgroundImageHelper.php
+++ b/protected/humhub/modules/user/helpers/LoginBackgroundImageHelper.php
@@ -1,0 +1,79 @@
+<?php
+
+/**
+ * @link https://www.humhub.org/
+ * @copyright Copyright (c) 2025 HumHub GmbH & Co. KG
+ * @license https://www.humhub.com/licences
+ */
+
+namespace humhub\modules\user\helpers;
+
+use Yii;
+use yii\imagine\Image;
+use yii\helpers\FileHelper;
+
+final class LoginBackgroundImageHelper
+{
+
+    private const ASSETS_PATH = 'login-bg';
+    private const ASSETS_FILE = 'background.png';
+
+    private const STORE_PATH = '@webroot/uploads/login-bg';
+
+    public static function set(?string $fileName): void
+    {
+        @unlink(static::getAssetsFile());
+        @unlink(static::getStoreFile());
+
+        if ($fileName) {
+            Image::getImagine()->open($fileName)->save(self::getStoreFile());
+        }
+    }
+
+    public static function getUrl(): ?string
+    {
+        // Check if the file exists at all
+        if (!static::hasImage()) {
+            return null;
+        }
+
+        if (!file_exists(self::getAssetsFile())) {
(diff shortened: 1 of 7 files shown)


luke-:
@marc-farre I've added the possibilty to migrate and also set background for the Login Page via Appearance. Would be great if you could do a quick review and merge if ok.

protected/humhub/modules/user/views/auth/login.php
```diff
@@ -19,6 +21,22 @@
...
+        .login-container {
+            height:100%;
+            background-image: url("' . LoginBackgroundImageHelper::getUrl() . '");
+            background-size: cover;
+            background-position: center;
+            background-repeat: no-repeat;
+        }');
+}
+
+
```
marc-farre, line 39: @luke- What do you think about moving this CSS code to `static/scss/_login.scss` instead?
It would make easier for themes to overwrite the values (if inline, we need `!important` to overwrite).
And it could be cleaner to have all the CSS in the same file.

We could add a class such as `login-container-bg-img` when a background image is defined:

image

What do you think?

marc-farre:
> @marc-farre I've added the possibilty to migrate and also set background for the Login Page via Appearance. Would be great if you could do a quick review and merge if ok.

@luke- Thanks for this great feature!
I think it's all good. Just the small CSS suggestion in the review. I can send a commit for it if you want.

protected/humhub/modules/user/helpers/LoginBackgroundImageHelper.php
```diff
@@ -0,0 +1,79 @@
...
+
+final class LoginBackgroundImageHelper
+{
+
+    private const ASSETS_PATH = 'login-bg';
+    private const ASSETS_FILE = 'background.png';
+
+    private const STORE_PATH = '@webroot/uploads/login-bg';
+
+    public static function set(?string $fileName): void
```
ArchBlood, line 23: I think the current approach might be a bit risky since it suppresses potential errors without handling them. It might be safer and more transparent to do something like this instead:

```php
public static function set(?string $fileName): void
{
    // Clean up any existing files
    foreach ([static::getAssetsFile(), static::getStoreFile()] as $file) {
        if (file_exists($file)) {
            unlink($file); // You could consider handling failure here if needed
        }
    }

    // Save new image if one is provided
    if ($fileName !== null) {
        Image::getImagine()->open($fileName)->save(self::getStoreFile());
    }
}
```
This avoids silent failures and makes it easier to track down issues if something goes wrong.

luke-: @ArchBlood Thanks, I've improved it.

protected/humhub/modules/user/views/auth/login.php
luke-: Changed here: https://github.com/humhub/humhub/commit/fc57036516589b61039bcacc7960d40ebeaa40e6

marc-farre: @luke- Thanks!

PR #7565, opened 2025-06-29
Fixstream filter placeholder

gevorgmansuryan:
https://github.com/humhub/humhub-internal/issues/682

### DIFF
--- a/protected/humhub/modules/ui/filter/widgets/DatePickerFilterInput.php
+++ b/protected/humhub/modules/ui/filter/widgets/DatePickerFilterInput.php
@@ -12,7 +12,6 @@ namespace humhub\modules\ui\filter\widgets;
 use DateTime;
 use humhub\modules\ui\form\widgets\DatePicker;
 use Yii;
-use yii\helpers\FormatConverter;
 
 class DatePickerFilterInput extends FilterInput
 {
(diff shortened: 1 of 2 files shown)


protected/humhub/modules/ui/filter/widgets/DatePickerFilterInput.php
```diff
@@ -40,8 +39,7 @@ protected function initFromRequest()
         $this->value = $filter;
 
         if (!isset($this->options['placeholder'])) {
-            $this->options['placeholder'] = (new DateTime())
-                ->format(FormatConverter::convertDateIcuToPhp(Yii::$app->formatter->dateInputFormat));
+            $this->options['placeholder'] = Yii::$app->formatter->asDate(new DateTime(), Yii::$app->formatter->dateInputFormat);
```
luke-, line 42: @gevorgmansuryan Does it now display the current date instead of e.g. “DD-MM-YY”?

gevorgmansuryan: @luke- it always showed current date. The problem was that `FormatConverter::convertDateIcuToPhp(Yii::$app->formatter->dateInputFormat` returned `php:xxx` and `short` options which are not acceptable by `DateTime::format`. Ive just changed `DateTime::format` to `Yii::$app->formatter::asDate`

PR #7566, opened 2025-06-29
Registration with error just reloads

gevorgmansuryan:
https://github.com/humhub/humhub-internal/issues/679

### DIFF
--- a/protected/humhub/modules/user/models/Invite.php
+++ b/protected/humhub/modules/user/models/Invite.php
@@ -13,6 +13,7 @@ use humhub\components\ActiveRecord;
 use humhub\modules\space\models\Space;
 use humhub\modules\user\Module;
 use Yii;
+use yii\base\InvalidConfigException;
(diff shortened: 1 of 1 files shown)


protected/humhub/modules/user/models/Invite.php
```diff
@@ -157,8 +158,25 @@ public function selfInvite()
             return false;
         }
 
-        if ($this->save()) {
-            return $this->sendInviteMail();
+        $transaction = Yii::$app->db->beginTransaction();
+
+        try {
+            if ($this->save()) {
+                if (!$this->sendInviteMail()) {
```
luke-, line 165: Is the error occurring here in this method?

I would prefer to catch and log the error in this method and return `false`. Then an 500 error should be shown on registration?

gevorgmansuryan: @luke- there is no error. We just rely on returning value of `Mailer::send` which is called inside `sendInviteMail`

PR #7608, opened 2025-07-31
Added Rector

luke-:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**


protected/humhub/libs/ParameterEvent.php
```diff
@@ -18,17 +18,15 @@
...
-
     /**
      * @inheritdoc
+     * @param mixed[] $parameters
+     */
+    public function __construct(/**
+     * @var array the parameter references
      */
-    public function __construct($parameters)
+    public $parameters)
```
marc-farre, line 28: @luke- I think there is an issue here. See the final result: https://github.com/humhub/humhub/blob/7aaeefb95638b019243166eb94fdaa2deb606c93/protected/humhub/libs/ParameterEvent.php

rector.php
```diff
@@ -0,0 +1,22 @@
...
+    ])
+    ->withSkip([
+        \Rector\Php81\Rector\Array_\FirstClassCallableRector::class,
+        \Rector\TypeDeclaration\Rector\ClassMethod\ReturnNeverTypeRector::class,
+        __DIR__ . '/protected/humhub/config/*',
+        __DIR__ . '/protected/humhub/messages/*',
+        __DIR__ . '/protected/humhub/modules/*/messages/*',
+        __DIR__ . '/protected/humhub/modules/*/config.php',
+    ])
+    ->withPhpSets(php82:true)
```
marc-farre, line 19: @luke- shouldn't we specify that we want to use PHP 8.2 rules as base and include up to PHP 8.4 rules?
```php
    ->withPhpSets(
        php82: true,
        php83: true,
        php84: true,
    )
 ```

luke-: @marc-farre As I understand it, Rector migrates the code to PHP MinVersion 8.2 with `php82`.
When using `php84` it would already use PHP 8.4 features, which may not compatible with PHP 8.2.

protected/humhub/libs/ParameterEvent.php
luke-: It's correct but a bit ugly with the PHPDoc. With newer PHP versions you can use `public/private` in constructor method parameters to automatically declare properties.

I've cleaned it up a bit: https://github.com/humhub/humhub/commit/6653b293d6f938fc2ab51f7c0652cc6de71a004f

rector.php
marc-farre: @luke- Yes, sorry, I got caught by AI mistake! I should have read the doc before...

The idea behind was to make sure that the code will also work with 8.3 and 8.4, but I was wrong.

However, the official doc https://getrector.com/documentation/set-lists#content-php-sets says:

The best practise is to use PHP version defined in composer.json. Rector will automatically pick it up with empty ->withPhpSets() method.

Perhaps we should go for that. This way, when we change PHP version, we don't forget to update rector.php too.
What do you think?

luke-: @marc-farre good point! I've changed that in `develop`!

PR #7638, opened 2025-08-19
BS5 - Installer not working when `assets/` dir is empty

luke-:
@marc-farre If the Assets Directory is empty and the theme is compiled before installation (without Settings DB), the following error occurs.

Is my fix okay?

image


marc-farre:
@luke- Yes, I think this fix is fine.
I tested a full new install process using this `bs5-installer-fix` branch, and it's all OK.

PR #7649, opened 2025-08-23
Update HH Branding

mbumpalumpa:

- [x] Other, please describe:
Idk if the HumHub logo in the installer is styled somewhere in CSS, but the old logo was 200x60px and the new one is 843x248px, so if it’s not limited via CSS, the logo might be gigantic. I changed the size to 248px so it’s more "HD" than the current one.

- [ ] Yes
- [x] No

**Other information:**
https://github.com/humhub/humhub-internal/issues/718


luke- (changes requested):
@mbumpalumpa We should the limit the size on the installer:

image


Maybe be logo a bit bigger here?

image

(But as you prefer.)

mbumpalumpa:
@luke- I have resized the logo to the old width, or would u rather we add inline styling to limit the width to 200px, like the old one? Also, see the mockup for the honeycomb; if I understood correctly, you were referring to it being too small?

Side by Side

luke-:
@mbumpalumpa Better would be to add a inline style zu reduce with. But as you like. We can also keep the current solution.

I thought the logo was a bit too small inside the box. But you decide :-)

mbumpalumpa:
@luke- In the first commit, I did 248px, but the difference wasn’t that "visible", as the original was 200px. So the current one is 296px wide. Let me know if it’s ok, and also mockups show 248 vs 296. I'm personally fine with 248px. 

Frame 534

Also, this SS is showing v 1.17.3 Installer, not the BS5 version.

---

Also, now re-reading your comment, were you referring to the size of the honeycomb for the Admin->Info logo, or the logo at the Installer being small?

luke-:
> @luke- In the first commit, I did 248px, but the difference wasn’t that "visible", as the original was 200px. So the current one is 296px wide. Let me know if it’s ok, and also mockups show 248 vs 296. I'm personally fine with 248px.
> 
The installer logo is now fine for me

> Also, now re-reading your comment, were you referring to the size of the honeycomb for the Admin->Info logo, or the logo at the Installer being small?

Not sure what is nicer, my point was to increase the size of the honeycomb. e.g.

image

PR #7703, opened 2025-09-13
Image upload widget

gevorgmansuryan:
https://github.com/humhub/news/issues/105


luke-:
@gevorgmansuryan Thanks!

marc-farre:
It could be helpful to have a dedicated HTML class, allowing so CSS styling in custom themes, or JS interactions.

yurabakhtin:
@luke- @gevorgmansuryan It works well, but it would be better to allow to pass options for each widget as:

```php
$form->field($model, 'cover')->widget(ActiveFileUpload::class, [
    'inputOptions' => [],
    'previewOptions' => [],
    'progressOptions' => [],
])
```

then merge them with default options what used for `UploadInput`, `FilePreview` and `UploadProgress` inside.

At least for the preview widget we should allow to set sizes(width and height), because currently it is a static:

preview

Also the control buttons must have better style position (maybe absolute inside the block with relative position) because on small screen:
small-screen

and the height must be auto(not set) by default to keep original image ratio.

protected/humhub/modules/file/widgets/ActiveFileUpload.php
```diff
@@ -0,0 +1,79 @@
...
+use humhub\modules\file\assets\ActiveFileUploadAsset;
+use humhub\modules\file\models\File;
+use yii\bootstrap5\InputWidget;
+use yii\helpers\Url;
+
+/**
+ * Example of usage
+ *
+ * ```php
+ * <?= $form->field($model, 'attribute')->widget(ActiveFileUpload::class. ['accept' = 'image/*', 'hideInStream' => true]) ?>
```
yurabakhtin: small typos here `class, ['accept' =>`

yurabakhtin (changes requested):

luke-:
@yurabakhtin @marc-farre Thanks for your feedback!

@gevorgmansuryan Can you please adjust?

gevorgmansuryan:
@luke- done

yurabakhtin:
@gevorgmansuryan I think the styles should be fixed when no image:

test-upload

gevorgmansuryan:
@yurabakhtin should be ok now

yurabakhtin:
fixed-height

yurabakhtin (approved):

PR #7708, opened 2025-09-15
Active Form Safe Fields only

luke-:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**

### DIFF
--- a/protected/humhub/modules/admin/views/group/edit.php
+++ b/protected/humhub/modules/admin/views/group/edit.php
@@ -18,59 +18,49 @@ use yii\helpers\Url;
 
 <?php $this->beginContent('@admin/views/group/_manageLayout.php', ['group' => $group]) ?>
 <div class="panel-body">
-    <?php $form = ActiveForm::begin(['acknowledge' => true]) ?>
-    <?= $form->field($group, 'name')->textInput(['readonly' => !$canManage]) ?>
+    <?php $form = ActiveForm::begin(['acknowledge' => true, 'renderOnlySafeFields' => true]) ?>
+
+    <?= $form->field($group, 'name', ['inputOptions' => ['readOnly' => !$canManage]])->textInput() ?>
     <?= $form->field($group, 'description')->textarea(['rows' => 5, 'readonly' => !$canManage]) ?>
+    <?= $form->field($group, 'type')->dropDownList($group::getTypeOptions()) ?>
+    <?= $form->field($group, 'subgroups')->widget(GroupPicker::class, ['groupType' => $group::TYPE_NORMAL]) ?>
+    <?= $form->field($group, 'parent')->widget(GroupPicker::class, ['groupType' => $group::TYPE_SUBGROUP]) ?>
 
-    <?php if ($canManage) : ?>
-        <?= $form->field($group, 'type')->dropDownList($group::getTypeOptions()) ?>
-        <?= $form->field($group, 'subgroups')->widget(GroupPicker::class, ['groupType' => $group::TYPE_NORMAL]) ?>
-        <?= $form->field($group, 'parent')->widget(GroupPicker::class, ['groupType' => $group::TYPE_SUBGROUP]) ?>
-    <?php endif ?>
+    <?= SpacePickerField::widget([
+        'form' => $form,
+        'model' => $group,
+        'attribute' => 'defaultSpaceGuid',
+        'selection' => $group->defaultSpaces,
+        'maxSelection' => 1000,
+        // Group managers should see spaces only where they are admins or owners
+        'url' => $canManage ? null : Url::to(['/space/browse/search-json', 'user' => 'admin-owner']),
+    ])
+    ?>
+    <?= $form->field($group, 'updateSpaceMemberships')->checkbox() ?>
+    <?php $url = ($group->isNewRecord) ? null : Url::to(['/admin/group/admin-user-search', 'id' => $group->id]) ?>
+    <?= $form->field($group, 'managerGuids')->widget(UserPickerField::class, [
+        'selection' => $group->manager,
+        'url' => $url,
+        'disabled' => !$canManage,
+    ]) ?>
 
-    <?php if (!$group->is_admin_group): ?>
-        <?= SpacePickerField::widget([
-            'form' => $form,
-            'model' => $group,
-            'attribute' => 'defaultSpaceGuid',
-            'selection' => $group->defaultSpaces,
-            'maxSelection' => 1000,
-            // Group managers should see spaces only where they are admins or owners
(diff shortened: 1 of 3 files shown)


protected/humhub/modules/admin/views/group/edit.php
```diff
@@ -18,71 +18,61 @@
...
-    <?php $form = ActiveForm::begin(['acknowledge' => true]) ?>
-    <?= $form->field($group, 'name')->textInput(['readonly' => !$canManage]) ?>
-    <?= $form->field($group, 'description')->textarea(['rows' => 5, 'readonly' => !$canManage]) ?>
-
-    <?php if ($canManage) : ?>
-        <?= $form->field($group, 'type')->dropDownList($group::getTypeOptions()) ?>
-        <?= $form->field($group, 'subgroups')->widget(GroupPicker::class, ['groupType' => $group::TYPE_NORMAL]) ?>
-        <?= $form->field($group, 'parent')->widget(GroupPicker::class, ['groupType' => $group::TYPE_SUBGROUP]) ?>
-    <?php endif ?>
+    <?php $form = ActiveForm::begin(['acknowledge' => true, 'renderOnlySafeFields' => true]) ?>
```
luke-: @yurabakhtin maybe `renderOnlySafeAttributeFields` would be a good naming?

yurabakhtin: @luke- I think better name is `renderOnlyActiveAttributes` because parent Yii model has methods as `isAttributeActive()` and `activeAttributes()` which return a result depending on scenarios.

luke-: @yurabakhtin Ok, thats better!

protected/humhub/components/ActiveRecord.php
```diff
@@ -280,4 +280,27 @@ public static function getObjectModel(): string
...
     }
+
+    /**
+     * Remove attributes from all scenarios
+     *
+     * @param array $scenarios
+     * @param string|array $attributes
+     * @since 1.18
+     */
+    public function removeScenarioAttributes(array &$scenarios, string|array $attributes): void
```
luke-, line 291: @yurabakhtin Maybe `protected` and `static`?

protected/humhub/modules/user/models/forms/EditGroupForm.php
```diff
@@ -55,18 +55,37 @@ public function afterFind()
...
+    /**
+     * @inheritdoc
+     */
+    public function scenarios()
+    {
+        $scenarios = parent::scenarios();
+
+        $scenarios[self::SCENARIO_MANAGER][] = 'updateSpaceMemberships';
+
+        if ($this->isNewRecord) {
```
luke-, line 82: Not tested, but couldn't we use something like
'$rules[] = [['updateSpaceMemberships'], 'boolean', 'when' => ($this->isNewRecord)];'
instead of moving some rules into the `scenario` section?

protected/humhub/widgets/form/ActiveField.php
```diff
@@ -155,4 +161,20 @@ public function radioList($items, $options = [])
...
+    /**
+     * @inheritdoc
+     */
+    protected function addAriaAttributes(&$options)
+    {
+        $this->setReadOnlyFromOptions($options);
+        parent::addAriaAttributes($options);
+    }
+
+    protected function setReadOnlyFromOptions(array $options): void
```
luke-: Can we make this private?

protected/humhub/widgets/form/ActiveField.php
```diff
@@ -155,4 +161,20 @@ public function radioList($items, $options = [])
 
         return parent::radioList($items, $options);
     }
+
+    /**
+     * @inheritdoc
+     */
+    protected function addAriaAttributes(&$options)
+    {
+        $this->setReadOnlyFromOptions($options);
```
luke-, line 170: I really like this "hack".  But we should add a comment here, why we do this here. e.g.

```
// Directly calling, e.g., `textInput($inputOptions)` does not merge the options into `ActiveField::$inputOptions`. 
// Since the `readOnly` option is needed later for `$preventRendering`, the option is intercepted here.   
```

luke- (commented):
@yurabakhtin Thanks for the changes. I really like. Please see my comments.

protected/humhub/components/ActiveRecord.php
yurabakhtin: @luke- What reason to keep it as and `protected` and `static`?
If we make it `protected` then it is impossible to use it from outside the class like `ActiveRecord::removeScenarioAttributes()`.
If we don't need to use it outside then it is enough to make it only `protected`, otherwise `public` + `static`.
Am I wrong somewhere?

protected/humhub/modules/user/models/forms/EditGroupForm.php
yurabakhtin: @luke- I tried to use the option/filter `when` but it doesn't do any effect to a result of the methods `isAttributeActive()` and `activeAttributes()` because the result is built from the method `scenarios()` which use only rule options/filters `on` and `except`.

protected/humhub/widgets/form/ActiveField.php
yurabakhtin: @luke- Ok, we can, at this moment I don't find where the class is extended, I just did it protected to avoid restriction for future extended classes.

protected/humhub/widgets/form/ActiveField.php
yurabakhtin: https://github.com/humhub/humhub/pull/7708/commits/10515bb38f9fd589515c5be9c86473e96705884b

protected/humhub/widgets/form/ActiveField.php
yurabakhtin: https://github.com/humhub/humhub/pull/7708/commits/10515bb38f9fd589515c5be9c86473e96705884b

luke-: I think we should be rather restrictive from the beginning. If necessary, we can always relax the rules later.

protected/humhub/modules/user/models/forms/EditGroupForm.php
luke-: @yurabakhtin Ok, too bad,then let's leave it as it is.

protected/humhub/components/ActiveRecord.php
luke-: @yurabakhtin I think, we should be restrictive as possible. So `protected` would be fine. Static because it doesn't use any instance properties. But we can leave it as it is.

PR #7728, opened 2025-09-21
Fix: Acceptance tests for Altcha captcha in password recovery

marc-farre:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**

### DIFF
--- a/protected/humhub/modules/user/models/forms/AccountRecoverPassword.php
+++ b/protected/humhub/modules/user/models/forms/AccountRecoverPassword.php
@@ -16,17 +16,22 @@ class AccountRecoverPassword extends Model
     public $captcha = '';
     public $email;
 
+    public const TESTING_BYPASS_CAPTCHA_EMAIL = 'test-captcha@mail.de';
+
     /**
      * Declares the validation rules.
      */
     public function rules()
     {
-        return [
+        $rules = [
             ['email', 'required'],
             ['email', 'email'],
-            ['captcha', Yii::$app->captcha->getValidatorClass()],
             ['email', 'verifyEmail'],
         ];
+        if (!$this->skipCaptchaInTestEnv()) {
+            $rules[] = ['captcha', Yii::$app->captcha->getValidatorClass()];
+        }
+        return $rules;
     }
 
     /**
@@ -84,4 +89,9 @@ class AccountRecoverPassword extends Model
         return $user->getPasswordRecoveryService()->sendRecoveryInfo();
     }
 
+    private function skipCaptchaInTestEnv(): bool
+    {
+        return YII_ENV_TEST
+            && (Yii::$app->request->post('AccountRecoverPassword')['email'] ?? null) !== static::TESTING_BYPASS_CAPTCHA_EMAIL;
+    }
 }
--- a/protected/humhub/modules/user/tests/codeception/acceptance/PasswordRecoveryCest.php
+++ b/protected/humhub/modules/user/tests/codeception/acceptance/PasswordRecoveryCest.php
@@ -2,6 +2,7 @@
 
 namespace user\acceptance;
 
+use humhub\modules\user\models\forms\AccountRecoverPassword;
 use tests\codeception\_pages\LoginPage;
 use user\AcceptanceTester;
 
@@ -17,7 +18,7 @@ class PasswordRecoveryCest
         $I->waitForText('Forgot your password?');
         $I->jsClick('#password-recovery-link');
         $I->waitForText('Password recovery');
-        $I->fillField('#email_txt', 'wrong@mail.de');
+        $I->fillField('#email_txt', AccountRecoverPassword::TESTING_BYPASS_CAPTCHA_EMAIL);
         $I->executeJS('$(\'[class*="captcha"]\').remove();'); // Remove any existing captcha to simulate not checking it or hacking
         $I->click('Reset password');
         $I->wait(3);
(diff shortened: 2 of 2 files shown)


luke- (approved):

marc-farre:
@luke- I can validate the captcha with JS with a simple click().
It also works on https://altcha.org/ (tested in the browser console).
So I cannot tell if it's a security issue, but if it's the case, it's not on our side.

So now I managed to test if the captcha is not ticked by deleting the input via JS (which also simulates a hacker), to check if the server side blocks the request. That's fine.

But after simulating the captcha click, on my local server, I can submit the form, but not on GittHub.
This is GitHub output:

image

I tried to add some extra delay, and to change the way of submitting the form, but same result. 

Also, I have fixed the buttons order and color.

Before:
image

After:
image

marc-farre:
@yurabakhtin would you have an idea maybe why the form is not submitted on GitHub (it works on my local machine)? Thanks!

luke-:
> @luke- I can validate the captcha with JS with a simple click(). It also works on https://altcha.org/ (tested in the browser console). So I cannot tell if it's a security issue, but if it's the case, it's not on our side.

Altcha is using PoW, so it causes "just" some load on the client side. So a click should work. 

> So now I managed to test if the captcha is not ticked by deleting the input via JS (which also simulates a hacker), to check if the server side blocks the request. That's fine.
> 
> But after simulating the captcha click, on my local server, I can submit the form, but not on GittHub. This is GitHub output:
> 

Maybe GitHub is a bit slower here. I would say we should remove the Captcha solving in the tests.

marc-farre:
@luke- If we want to skip captcha, we need to do it on the server side, and acceptances shouldn't be able to self-skip them, because it simulates the browser which should have no way to skip it.

So I have coded the skip in the model:
https://github.com/humhub/humhub/pull/7728/files#diff-be06f02d9e168ff8542eed25a5b1e5bdc68285dc2fc8341cf122a33204cbc2fe by adding the new `skipCaptchaInTestEnv()` method.

What do you think about this solution?

Now AccountRecoverPassword tests work:

image

But others are failing. I don't see how this PR can be the cause of the other errors.
Should I try to rebase the PR from develop?
Or investigate more the other failures?

luke-:
@marc-farre Wouldn't it be better, to create here a Dummy Captcha Driver and enable this for testing?
https://github.com/humhub/humhub/tree/develop/protected/humhub/components/captcha

This would also allow users to turn off captcha at all?

marc-farre:
> @marc-farre Wouldn't it be better, to create here a Dummy Captcha Driver and enable this for testing? 

@luke- But if I do this, then I cannot test that captcha blocks form submitting if not ticked.
The idea here is also to make sure that captcha is working as expected.
Or did you have an idea in mind to make this captcha test possible with a Dummy Captcha Driver?

luke-:
@marc-farre Hmm, problem is, that for the acceptance tests, we can currently only use one specific captcha driver defined in the configuration file. 

I think our options are:

1. Switch to YiiCaptcha in the AcceptanceTests for now.  (my current preferred solution)
2. Create some option to force positive response from Altcha. e.g.  Implement a `YII_TEST` always succeed in `AltchaCaptchaValidator::validateAttribute` with specific string and introduce something `$I->solveAltcha(fieldName)`. 
3. Completely remove Captcha from A. Testing via  a `DummyCaptcha` provider
4. Make Altcha reliable running during acceptance tests. (not preferred) Probably newer Altcha versions could fail.
5. Make Captcha Provider configurable via DB Settings and switch Captcha Provider dynamic during tests. Probably the best solution, but then we need to introduce a Admin UI option somewhere. Maybe a new `Security` settings tab. Complex

ArchBlood:
> @marc-farre Hmm, problem is, that for the acceptance tests, we can currently only use one specific captcha driver defined in the configuration file. 
> 
> I think our options are:
> 
> 1. Switch to YiiCaptcha in the AcceptanceTests for now.  (my current preferred solution)
> 2. Create some option to force positive response from Altcha. e.g.  Implement a `YII_TEST` always succeed in `AltchaCaptchaValidator::validateAttribute` with specific string and introduce something `$I->solveAltcha(fieldName)`. 
> 3. Completely remove Captcha from A. Testing via  a `DummyCaptcha` provider
> 4. Make Altcha reliable running during acceptance tests. (not preferred) Probably newer Altcha versions could fail.
> 5. Make Captcha Provider configurable via DB Settings and switch Captcha Provider dynamic during tests. Probably the best solution, but then we need to introduce a Admin UI option somewhere. Maybe a new `Security` settings tab. Complex 
> 
> 

Normally wouldn't using `fixedVerifyCode` be enough? Otherwise extending this function would be a good idea but requires more work, maybe for now the first option should be adopted till a more solid method can be built?

```php
public function actions()
{
    return [
        'captcha' => [
            'class' => 'yii\captcha\CaptchaAction',
            'fixedVerifyCode' => YII_ENV_TEST ? 'testme' : null,
        ],
    ];
}
```

```php
// In your acceptance test (e.g., ExampleCest.php)
$I->fillField('form_field_name[captcha_attribute]', 'testme');
$I->click('Submit Button');
```

luke-:
> Normally wouldn't using `fixedVerifyCode` be enough? Otherwise extending this function would be a good idea but requires more work, maybe for now the first option should be adopted till a more solid method can be built?

Yes, that would be option 1, but this is not available for Altcha PoW

marc-farre:
@luke- I have created a new PR: https://github.com/humhub/humhub/pull/7740/
Closing this one.

PR #7746, opened 2025-09-27
fix is initial installation state

gevorgmansuryan:
https://github.com/humhub/humhub-internal/issues/837


gevorgmansuryan:
@luke- done

luke- (approved):

PR #7759, opened 2025-10-03
Flush cache after rebuild content search results

yurabakhtin:

- [x] Bugfix

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
https://github.com/humhub/humhub-internal/issues/843

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ HumHub Changelog
 - Enh #7753: New modal full screen size
 - Fix #7754: Tour module - the bottom part of the Profile image is hidden
 - Fix #7752: Fix content resetForm after form submit
-- Fix #7759: Flush cache after rebuild content search results
+- Fix #7759: Flush content search cache after update a record
 - Fix #7760: Content form menu text color same as background
 - Fix #7764: File Handler Dropdown menu items styling
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ HumHub Changelog
 - Enh #7753: New modal full screen size
 - Fix #7754: Tour module - the bottom part of the Profile image is hidden
 - Fix #7752: Fix content resetForm after form submit
+- Fix #7759: Flush cache after rebuild content search results
 
 1.18.0-beta.4 (September 19, 2025)
 ----------------------------------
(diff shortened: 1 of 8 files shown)


yurabakhtin:
> @yurabakhtin Can we somehow only flush the cache related parts?

@luke- The search results are cached by search request params [here](https://github.com/humhub/humhub/blob/develop/protected/humhub/modules/content/search/driver/AbstractDriver.php#L58), and we cannot know what request params were used to find a record, so we cannot know the cache key on deleting of some record, I see only one way is to clear all cache after search rebuilding.

luke-:
@yurabakhtin Can you please try to use something like: https://www.yiiframework.com/doc/api/2.0/yii-caching-tagdependency - We don't use anything like that yet, but since content is often reindexed, e.g. when a new comment is posted, we should using this.

yurabakhtin:
@luke- We could replace the [code](https://github.com/humhub/humhub/blob/develop/protected/humhub/services/MetaSearchService.php#L58) from:

```php
$data = Yii::$app->cache->getOrSet($cacheKey, fn() => $this->provider->getResults($this->pageSize), $this->cacheTimeout);
```

to:

```php
$data = Yii::$app->cache->get($cacheKey);
if ($data === false) {
    $results = $this->provider->getResults($this->pageSize);
    if ($this->provider instanceof ContentSearchProvider) {
        $contentTags = [];
        foreach ($results['results'] as $result) {
            $contentTags[] = 'meta-search-content-' . $result->content->id;
        }
        $tagDependency = new TagDependency(['tags' => $contentTags]);
    } else {
        $tagDependency = null;
    }
    Yii::$app->cache->set($cacheKey, $results, $this->cacheTimeout, $tagDependency);
}
```

this solution will solve only the case on deleting of a record because we know the record was tagged to a cache by Content ID, but it doesn't solve the case when a searching was done a record was created because we don't know all future Content IDs which will be created and what will be searched by some keyword.

luke-:
@yurabakhtin That's also a good idea. But I think the current case concerns newly created content. So I would simply reset the entire search cache.

yurabakhtin:
> So I would simply reset the entire search cache.

@luke- Ok, then I will add the tag `search-content` for all caches where it is used for searching a content records, and then reset it by the tag on create/update/delete a content record.
Should we do the same for space/user searching?

luke-:
@yurabakhtin Yes please!

luke- (approved):

PR #7780, opened 2025-10-14
Fix for PHP 8.4

yurabakhtin:

- [x] Bugfix

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**
I see errors on PHP 8.4.1 like:

`During class fetch: Uncaught yii\base\ErrorException: humhub\widgets\bootstrap\BootstrapVariationsTrait::light(): Implicitly marking parameter $label as nullable is deprecated`

`humhub\widgets\bootstrap\Button::asLink(): Implicitly marking parameter $label as nullable is deprecated, the explicit nullable type must be used instead`

I'd like to merge these fixes, also later I will do similar fixes for other modules which has min version 1.18.

### DIFF
--- a/protected/humhub/widgets/bootstrap/BootstrapVariationsTrait.php
+++ b/protected/humhub/widgets/bootstrap/BootstrapVariationsTrait.php
@@ -50,52 +50,52 @@ trait BootstrapVariationsTrait
         return self::light($text);
     }
 
-    public static function primary(string $label = null): static
+    public static function primary(?string $label = null): static
     {
         return static::instance($label, 'primary');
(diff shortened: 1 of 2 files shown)


luke-:
@yurabakhtin Thanks. I would have expected Rector to fix something like that. I ran it again, could you please check the PR?

https://github.com/humhub/humhub/pull/7781/files

yurabakhtin:
@luke- Related error from composer package:

scss

To reproduce: remove assets folders or save the page "Appearance Settings" `/admin/setting/design`

PR #7806, opened 2025-10-29
New live event for theme changing to reload the current page

yurabakhtin:

- [x] Feature

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
https://github.com/humhub/humhub-internal/issues/869

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ HumHub Changelog
 - Fix #7750: Mail summary may not sent for new users on higher intervals
 - Fix #7796: Administrator group cannot have group managers
 - Fix #7803: Fix number of pending space members
-- Enh #7806: New live event for theme changing to reload the current page
+- Enh #7806: New live event to prevent PJAX on the next click after theme changing
 
 1.17.3 (July 17, 2025)
 ----------------------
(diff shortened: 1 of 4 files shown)


protected/humhub/modules/live/live/ThemeChanged.php
```diff
@@ -0,0 +1,22 @@
...
+namespace humhub\modules\live\live;
+
+use humhub\modules\live\components\LiveEvent;
+
+/**
+ * Live event for theme changing
+ *
+ * @since 1.17.4
+ */
+class ThemeChanged extends LiveEvent
```
luke-: @yurabakhtin Since this event could be used in many other situations. 

e.g. "Assets flushed", "New Module installed", "Theme Changed" we should rename `PreventPjaxOnNextClick`.

It would be also good to pass an "$reason" which will be logged in the JS Debug.

yurabakhtin: @luke- Commit https://github.com/humhub/humhub/pull/7806/commits/1a933d3ba824ebcff8f73f9eaf9ebf950a6b8212:
prevent-pjax-log

PR #7831, opened 2025-11-13
Mercure push driver

yurabakhtin:

- [x] Feature

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
https://github.com/humhub/humhub-internal/issues/902

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -44,7 +44,8 @@
 #HUMHUB_CONFIG__COMPONENTS__DB__PASSWORD="secret"
 
 #--- Mailer (Overwrite)
-#HUMHUB_CONFIG__COMPONENTS__MAILER__DSN="smtp://localhost:25"
+#HUMHUB_FIXED_SETTINGS__BASE__MAILER_TRANSPORT_TYPE="dsn"
+#HUMHUB_FIXED_SETTINGS__BASE__MAILER_DSN="smtp://localhost:1025"
 
 #--- Various Settings
 #HUMHUB_CONFIG__HOME_URL="/s/welcome-space/"
--- a/protected/humhub/modules/live/driver/MercurePushDriver.php
+++ b/protected/humhub/modules/live/driver/MercurePushDriver.php
@@ -22,6 +22,7 @@ use Symfony\Component\Mercure\Jwt\LcobucciFactory;
 use Symfony\Component\Mercure\Update;
 use yii\base\InvalidConfigException;
 use Yii;
+use yii\helpers\Url;
 
 /**
  * Mercure Push driver for live events
@@ -31,7 +32,7 @@ use Yii;
  */
 class MercurePushDriver extends BaseDriver
 {
-    public string $hubUrl = 'https://localhost/.well-known/mercure';
+    public string $hubUrl = '';
     public string $jwtKeySubscriber = '';
     public string $jwtKeyPublisher = '';
     public string $topicPrefix = '/humhub/live/';
@@ -47,7 +48,7 @@ class MercurePushDriver extends BaseDriver
         parent::init();
 
         if (empty($this->hubUrl)) {
-            throw new InvalidConfigException('Mercure driver hub URL is not specified.');
+            $this->hubUrl = Url::to('/.well-known/mercure', true);
         }
 
         if (empty($this->jwtKeyPublisher) || empty($this->jwtKeySubscriber)) {


protected/humhub/modules/live/resources/js/humhub.live.mercure.js
```diff
@@ -11,19 +11,14 @@ humhub.module('live.mercure', function (module, require, $) {
...
-        const topics = this.options.topics || [];
-
-        topics.forEach(topic => {
-            sourceUrl += '&topic=' + encodeURIComponent(topic);
-        });
-
-        const source = new EventSource(sourceUrl);
+        const source = new EventSource(this.options.url
+            + '?jwt=' + encodeURIComponent(this.options.jwt)
+            + '&topic=' + encodeURIComponent(this.options.topic));
```
luke-: Do we need to provide the topics here again? There should be already in the JWT?

protected/humhub/modules/live/resources/js/humhub.live.mercure.js
```diff
@@ -0,0 +1,45 @@
...
+            url.searchParams.append('topic', topic);
+        });
+
+        const source = new EventSource(url.toString());
+
+        source.onerror = function (err) {
+            module.log.error('Mercure error', err);
+        };
+
+        source.onmessage = function (messageEvent) {
```
luke-, line 29: Can you please add some debugging for incoming messages here.

protected/humhub/modules/notification/resources/js/humhub.notification.js
```diff
@@ -69,6 +69,12 @@ humhub.module('notification', function (module, require, $) {
 
         var result = [];
         events.forEach(function (event) {
+            if (module.config.currentUserId && event.data.notificationOriginatorId
```
luke-: Do we still need this even with the topic change?

protected/humhub/modules/live/resources/js/humhub.live.mercure.js
yurabakhtin: @luke- As I understand from chatgpt and the doc https://mercure.rocks/spec the topic is required for subscription but the jwt is used only for authorization.

luke-: @yurabakhtin Ok, maybe if you don't pass this parameter, it uses all authorized topics. Accordingly to the doc you can also use a "*".

protected/humhub/modules/live/resources/js/humhub.live.mercure.js
yurabakhtin: @luke- https://github.com/humhub/humhub/pull/7831/commits/9c3bfc930a2a7d8be5b89ed66f5a719ba9f82079#diff-cdf25ab5495fa3d0dabc598494c82ab2b9819c266c1d14f7fd4cf1aa39ef0688R36

protected/humhub/modules/live/resources/js/humhub.live.mercure.js
yurabakhtin: @luke- I have tested with `topic=*`:

https://github.com/user-attachments/assets/e0d0c55b-fafe-49f0-bdb3-fddf5a4d5bce
As you can see the counter is updated even the user has no notification.

When topics are specified:
https://github.com/user-attachments/assets/1e3f5530-edd3-4abb-b40e-7dfa550ad56b

If the user is a member of the space:
https://github.com/user-attachments/assets/58fbcd1a-e9bc-4804-a987-c8cecb8f8866
It is strange why Sara(FireFox) and Admin(Chrome) has the counter increased twice, however Mercure(Opera) case works more correctly.

protected/humhub/modules/notification/resources/js/humhub.notification.js
yurabakhtin: @luke- Please see how it works without the filter without excluding originator(i.e. if I remove that my code):

1 - Mercure(Opera) | 2 - Sara(FireFox) | 3 - Yura/admin (Chrome)
https://github.com/user-attachments/assets/391195f4-dbc7-44a4-a6b2-713b5e909fcf

For example, when I am admin/Yura click "Like" I see the notification counter is updated also for me(admin), but if I go to review the new notification I don't see it because really it was only for Sara.
I have such issues on FireFox and Chrome, so I have implemented the filter to avoid the wrong counter increasing.
However I don't see such issue when I click "Like" from Opera by Mercure user.

But I don't think it depends on browser, because I have done one more test and I see the counter is updated incorrectly only when I vote like by Sara or Admin, but it works well when I vote by new created user Mercure or vote to Mercure user by other users, even if I switch browsers for Sara and Mercure:

1 - Sara(Opera) | 2 - Mercure(FireFox) | 3 - Yura/admin (Chrome)
https://github.com/user-attachments/assets/953cb2c4-1828-4d62-a93c-0ee0c295d781

luke-: @yurabakhtin I think the main problem here is that the user “Admin Yura User” is incorrectly receiving the ‘NewNotification’ event from “Sara Schuster.”

Image

We should look for the underlying problem here, rather than skipping the information in the front end via JS.

---

Probably the Admin user is following or friends with Sara?

---

Probably we need to refactor the channels/user legtimation handling, to ensure we have one private channel/topic to a specific user. 
https://staff.humhub.org/s/development/wiki/468/live-messaging#channels

yurabakhtin: > Probably the Admin user is following or friends with Sara?

@luke- Thanks for the hint, yes the 2 users were followed, please see my test when I unfollowed the Admin from Sara:
(and I removed the filter `module.config.currentUserId === event.data.notificationOriginatorId)` for the current test)
following

https://github.com/user-attachments/assets/c5969352-78c5-4b9d-a96e-6b0ed99f7629

Here we see the fake counter inscreasing only for Sara because she is following to Admin.
But Admin doesn't have the fake counter because it was unfollowed from Sara before this test.

PR #7836, opened 2025-11-17
User DisplayName - Whitespace Only

gevorgmansuryan:
https://github.com/humhub/humhub-internal/issues/908

### DIFF
--- a/protected/humhub/modules/space/models/Space.php
+++ b/protected/humhub/modules/space/models/Space.php
@@ -119,7 +119,6 @@ class Space extends ContentContainerActiveRecord
     {
         $rules = [
             [['join_policy', 'visibility', 'status', 'sort_order', 'auto_add_new_members', 'default_content_visibility'], 'integer'],
-            [['name'], 'trim'],
             [['name'], 'required'],
             [['name'], 'string', 'max' => 45, 'min' => 2],
             [['description', 'about', 'color'], 'string'],


protected/humhub/components/validators/RequiredValidator.php
```diff
@@ -1,12 +1,15 @@
...
-namespace humhub\libs;
+namespace humhub\components\validators;
 
 use yii\helpers\Json;
 use yii\validators\ValidationAsset;
 
 class RequiredValidator extends \yii\validators\RequiredValidator
 {
+    /**
+     * @inerhitdoc
```
luke-: Typo

protected/humhub/components/validators/RequiredValidator.php
```diff
@@ -1,12 +1,15 @@
...
+     */
     public function isEmpty($value)
     {
         if (empty(preg_replace('/[\p{Z}\s]+/u', '', $value))) {
```
luke-: Did you tested this validator with some other field types , e.g. Numeric, Bool, File, etc?

protected/humhub/components/validators/TrimValidator.php
```diff
@@ -1,14 +1,12 @@
...
 
 use yii\helpers\Json;
 use yii\validators\ValidationAsset;
 
 class TrimValidator extends \yii\validators\TrimValidator
 {
-    public $skipOnEmpty = false;
-
     /**
      * @inerhitdoc
```
luke-: Typo

protected/humhub/modules/space/models/Space.php
```diff
@@ -119,6 +119,7 @@ public function rules()
     {
         $rules = [
             [['join_policy', 'visibility', 'status', 'sort_order', 'auto_add_new_members', 'default_content_visibility'], 'integer'],
+            [['name'], 'trim'],
```
luke-: Should the required validator here suffient?

gevorgmansuryan: its another problem. it will be possible to create space with name `              a` without trim

protected/humhub/components/validators/TrimValidator.php
gevorgmansuryan: fixed

protected/humhub/components/validators/RequiredValidator.php
gevorgmansuryan: added string check, added new unit tests

protected/humhub/modules/space/models/Space.php
luke-: @gevorgmansuryan I know there are some users using this kind of hack, to move some Spaces to the top in the Space navigation. So I would not add a trim at the moment. 

As soon we have some more sorting options we should add this trim. 

Can you please remove it for now?

gevorgmansuryan: @luke- got it. done

PR #7850, opened 2025-11-26
Fix/required validator

gevorgmansuryan:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**

### DIFF
--- a/protected/humhub/components/validators/RequiredValidator.php
+++ b/protected/humhub/components/validators/RequiredValidator.php
@@ -13,8 +13,8 @@ class RequiredValidator extends \yii\validators\RequiredValidator
      */
     public function isEmpty($value)
     {
-        if (is_string($value) && preg_match(self::REGEX, $value)) {
-            $value = preg_replace(self::REGEX, '', $value);
+        if (is_string($value) && empty(preg_replace(self::REGEX, '', $value))) {
+            return true;
         }
 
         return parent::isEmpty($value);


luke-:
@gevorgmansuryan I don't think the required validator should manipulate the value. It should only check whether the value consists solely of spaces; otherwise, it should not make any changes to the content.

gevorgmansuryan:
@luke- its not changing/manipulating the original value

luke-:
@gevorgmansuryan Ok, but wouldn't it sufficient to check it contains only of Whitespace Chars? Instead it looks like we detect them everywhere?

gevorgmansuryan:
@luke- current regex removes all whitespace characters only. but in unicode mode, so all non unicode chars are considered invalid. So with new implementation it will not touch non unicode strings

luke-:
@gevorgmansuryan Can we please change the implementation and not remove any characters. Only check if the string only contains of Whitespace chars.

If a form wants to remove these chars, we can use trim.

gevorgmansuryan:
@luke- ready

protected/humhub/components/validators/RequiredValidator.php
```diff
@@ -7,18 +7,14 @@
...
     public function isEmpty($value)
     {
-        if (
-            is_string($value) && !( // check if string and not contains binary data
-            str_contains($value, "\0") && // check if contains NUL
-            !mb_check_encoding($value, 'UTF-8') && // check if invalid UTF-8 encoding
-            !!preg_match('/[^\x09\x0A\x0D\x20-\x7E]/', $value)) // check for binary control chars
-        ) {
-            $value = preg_replace('/[\p{Z}\s]+/u', '', $value);
+        if (is_string($value) && preg_match(self::REGEX, $value)) {
```
luke-: @gevorgmansuryan I would prefer something like and don't manipulate the value (e.g. file content). 

``` 
if (empty(preg_replace('/[\p{Z}\s]+/u', '', $value))) {
     return true;
}
```

PR #7862, opened 2025-11-29
Installer: Redirect loop on first access in web browser

gevorgmansuryan:
https://github.com/humhub/humhub-internal/issues/930

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,16 @@ HumHub Changelog
 - Fix #7847: Fixed missing translations in `Admin/People` page
 - Fix #7854: Tour for Spaces if no visible and active Space
 - Fix #7862: Fix redirect loop on `ErrorAction`
+- Fix #7860: Fix type error in installer
+- Enh #7852: New `humhub\helpers\ScssHelper::getColorContrast()` for text color in emails, to contrast with the button background when customized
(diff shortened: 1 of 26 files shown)


protected/humhub/controllers/ErrorController.php
```diff
@@ -22,6 +23,8 @@
  */
 class ErrorController extends Controller
 {
+    public $access = ControllerAccess::class;
```
luke-, line 23: Not sure here. The error controller uses the standard layout with navigation, etc. This should not be displayed for installations without guest access, even in the event of an error. Is there no other way to solve the installer problem?
 
Otherwise, we could possibly try to set the layout of the error controller dynamically.
 .

luke- (changes requested):

protected/humhub/controllers/ErrorController.php
gevorgmansuryan: @luke- then we need other error action for this cases, just to show 'internal server error'

luke-: @gevorgmansuryan Can you try to switch the layout of the error controller if the user is not logged in?

gevorgmansuryan: @luke- yes

PR #7913, opened 2025-12-20
js translations

gevorgmansuryan:
https://github.com/humhub/humhub-internal/issues/986


static/js/humhub/humhub.i18n.js
```diff
@@ -0,0 +1,88 @@
+humhub.module('i18n', function(module, require, $) {
+    var IntlMessageFormat = require('intl-messageformat');
+
+    var locale = detectLocale();
+    var globalMessages = {};
+    var loadedCategories = new Set();
+    var pendingLoads = new Map();
+    var compiledCache = new Map();
+
+    function detectLocale() {
```
luke-: @gevorgmansuryan Looks nice.

We should pass the language via the config:
https://github.com/humhub/humhub/blob/develop/protected/humhub/widgets/CoreJsConfig.php#L41

---

Once we are able to extract the language string from JS, we can clean up CoreJS here and remove all languages and replace them with `humhub.i18n.t`.

luke- (changes requested):

static/js/humhub/humhub.i18n.js
gevorgmansuryan: @luke- sure, also translations load url too

luke-:
@gevorgmansuryan Would be cool, if you could migrate some (for the first step just a few) JS translations from here

https://github.com/humhub/humhub/blob/next/protected/humhub/widgets/CoreJsConfig.php#L51-L54

into to the new JS Translation feature.

gevorgmansuryan:
@luke- done, ready to review. Migrated to new i18n for `spaceChooser` and `client` modules

package.json
```diff
@@ -14,5 +14,8 @@
     "grunt-contrib-less": "^2.0.0",
     "grunt-contrib-uglify": "^5.0",
     "grunt-shell": "^2.1.0"
+  },
+  "dependencies": {
+    "intl-messageformat": "^11.1.0"
```
luke-: @gevorgmansuryan Any reason to not include this as other assets via Composer?

protected/humhub/modules/space/resources/js/humhub.space.chooser.js
```diff
@@ -15,12 +15,19 @@ humhub.module('space.chooser', function (module, require, $) {
...
 
     var SELECTOR_ITEM = '[data-space-chooser-item]';
     var SELECTOR_ITEM_REMOTE = '[data-space-none],[data-space-archived]';
 
     var SpaceChooser = Widget.extend();
 
+    const init = function() {
+        SpaceChooser.instance($('#space-menu-dropdown'));
+
+        return i18n.preload('SpaceModule.chooser');
```
luke-: @gevorgmansuryan Do you see any chance, to skip this additional step?

Would be great when `i18n.t` blocks, until the the messages are loaded.

package.json
gevorgmansuryan: To be used in react/vue

protected/humhub/modules/space/resources/js/humhub.space.chooser.js
gevorgmansuryan: Yes,we need to define all translations in variables in init, but did not like this idea

package.json
luke-: Hmm, can you please ensure for the current PR that it runs via Composer and Assets.

There will be various challenges when dependencies are also located elsewhere, such as in `node_modules`. Not sure how we can handle this in future.

protected/humhub/modules/space/resources/js/humhub.space.chooser.js
luke-: Do you see any other option here?

protected/humhub/resources/js/humhub/humhub.client.js
```diff
@@ -379,8 +380,10 @@ humhub.module('client', function (module, require, $) {
 
         $form.data('state', serializeFormState($form));
 
-        msg = msg || module.text('warn.onBeforeLoad');
```
luke-, line 382: Can you please also remove these module texts?

protected/humhub/modules/space/resources/js/humhub.space.chooser.js
gevorgmansuryan: @luke- unfortunately, I don’t really see a viable way to skip this step.
In JS modules the init() method is executed first, and there is no native way to synchronously block execution until async translations is loaded. Returning a Promise from init() is currently the only clean and predictable way to ensure the whole module waits until the message category is fully loaded before any translations are used.

protected/humhub/resources/js/humhub/humhub.client.js
gevorgmansuryan: @luke- done

package.json
gevorgmansuryan: it seems there is a lot of problems in https://asset-packagist.org/, latest composer can not find packages right now. Will try again after some time

protected/humhub/assets/IntlMessageFormatAsset.php
```diff
@@ -0,0 +1,17 @@
...
+
+use humhub\components\assets\AssetBundle;
+use humhub\components\View;
+
+class IntlMessageFormatAsset extends AssetBundle
+{
+    public $defaultDepends = false;
+    public $defer = false;
+    public $jsPosition = View::POS_HEAD;
+    public $sourcePath = '@app/../node_modules/intl-messageformat';
```
luke-: Can you please use this composer dependency:

https://asset-packagist.org/package/npm-asset/intl-messageformat

gevorgmansuryan: @luke- https://github.com/humhub/humhub/pull/7913#discussion_r2706039632

but seems its working now

gevorgmansuryan: @luke- done

protected/humhub/modules/space/resources/js/humhub.space.chooser.js
luke-: @gevorgmansuryan Hmm, ok. 

- Then it would make sense to drop a warning, I've a module wants to use translate without a preloaded category?

- The translate method, currently tries to load a message category if not available yet. https://github.com/humhub/humhub/pull/7913/changes#diff-462cce2e63280f40ba3f72be860f9bc37e7815fdc7e012e44456a4cf9f8dc88cR79 I'm not fully understanding this. Why there is loading code but it seems not possible to load?

package.json
luke-: Ok. Let me know when I should check.

protected/humhub/modules/space/resources/js/humhub.space.chooser.js
gevorgmansuryan: @luke- it works as fallback, it will try to load category translation and will save it in localStorage, but will start work after next page load, as it can not block js execution at this point. 
Warning added when when called without preloading

protected/humhub/controllers/I18nController.php
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Used to fetch translation messages from UI.
+ *
+ * @since 1.18
+ */
+class I18nController extends Controller
+{
+    public $access = ControllerAccess::class;
+
+    public function actionTranslations(string $category): Response
```
luke-, line 27: Probably it would be good here, to also support, loading multiple message files at once?

gevorgmansuryan: @luke- makes sense, will be

protected/humhub/modules/space/resources/js/humhub.space.chooser.js
luke-: @gevorgmansuryan I would prefer that if the category has not yet been loaded, an error is always thrown and the English language is returned.

Otherwise, the behavior here is a bit random if only the first call does not work.

luke-:
@gevorgmansuryan Can you please add some AccentanceTests here (JS Translation UI) and Unit Tests (Message Extract).

package.json
luke-: @gevorgmansuryan Can you please remove the package here for now?

static/js/humhub/humhub.intl-messageformat.js
```diff
@@ -0,0 +1,3 @@
+humhub.module('intl-messageformat', function(module, require, $) {
```
luke-: @gevorgmansuryan Can't we simply use it inside the humhub.i18n. Do we really need a seperate module here?

protected/humhub/resources/js/humhub/humhub.i18n.js
```diff
@@ -0,0 +1,121 @@
+humhub.module('i18n', function(module, require, $) {
+    var IntlMessageFormat = require('intl-messageformat');
+
+    var locale = module.config.language;
+    var globalMessages = {};
+    var loadedCategories = new Set();
+    var pendingLoads = new Map();
```
luke-, line 10: Are the message files stored somewhere in locale storage or just per request?

luke- (changes requested):
@gevorgmansuryan Please see my remarks

static/js/humhub/humhub.intl-messageformat.js
gevorgmansuryan: @luke- its something like wrapper for IntlMessageFormat in humhub module way, so it can be used with humhub's require, so in future here can be some global configs for IntlMessageFormat. but can be removed and used directly

protected/humhub/resources/js/humhub/humhub.i18n.js
gevorgmansuryan: @luke- yes, its stored in localStorage with key = `'humhub.i18n.' + version + '.' + language + '.' + category`
in
- static/js/humhub/humhub.i18n.js:80
- static/js/humhub/humhub.i18n.js:48

luke-:
@gevorgmansuryan 

- Seems the `translate` method, still loads `loadTranslations()`. Please remove this, if it's not working on the first method call.
- Once we migrate more translation to JS, the initial HumHub request, needs to load several translation files. To avoid this, it would be good, when we could define some `preloadCategories` directly in `humhub.i18n`. These should be combined and loaded with one request.
- Please remove `humhub.intl-messageformat` for now. Either integrate it to the i18n module or use it directly.
- I've created a new revision to detect potential changes to the translation messages. https://github.com/humhub/humhub/pull/7913/changes/b21b63652307e84db6a34d5f289bb7305e31d220 - Please store this revision somewhere in local storage,  when changed, remove the old translations and load the latest. 
- Currently the tests are failing https://github.com/humhub/humhub/actions/runs/21762566138/job/62789761831
- Do you see any (potential) problems with the current implementation? E.g., browser support, loading time, etc.
- Does the acceptance test (User3) currently use a different language or only `en`? It would be good to check an actual translation as well. 
- I'm unsure whether I want to change the `init` method of the standard HumHub module API. Alternatively, could we add an attribute to modules, “requiredI18nCategories,” which are loaded when defined?

**After** clarifying the above points, I would like to convert the entire HumHub Core to the new `i18n.t`. Please wait until the points have been clarified and approved. Otherwise, the PR will become confusing.

gevorgmansuryan:
@luke- 

> - Seems the `translate` method, still loads `loadTranslations()`. Please remove this, if it's not working on the first method call.

Fixed

> - Once we migrate more translation to JS, the initial HumHub request, needs to load several translation files. To avoid this, it would be good, when we could define some `preloadCategories` directly in `humhub.i18n`. These should be combined and loaded with one request.

i18n will try to load all preloaded categories with one request now

> - Please remove `humhub.intl-messageformat` for now. Either integrate it to the i18n module or use it directly.

done

> - I've created a new revision to detect potential changes to the translation messages. https://github.com/humhub/humhub/pull/7913/changes/b21b63652307e84db6a34d5f289bb7305e31d220 - Please store this revision somewhere in local storage,  when changed, remove the old translations and load the latest. 

revision is used in building storageKey now

> - Currently the tests are failing https://github.com/humhub/humhub/actions/runs/21762566138/job/62789761831

Fixed, all tests are ok now

> - Do you see any (potential) problems with the current implementation? E.g., browser support, loading time, etc.

Can be problematic only on old legacy browsers, eg. internet explorer / old safari

> - Does the acceptance test (User3) currently use a different language or only `en`? It would be good to check an actual translation as well. 

It's testing with EN an DE languages now

> - I'm unsure whether I want to change the `init` method of the standard HumHub module API. Alternatively, could we add an attribute to modules, “requiredI18nCategories,” which are loaded when defined?

Good idea, Done. implemented in space.chooser

luke-:
@gevorgmansuryan 

> > * Once we migrate more translation to JS, the initial HumHub request, needs to load several translation files. To avoid this, it would be good, when we could define some `preloadCategories` directly in `humhub.i18n`. These should be combined and loaded with one request.
> 
> i18n will try to load all preloaded categories with one request now
> 
Great. So it checks all modules for `requiredI18nCategories` and preloads them together?

> > * I've created a new revision to detect potential changes to the translation messages. [b21b636](https://github.com/humhub/humhub/commit/b21b63652307e84db6a34d5f289bb7305e31d220) - Please store this revision somewhere in local storage,  when changed, remove the old translations and load the latest.
> 
> revision is used in building storageKey now
> 

Is there any cleanup of old revisions?

> > * Do you see any (potential) problems with the current implementation? E.g., browser support, loading time, etc.
> 
> Can be problematic only on old legacy browsers, eg. internet explorer / old safari

Can be "localStorage" somehow disabled on some Browsers? Is there a handling for such cases? e.g. Skip Storage?
> 
> > * I'm unsure whether I want to change the `init` method of the standard HumHub module API. Alternatively, could we add an attribute to modules, “requiredI18nCategories,” which are loaded when defined?
> 
> Good idea, Done. implemented in space.chooser

:+1: 

- Just to understand JS modules can now implement, `module.requiredI18nCategories = ['SpaceModule.chooser'];`, when this is set, the categories will be automatically loaded on requests or module init. No more JS module modifications needed?
- Is it now possible to revert the previous `init` modification, that the module init can be now async/sync? 
- The SpaceChooser has now changes, I didn't understand can you please explain:
  - https://github.com/humhub/humhub/blob/9c52648c6e79d230563c45637d8c0260ba86836f/protected/humhub/modules/space/resources/js/humhub.space.chooser.js#L27-L29
  - https://github.com/humhub/humhub/blob/9c52648c6e79d230563c45637d8c0260ba86836f/protected/humhub/modules/space/resources/js/humhub.space.chooser.js#L84-L100

gevorgmansuryan:
@luke- 

> Great. So it checks all modules for requiredI18nCategories and preloads them together?

Yes

> Is there any cleanup of old revisions?

Yes

> Can be "localStorage" somehow disabled on some Browsers? Is there a handling for such cases? e.g. Skip Storage?

Yes, that’s possible. It can be unavailable in some environments like Safari's private mode, strict privacy settings, or certain WebViews/iframes. In those cases it may exist but still throw an exception on write.
Currently I’m already wrapping all localStorage access in try/catch. If anything fails, we simply skip caching and keep translations for the current session, So app will continue to work correctly. So we will re-fetch translations on every page load, but there’s no functional impact.

> Just to understand JS modules can now implement, module.requiredI18nCategories = ['SpaceModule.chooser'];, when this is set, the categories will be automatically loaded on requests or module init. No more JS module modifications needed?

Yes, just by adding module.requiredI18nCategories = ['SpaceModule.chooser', '...', ...] in eevery module it'll load all needed categories 

> Is it now possible to revert the previous init modification, that the module init can be now async/sync?

Yes, it can be reverted, but I’d prefer to keep async init() support. Originally I made init() sync-or-async intentionally- if it returns a Promise, the module waits for it. Translations were one use case, and now requiredI18nCategories covers that, but async init is still useful for other hard prerequisites (eg fetching config, preloading data, waiting for dependencies).
I’d suggest keeping it as-is, It’s already [documented](https://github.com/humhub/humhub/pull/7913/changes#diff-e58e8662213684a9f6b13971c8a8560d61b19ff2fe72fbe0b4a2abb5d82c2c08R82) and fully backward-compatible, modules that don’t return a Promise behave exactly as before. What to you think?

> The SpaceChooser has now changes, I didn't understand can you please explain: 

The changes are related to the new acceptance test I added. Its mainly checking JS translations, and a few small adjustments were needed so the test can run correctly. There’s no functional change in behavior.

protected/humhub/modules/space/resources/js/humhub.space.chooser.js
```diff
@@ -464,8 +491,6 @@ humhub.module('space.chooser', function (module, require, $) {
 
     module.export({
         SpaceChooser: SpaceChooser,
-        init: function () {
-            SpaceChooser.instance($('#space-menu-dropdown'));
-        }
+        init: init
```
luke-, line 494: This change is just a cleanup? Or is this necessary somehow?

protected/humhub/widgets/CoreJsConfig.php
```diff
@@ -48,10 +48,13 @@ public function run()
                     'baseUrl' => Yii::$app->settings->get('baseUrl'),
                     'reloadableScripts' => CacheHelper::getReloadableScriptUrls(),
                     'cspViolationReloadInterval' => Security::CSP_VIOLATION_RELOAD_INTERVAL,
-                    'text' => [
-                        'warn.onBeforeLoad' => Yii::t('base', 'Unsaved changes will be lost. Do you want to proceed?'),
-                        'warn.onBeforeCloseModal' => Yii::t('base', 'Unsaved changes will be lost. Do you want to proceed?'),
-                    ],
+                ],
+                'i18n' => [
+                    'debug' => YII_DEBUG,
```
luke-, line 53: Probably it would be cleaner, if we could put the Debug State and Revision somewhere more neutral? Probably other JS modules need this value too?

protected/humhub/widgets/CoreJsConfig.php
```diff
@@ -48,10 +48,13 @@ public function run()
...
+                    'revision' => Yii::$app->systemRevision->getPublicSignature(),
+                    'language' => Yii::$app->language,
+                    'translationUrl' => Url::to(['/i18n/translations']),
+                    'version' => Yii::$app->version,
```
luke-: Do we still need version?

static/js/humhub/humhub.core.js
```diff
@@ -195,7 +199,7 @@ var humhub = humhub || (function ($) {
             initialModules.push(instance);
         } else { // Init modules added asynchronously (ajax/pjax)
             addModuleLogger(instance);
-            initModule(instance);
+            initModuleWithRequiredI18n(instance);
```
luke-: Somehow, this “initWithXyz” wrapper looks a bit strange for me. Isn't there a other probably cleaner way to do this?

I'm no JS expert, but maybe integrating it into `initModule` or something like `beforeInit`?

static/js/humhub/humhub.core.js
```diff
@@ -579,16 +583,52 @@ var humhub = humhub || (function ($) {
             return 1;
         });
 
-        $.each(initialModules, function (i, module) {
-            initModule(module);
+        preloadRequiredI18n(initialModules).then(function() {
```
luke-: Similar like the `initWithXyz` comment, somehow it's strange to have `i18n` wrappers. Maybe also something like `beforeInit`?

protected/humhub/modules/space/resources/js/humhub.space.chooser.js
gevorgmansuryan: @luke- it was 
```js
init: function () {
      SpaceChooser.instance($('#space-menu-dropdown'));
}
```
befire.
i just moved it to `init`

protected/humhub/widgets/CoreJsConfig.php
gevorgmansuryan: @luke- Yes, we can but I think we should still keep debug as a separate because revision only changes the cache key, while debug should disable the local cache completely.

protected/humhub/widgets/CoreJsConfig.php
gevorgmansuryan: Agree, removed

protected/humhub/widgets/CoreJsConfig.php
luke-: @gevorgmansuryan I was thinking to move these options into a JS "Core" module, so instead of `humhub.i18n.debug|systemRevision`, something like `humhub.core.debug|systemRevision`.

gevorgmansuryan: @luke- Yes, that makes sense, will try

luke- (approved):

PR #7938, opened 2026-01-15
Use current user as originator for mentioning notification

yurabakhtin:

- [x] Bugfix

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
https://github.com/humhub/humhub-internal/issues/1013

### DIFF
--- a/protected/humhub/modules/user/models/Mentioning.php
+++ b/protected/humhub/modules/user/models/Mentioning.php
@@ -12,8 +12,8 @@ use humhub\components\ActiveRecord;
 use humhub\components\behaviors\PolymorphicRelation;
 use humhub\modules\content\components\ContentActiveRecord;
 use humhub\modules\content\components\ContentAddonActiveRecord;
-use humhub\modules\content\models\Content;
 use humhub\modules\user\notifications\Mentioned;
+use Yii;
 use yii\base\Exception;
 use yii\base\InvalidArgumentException;
 use yii\db\ActiveQuery;
@@ -87,11 +87,11 @@ class Mentioning extends ActiveRecord
      */
     private function getOriginatorBySource($source)
     {
-        if ($source instanceof ContentActiveRecord) {
-            /** @var ContentActiveRecord $source */
+        if (!Yii::$app->user->isGuest) {
+            return Yii::$app->user->getIdentity();
+        } elseif ($source instanceof ContentActiveRecord) {
             return $source->content->createdBy;
         } elseif ($source instanceof ContentAddonActiveRecord) {
-            /** @var ContentAddonActiveRecord $source */
             return $source->user;
         }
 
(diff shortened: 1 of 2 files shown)


protected/humhub/modules/user/models/Mentioning.php
```diff
@@ -87,11 +87,11 @@ public function afterSave($insert, $changedAttributes)
      */
     private function getOriginatorBySource($source)
     {
-        if ($source instanceof ContentActiveRecord) {
-            /** @var ContentActiveRecord $source */
+        if (!Yii::$app->user->isGuest) {
+            return Yii::$app->user->getIdentity();
```
luke-, line 91: @yurabakhtin Hmm I would prefer to use somehow the user from the `source` and not using always the currently logged in user.

Is the `source` the WikiPage revision? Maybe we could use the user from this model?

yurabakhtin: @luke- The mentioning feature is a RichText Extension.
For Wiki module it is started from [here](https://github.com/humhub/wiki/blob/master/models/forms/PageEditForm.php#L256):

- `WikiRichText::postProcess($this->revision->content, $this->page);`

We could change the wiki module code from `$this->page` to `$this->revision` and then change the core method `Mentioning::getOriginatorBySource()` by adding:

```php
elseif ($source instanceof ActiveRecord && property_exists($source, 'user_id')) {
    return User::findOne($source->user_id);
}
```

because the `WikiPageRevision` is `humhub\components\ActiveRecord`.
I understand the using `property_exists` is not good here, better is to use some interface.

but currently the core Mentioning Extension is restricted only to `ContentActiveRecord` and `ContentAddonActiveRecord`:
- https://github.com/humhub/humhub/blob/master/protected/humhub/modules/content/widgets/richtext/AbstractRichText.php#L21
- https://github.com/humhub/humhub/blob/master/protected/humhub/modules/content/widgets/richtext/extensions/mentioning/MentioningExtension.php#L101-L104
- https://github.com/humhub/humhub/blob/master/protected/humhub/modules/user/models/Mentioning.php#L48
- https://github.com/humhub/humhub/blob/master/protected/humhub/modules/user/models/Mentioning.php#L114
- https://github.com/humhub/humhub/blob/master/protected/humhub/modules/user/models/Mentioning.php#L136-L138

I think we have the restriction because of the [code](https://github.com/humhub/humhub/blob/master/protected/humhub/modules/user/models/Mentioning.php#L141) `$record->content->getStateService()->isPublished()` to know the record is published.
I am not sure we should modify the `WikiPageRevision` to Content Record.

luke-: @yurabakhtin With v.19, I have already cleaned up the `ContentAddonActiveRecord` and introduced a `ContentProvider` interface.

We should make sure that all dependent records, such as `WikiPageRevision`, implement a `ContentAddon`. Then the mentioning should work properly here.

I would therefore put the issue/PR on hold for now.

PR #8014, opened 2026-02-10
Module - Message Command - Vendor Folder

gevorgmansuryan:
https://github.com/humhub/humhub-internal/issues/1071


luke-:
@gevorgmansuryan Thanks, can you please merge it into `master`.

Probably it's also fine to add it here `/vendor`, no problem when it's also used by the core. https://github.com/humhub/humhub/blob/master/protected/humhub/config/i18n.php#L90

gevorgmansuryan:
@luke – For core, do we need to use/export translations from the yii category, or possibly some from Kartik widgets?

luke-:
@gevorgmansuryan No, we can exclude the Vendor Folder also from core. But please use full paths e.g. /vendor and /protected/vendor

PR #8029, opened 2026-02-17
Display and cache virtual profile field in the user name subtitle

yurabakhtin:

- [x] Bugfix
- [x] Feature

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
https://github.com/humhub/humhub-internal/issues/1060


protected/humhub/modules/user/models/User.php
```diff
@@ -1046,4 +1050,12 @@ public function getPasswordRecoveryService(): PasswordRecoveryService
     {
         return new PasswordRecoveryService($this);
     }
+
+    /**
+     * @since 1.18.1
+     */
+    public function flushCache(): void
```
luke-: I'm not sure whether we should introduce a new `User::flushCache`. 

This only makes sense if we expect to have other user caches in the future, but then it's unclear whether we always want to delete the entire user caches.

So probably it's cleaner to introduce a BaseTypeVirtual::flushCache(User $user): void, then it's more clear what this invalidation does?

yurabakhtin: Done https://github.com/humhub/humhub/pull/8029/changes/fecf1ee0e26fd041f948f5172aba3a65f4f34368

PR #8032, opened 2026-02-20
Improve menu link adding

yurabakhtin:

- [x] Feature

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
We have many code in the array format like this:

```php
$this->addEntry(new MenuLink([
  'id' => 'spaces',
  'label' => Yii::t('AdminModule.base', 'Spaces'),
  'url' => ['/admin/space'],
  'icon' => 'dot-circle-o',
  'sortOrder' => 400,
  'isActive' => ControllerHelper::isActivePath('admin', 'space'),
  'isVisible' => Yii::$app->user->can([
    ManageSpaces::class,
    ManageSettings::class,
  ]),
]));
```

I prefer to apply options by methods as we have for bootstrap Button, so new code will be:

```php
$this->addLink(Yii::t('AdminModule.base', 'Spaces'))
    ->link(['/admin/space'])
    ->id('spaces')
    ->icon('dot-circle-o')
    ->sortOrder(400)
    ->active('admin', 'space')
    ->visible(Yii::$app->user->can([ManageSpaces::class, ManageSettings::class]));
```

I have started this only for the widget `AdminMenu`, but it is used in almost all modules, if you will approve it then I will do this modification when I work on this https://github.com/humhub/humhub-internal/issues/1042.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,6 @@ HumHub Changelog
 - Fix #8039: Encode `Button`, `Link`, `Badge` and `DropdownMenu` labels by default to fix potential XSS security issue (see [migration guide](https://github.com/humhub/humhub/blob/master/MIGRATE-DEV.md#version-1181) for details)
 - Fix #8038: Improve paragraph and image styles in RichText content
 - Enh #7978: Show `DisplayNameSub` in UserPicker
->>>>>>> develop
 
 1.18.0 (February 4, 2026)
 -------------------------


luke-:
@yurabakhtin I'm a bit ambivalent about method chaining. On the one hand, it's very nice, but on the other hand, it's confusing when there are too many options and it's unclear which ones are required and which ones are optional. 

In general, imo it would be best to enforce mandatory (or very typical) attributes via the constructor/create method and make "optional" attributes available via method chaining.  

```php
$this->addEntry(MenuLink::create(Yii::t('AdminModule.base', 'Spaces'), ['/admin/space'])
    ->id('spaces')
    ->icon('dot-circle-o')
    ->sortOrder(400)
    ->active('admin', 'space')
    ->visible(Yii::$app->user->can([ManageSpaces::class, ManageSettings::class]));
```

The second route parameter `MenuLink::create(string $label, ?string|array $route)` should also be allowed to be null here, but should have to be explicitly set to `null`.

What do you think about this change?

yurabakhtin:
@luke- Ok, for me it is enough too, main idea is to replace the array code to functions, because IDE suggests what methods are allowed for the object when I write `->` but it is not fast to go to inside the class and find what properties the class has when it is used as array options.

Maybe we should use the method name as `instance` instead of `create` because I see this name is used more often.

> The second route parameter `MenuLink::create(string $label, ?string|array $route)` should also be allowed to be null here, but should have to be explicitly set to `null`.

I see `MenuLink` is always used with URL, (except of [this single case](https://github.com/humhub/humhub/blob/develop/protected/humhub/modules/user/widgets/AccountTopMenu.php#L52-L55) where property `link` is directly built as button, but we can keep it as is, without changing) so we can make the second param `$route` as mandatory `string|array`. Or when do you think it can be used with `null`?

luke-:
> @luke- Ok, for me it is enough too, main idea is to replace the array code to functions, because IDE suggests what methods are allowed for the object when I write `->` but it is not fast to go to inside the class and find what properties the class has when it is used as array options.

:+1: 

> 
> Maybe we should use the method name as `instance` instead of `create` because I see this name is used more often.
> 

`::instance` is fine too! 

> > The second route parameter `MenuLink::create(string $label, ?string|array $route)` should also be allowed to be null here, but should have to be explicitly set to `null`.
> 
> I see `MenuLink` is always used with URL, (except of [this single case](https://github.com/humhub/humhub/blob/develop/protected/humhub/modules/user/widgets/AccountTopMenu.php#L52-L55) where property `link` is directly built as button, but we can keep it as is, without changing) so we can make the second param `$route` as mandatory `string|array`. Or when do you think it can be used with `null`?

Maybe there are some JS links, so I think we should allow `null` to avoid problems in future.

yurabakhtin:
@luke- New changes https://github.com/humhub/humhub/pull/8032/changes/a91a2feb634fb888e658d6e522557b24875e8c10.
Please note one case where I decided to define URL in the separate method `url()` instead of the second param because the link with title is long, and maybe. we could use this way for all others, i.e. one option is on single line:

```php
        $this->addEntry(MenuLink::instance(Yii::t('AdminModule.base', 'Modules') . $this->getMarketplaceUpdatesBadge())
            ->url(['/admin/module'])
```

---

If the method name `MenuLink::instance()` is long we can name it as `MenuLink::to()`,
same method name we have to the class `humhub\widgets\bootstrap\Link`.

luke-:
@yurabakhtin Thanks, looks good! If the PR isn't urgent, I'd like to put it on hold. We also wanted to take a closer look at the link/button classes, and it really makes sense to standardize them here as well.

yurabakhtin:
@luke- Yes, it is not urgent, but it would be good to refactor the menu link adding in future.

PR #8036, opened 2026-02-25
UI picker refresh

gevorgmansuryan:
https://github.com/humhub/humhub-internal/issues/866
https://www.figma.com/design/tshvUVBzxy7uW8eZs6n6IG/HumHub---Design-System?node-id=9190-2038


gevorgmansuryan:
@mbumpalumpa can you please review?

mbumpalumpa:
@gevorgmansuryan there are some issues and one change.

So you can see in the screenshots that the alignment isn't the best. People/Space images are outside of the pill itself, and in the case of Post Type, you can see that the icon is sticking to the bottom rather than being centered. Also, you can see that the pill position in the picker itself looks like the bottom padding/margin is larger.

Additionally, the change would be the X within the pill itself. The original concept indeed used the "X" that you implemented; however, the Font Awesome X is way too thick and dominates the pill itself, so we should use `fa-times-circle` within the pills.

Additionally, I would expect pills and pickers to look the same regardless of where they are. You can see that the User Picker Pill in Stream Filter and the User Picker Pill in Group Members look different, aka the pill is larger, which is good. Basically, we currently have two different sizes/heights: 23px and 28px. I would much rather standardize this to 28px across the board.

Screen Shot 2026-04-20 at 11 49 22

Screen Shot 2026-04-26 at 20 20 12

----

Here is an example of pickers for [Design System v2](https://dsv2.xusko.dev/toasts); you can see that all spacing is unified (Ignore the design itself, which is a lot different than what we currently have).

Screen Shot 2026-04-26 at 20 50 47

gevorgmansuryan:
@mbumpalumpa thanks! Should be fixed now

mbumpalumpa:
@gevorgmansuryan looks much better, thanks. Tho I have 3 (4) requests.

1. Images of Spaces and Users need to have a 2px border radius. Currently, their border radius is larger than the pill's, which looks odd.
2. It seems that the gap between the elements within the topic (icon/img - name - remove) is too small, and it looks odd. We should increase this to min 4px.
3. When typing a topic that does not exist, instead of showing only a single entry, e.g., "+ MyCoolTopic", I get 2 entries: “MyCoolTopic” and “+ MyCoolTopic”. Idk, this might be on my end, but if it is not, we should only see the "+" version.
4. This one is not related to this redesign, but rather to the old behavior. Basically, if you add 2-3 pills and then press backspace to delete only one, it will clear all 3. That should be changed because we already have the X in the field itself that removes all topics. So to sum it up, one backspace should delete only a single entry in the picker.

PR #8056, opened 2026-03-06
Fix linked badge icon

yurabakhtin:

- [x] Bugfix

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
I removed the `->icon($this->icon)` because the module Question has the error on render the badge "BEST ANSWER" [here](https://github.com/humhub/questions/blob/master/widgets/BestAnswerButton.php#L44): `humhub\widgets\bootstrap\Button::icon(): Argument #1 ($icon) must be of type humhub\modules\ui\icon\widgets\Icon|string, null given, called in /Users/yb/Sites/hh/humhub/protected/humhub/widgets/bootstrap/Badge.php on line 127`

It is related to [this change](https://github.com/humhub/humhub/commit/bd06ab4c75f6c65295ee3e1ce3643437e8f9d10a#diff-d3881698349664a6332f7d1d9864fc0cfb6e76aa7017c4a3d7f455281a0c1949L125-L129). But I don't think we need to prepend icon to the link label because the icon is already rendered in the method `run()` [here](https://github.com/humhub/humhub/blob/master/protected/humhub/widgets/bootstrap/Badge.php#L57-L59).

Currently the "BEST ANSWER" badge doesn't have any icon, it is why we have the error there, but if I try to add some icon to the badge with this new PR change it looks correctly:
best-icon

but if we keep previous code(before the PR https://github.com/humhub/humhub/pull/8039) like:

```php
public function action($handler, $url = null, $target = null)
{
    $this->link = Link::withAction($this->label, $handler, $url, $target)
        ->encodeLabel($this->encodeLabel);
    if ($this->icon) {
        $this->link->icon($this->icon);
    }
    return $this;
}
```

then the icon is rendered twice:
best-icon-twice
it is why I decided to remove the line completely.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ HumHub Changelog
 - Enh #8051: Add `AltchaCaptchaInput::$showOnFocusElement` and `YiiCaptchaInput::$showOnFocusElement` (see [migration guide](https://github.com/humhub/humhub/blob/master/MIGRATE-DEV.md#version-1181) for details)
 - Fix #8054: Login layout widths (Default: Bootstrap width, Registration: 500px, Login & Password: 300px, Login with multiple SSO buttons: 500px)
 - Enh #8044: Update package `firebase/php-jwt` to v7
+- Fix #8056: Fix linked badge icon
 
 1.18.1 (March 2, 2026)
 ----------------------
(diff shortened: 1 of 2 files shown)


marc-farre:
@yurabakhtin yes, I agree. Tested, it works well.

PR #8071, opened 2026-03-17
Fix button link with icon only

yurabakhtin:

- [x] Bugfix

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
Before:
link-icon-right-space
After:
after-fix-link-icon

@marc-farre I find the style class `link-icon-only` was added [here](https://github.com/humhub/humhub/commit/977ee05e92c3e07be6462e5025be73426995046b#diff-2928d8fb18b3e93a39bd9e4830f58d3b5473fa7f85808754ce92682157ea8c23R5) only in the scss file, but it is not applied.
Do you agree this change so now button link with only icon has styles as `btn link link-icon-only btn-icon-only`, or maybe we should remove the link styles from such button link and keep only `btn btn-icon-only`?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ HumHub Changelog
 - Fix #8031: Border Radius for Color Picker under Theme Customisation
 - Fix #8006: Web Installer Nginx Example
 - Fix #8041: Update `Button::asLink` to `Link::to`
+- Fix #8071: Fix button link with icon only
 
 1.18.1 (Unreleased)
 -------------------------
--- a/protected/humhub/widgets/bootstrap/Link.php
+++ b/protected/humhub/widgets/bootstrap/Link.php
@@ -94,4 +94,16 @@ class Link extends Button
     {
         return $this->target('_blank');
     }
+
+    /**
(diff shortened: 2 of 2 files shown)


luke-:
@yurabakhtin Shouldn't that fixed in master?

yurabakhtin:
> Shouldn't that fixed in master?

@luke- `master` branch doesn't have the issue because when a link is built as `Link::primary()` it has only `btn btn-primary` class:

master-btn-icon

but in `develop` when we call `Link::primary()` we have the additional class `link`, it is started after my change [here](https://github.com/humhub/humhub/blob/e73419390aac63ed280d3223fecd9418cd8f3d4e/protected/humhub/widgets/bootstrap/Link.php#L37), so `Link::primary()` is rendered as `btn btn-primary link`.
Maybe we should add the class `link` only when we call `Link::to()`.
Or maybe we should deny to call `Link::primary()` because correctly is `Button::primary()`.

luke-:
@yurabakhtin Thanks for the explaination. I'll merge it for now to `develop`. We may need to revise the Link/Button stuff later.

Issue #8066, opened 2026-03-11
topics in new post won't save


schneiderws:
### What steps will reproduce the problem?

when editing a new post in a space or user profile I can add topics.
These topics aren't saved

### What is the expected result?

that the topics of the post will be saved

### What do you get instead?

just the post without the topics

### Additional info

when I edit an existing post I can add topics and these will be saved

| Q                | A
| ---------------- | ---
| HumHub version   | 1.17.3
| PHP version      |  8.4.18
| Operating system | linux

<img width="1198" height="447" alt="Image" src="https://github.com/user-attachments/assets/5578fdae-44a5-4965-918c-a1d86b8ef09f" />
<img width="2427" height="650" alt="Image" src="https://github.com/user-attachments/assets/31e004b8-38a9-4517-88c7-8ca4e2918025" />

> Note: Also provide the full error stacks from `Administration --> Information --> Logging` and errors from your browser console if possible.
> Note: Please use the [community](https://community.humhub.com) site for any questions on customizing HumHub.

marc-farre:
@luke- I confirm the issue.

gevorgmansuryan:
@luke- should be ok now

PR #8077, opened 2026-03-20
topics in new post won't save

gevorgmansuryan:
https://github.com/humhub/humhub/issues/8066

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ HumHub Changelog
 - Enh #8061: Display warning when URL cannot be embedded
 - Fix #8062: Fix global topics
 - Fix #8072: Fix missing translations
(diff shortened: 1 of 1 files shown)


protected/humhub/modules/content/widgets/WallCreateContentForm.php
```diff
@@ -141,7 +141,7 @@ public static function create(ContentActiveRecord $record, ?ContentContainerActi
 
         if ($record->save()) {
             $topics = Yii::$app->request->post('postTopicInput');
-            if (!empty($topics) && Topic::isAllowedToCreate($contentContainer)) {
+            if (!empty($topics)) {
```
gevorgmansuryan, line 144: @luke- there is correct `Topic::isAllowedToCreate` inside `Topic::attach`

PR #8098, opened 2026-04-09
Improve notification authorization checks

gevorgmansuryan:
https://github.com/humhub/humhub-internal/issues/1145

### DIFF
--- a/protected/humhub/modules/notification/components/BaseNotification.php
+++ b/protected/humhub/modules/notification/components/BaseNotification.php
@@ -296,19 +296,6 @@ abstract class BaseNotification extends SocialActivity
     }
 
     /**
-     * @inheritdoc
-     */
-    public function getContentInfo(?ContentOwner $content = null, $withContentName = true)
-    {
-        $user = $this->record->user ?? null;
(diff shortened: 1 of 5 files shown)


protected/humhub/modules/notification/components/BaseNotification.php
```diff
@@ -290,6 +295,19 @@ public function isBlockedForUser(User $user): bool
         return false;
     }
 
+    /**
+     * @inheritdoc
+     */
+    public function getContentInfo(?ContentOwner $content = null, $withContentName = true)
```
luke-: @gevorgmansuryan Why do we need the ContentInfo here? 

Is this for existing notifications?

gevorgmansuryan: @luke- Yes, for existing notifications. It is extra protection, so if a notification record already exists and gets rendered later, it still will not show the preview text.

luke-: @gevorgmansuryan I’d rather remove that; it could cause empty notifications, for example. I’d say that once the user has received the notification, they should be able to see it. Maybe we could optimize that again

luke- (approved):

PR #8105, opened 2026-04-22
Fix adding of new individual user and space topics

yurabakhtin:

- [x] Bugfix

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
https://github.com/humhub/humhub-internal/issues/1153

### DIFF
--- a/protected/humhub/modules/user/components/BaseAccountController.php
+++ b/protected/humhub/modules/user/components/BaseAccountController.php
@@ -8,9 +8,9 @@
 
 namespace humhub\modules\user\components;
 
-use humhub\components\Controller;
-use Yii;
 use humhub\components\access\ControllerAccess;
+use humhub\modules\content\components\ContentContainerController;
+use Yii;
 
 /**
  * BaseAccountController is the base controller for user account (settings) pages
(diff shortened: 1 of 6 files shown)


protected/humhub/modules/user/components/BaseAccountController.php
```diff
@@ -8,18 +8,23 @@
...
+use Yii;
 
 /**
  * BaseAccountController is the base controller for user account (settings) pages
  *
  * @since 1.1
  * @author luke
  */
-class BaseAccountController extends Controller
+class BaseAccountController extends ContentContainerController
```
luke-: @yurabakhtin I'm a little uneasy about the `ContentContainerController` here. The controller is very powerful and might load the wrong controllers based on the `cguid` parameter.

Can you see if you can find a solution without making this change, or does it make the code much cleaner?

luke- (changes requested):

protected/humhub/modules/user/components/BaseAccountController.php
yurabakhtin: @luke- Ok, please review https://github.com/humhub/humhub/pull/8105/changes/e367af65c463615da566e2ef94fb4c823e1fcec0

luke- (approved):

PR #8167, opened 2026-05-19
Improve safe migration methods

yurabakhtin:

- [x] Bugfix

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**
Some MySQL/MariaDB versions cannot detect foreign key by `information_schema.key_column_usage`, so `safeDropForeignKey()` skips the dropping of a foreign key when it really exists, so then error occurs on next deleting of a related column.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 HumHub Changelog
 ================
 
+1.18.4 (Unreleased)
+---------------------
+- Enh #8167: Improve safe migration methods
+
 1.18.3 (May 18, 2026)
 ---------------------
 - Fix #8079: Hide wall entry topics container when no topics
--- a/protected/humhub/components/Migration.php
+++ b/protected/humhub/components/Migration.php
@@ -139,16 +139,13 @@ class Migration extends \yii\db\Migration
      */
     protected function safeCreateTable(string $table, $columns, ?string $options = null)
     {
-        if (!$this->db->getTableSchema($table, true)) {
+        try {
             $this->createTable($table, $columns, $options);
             return true;
+        } catch (\Exception $e) {
+            $this->safeSkipCommand("Tried to create a table '$table'", $e);
+            return false;
         }
-
-        if (!$this->compact) {
-            echo "    > skipped create table $table, table does already exist ...\n";
-        }
-        $this->logWarning("Tried to create an already existing existing table '$table'");
-        return false;
     }
 
     /**
@@ -161,16 +158,13 @@ class Migration extends \yii\db\Migration
      */
     protected function safeDropTable(string $table)
     {
-        if ($this->db->getTableSchema($table, true)) {
+        try {
             $this->dropTable($table);
             return true;
+        } catch (\Exception $e) {
+            $this->safeSkipCommand("Tried to drop a table '$table'", $e);
+            return false;
         }
-
-        if (!$this->compact) {
-            echo "    > skipped drop table $table, table does not exist ...\n";
-        }
-        $this->logWarning("Tried to drop a non existing table '$table'");
-        return false;
     }
 
     /**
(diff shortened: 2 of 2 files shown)


luke-:
@yurabakhtin Should we stay with the current version or switch to the new approach?

yurabakhtin:
@luke- Current version checks only for FK existence, but new approach can catch all other errors, for example, if a table doesn't exist.

How new approach works:
- Test code:
```php
$this->safeAddForeignKey('fk-test', 'custom_pages_page_test', 'test', 'test', 'id', 'CASCADE');
$this->safeDropForeignKey('fk-test', 'custom_pages_page_test');
```

- Console:
```
*** applying m260520_053118_test
    > add foreign key fk-test: custom_pages_page_test (test) references test (id) ...    > skipped create foreign index fk-test in the table custom_pages_page_test, probably foreign index already exists ...
    > drop foreign key fk-test from table custom_pages_page_test ...    > skipped drop foreign index fk-test from the table custom_pages_page_test, probably foreign index does not exist ...
*** applied m260520_053118_test (time: 0.015s)
```

- Log:
log-migration

---
You can see here a small issue, the `{class}` is not replaced with expected migration name, it can be fixed [here](https://github.com/humhub/humhub/blob/master/protected/humhub/components/Migration.php#L764) with new code:
- `$params = array_combine(array_map(static fn($key) => "{{$key}}", array_keys($params)), $params);` (`array_keys` must be added):
log-migration-fixed

---

We can stay on the current version because as we detected the problem was because of wrong FK names on the server, it is a rare case and can be fixed only manually, but I think new approach is better because:
- no need to call the additional SQL query(method `foreignIndexExists()`), it may works a bit faster,
- it can catch all other errors and it writes into logs a real error.

We could also rewrite all other `safe*` methods of the class `humhub\components\Migration` to new approach with `try/catch` instead of the current checks by `columnExists`, `indexExists`, `db->getTableSchema`.

yurabakhtin:
@luke- Commit https://github.com/humhub/humhub/pull/8167/commits/866d5af72f8c1b9c8ea1ce53c4fdaf7bf339d088 for all safe methods:

my test code:
```
$this->safeAddForeignKey('fk-test-1', 'custom_pages_page_setting', 'page_id', 'custom_pages_page', 'id', 'CASCADE');
$this->safeDropForeignKey('fk-test-1', 'custom_pages_page_setting');
$this->safeAddForeignKey('fk-test', 'custom_pages_page_test', 'test', 'test', 'id', 'CASCADE');
$this->safeDropForeignKey('fk-test', 'custom_pages_page');
$this->safeAddPrimaryKey('pk-test', 'custom_pages_page_test', ['id']);
$this->safeDropPrimaryKey('pk-test', 'custom_pages_page_test');
$this->safeCreateIndex('idx-test', 'custom_pages_page_test', ['id']);
$this->safeCreateTable('custom_pages_page', ['id' => $this->primaryKey()]);
$this->safeDropTable('custom_pages_page_test');
$this->safeAddColumn('test', 'custom_pages_page_test', $this->integer()->notNull());
$this->safeDropColumn('test', 'custom_pages_page_test');
$this->safeRenameColumn('custom_pages_page_test', 'test', 'test2');
$this->safeAlterColumn('custom_pages_page', 'test', $this->string(40)->null());
```

Console:
```
*** applying m260520_053118_test
    > add foreign key fk-test-1: custom_pages_page_setting (page_id) references custom_pages_page (id) ... done (time: 0.062s)
    > drop foreign key fk-test-1 from table custom_pages_page_setting ... done (time: 0.049s)
    > add foreign key fk-test: custom_pages_page_test (test) references test (id) ...
      - Skipped: Table 'humhub_master.custom_pages_page_test' doesn't exist
    > drop foreign key fk-test from table custom_pages_page ...
      - Skipped: Can't DROP FOREIGN KEY `fk-test`; check that it exists
    > add primary key pk-test on custom_pages_page_test (id) ...
      - Skipped: Table 'humhub_master.custom_pages_page_test' doesn't exist
    > drop primary key pk-test ...
      - Skipped: Table 'humhub_master.custom_pages_page_test' doesn't exist
    > create index idx-test on custom_pages_page_test (id) ...
      - Skipped: Table 'humhub_master.custom_pages_page_test' doesn't exist
    > create table custom_pages_page ...
      - Skipped: Table 'custom_pages_page' already exists
    > drop table custom_pages_page_test ...
      - Skipped: Unknown table 'humhub_master.custom_pages_page_test'
    > add column custom_pages_page_test integer NOT NULL to table test ...
      - Skipped: Table 'humhub_master.test' doesn't exist
    > drop column custom_pages_page_test from table test ...
      - Skipped: Table 'humhub_master.test' doesn't exist
    > rename column test in table custom_pages_page_test to test2 ...
      - Skipped: Table 'humhub_master.custom_pages_page_test' doesn't exist
    > alter column test in table custom_pages_page to string(40) NULL DEFAULT NULL ...
      - Skipped: Unknown column 'test' in 'custom_pages_page'
*** applied m260520_053118_test (time: 0.149s)
```

Log:
log

yurabakhtin:
closed because `develop` PR https://github.com/humhub/humhub/pull/8168

PR #8189, opened 2026-05-29
fix: scroll to highlighted comment after stream animation completes

Semir1212:
## Summary

- Removes premature `scrollIntoView()` call from `comments.php` that fired during DOM insertion inside `_streamEntryAnimation()`, before the `hide().fadeIn()` animation caused a layout reflow discarding the scroll position
- Moves scroll logic to `Stream.prototype.triggerInitEvent()`, which is called after the `fadeIn()` Promise resolves and the layout is stable

## Root cause

`_streamEntryAnimation()` inserts HTML into the DOM (triggering inline scripts including `scrollIntoView()`), then immediately calls `hide().fadeIn()` in a `setTimeout`. The `hide()` causes a layout reflow that resets the scroll position set moments before.

## Test plan

- [ ] Navigate to Last Activities widget and click on a comment link
- [ ] Verify the stream scrolls to and highlights the correct comment
- [ ] Verify the scroll works when the stream entry requires animation (first load)
- [ ] Verify no regression when opening comments without a specific comment target

Fixes humhub/humhub-internal#1198

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ HumHub Changelog
 ---------------------
 - Enh #8170: Handle controllers with using external modules
 - Enh #8176: Upgrade Twig package to v3.26.0
+- Fix #8189: Scroll to highlighted comment when navigating from Last Activities
 
 1.18.3 (May 18, 2026)
 ---------------------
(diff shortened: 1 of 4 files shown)


luke- (changes requested):
I don't think that will work for comment links that don't go to the stream, e.g., wiki pages, Q&A...
Could you please check that?

luke- (approved):

PR #8191, opened 2026-05-31
[Security] Restrict allowed_classes in LiveEvent unserialize() to prevent PHP Object Injection

XananasX7:
## Summary

`PollController::unserializeEvent()` deserializes `LiveEvent` objects from the `live` database table without an `allowed_classes` restriction. Without this restriction, PHP will instantiate any class found in the serialized data, which can be exploited for **PHP Object Injection** via gadget chains.

## Attack Surface

The `live` table is written by the application when users perform actions (new content, comments, notifications). An attacker with a valid user account who can trigger live event storage could potentially inject a crafted serialized payload into the database, which would then be deserialized by other users' polling requests.

## Fix

Restrict `allowed_classes` to the 5 known `LiveEvent` subclasses used by HumHub core:

- `humhub\modules\comment\live\NewComment`
- `humhub\modules\content\live\NewContent`
- `humhub\modules\notification\live\NewNotification`
- `humhub\modules\live\live\LegitimationChanged`
- `humhub\modules\live\live\PreventPjaxOnNextClick`

Any unexpected class in serialized live event data will now be returned as `__PHP_Incomplete_Class` instead of a live instantiated object, neutralizing gadget chain exploitation.

> **Note for maintainers:** Third-party modules that define custom `LiveEvent` subclasses may need to extend this list or register their classes via a hook. Consider adding a `registerLiveEventClasses()` mechanism in `LiveEvent` to allow extensible allowed-classes registration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ HumHub Changelog
 ---------------------
 - Enh #8170: Handle controllers with using external modules
 - Enh #8176: Upgrade Twig package to v3.26.0
+- Fix #8189: Scroll to highlighted comment when navigating from Last Activities
 - Enh #8187: Update composer package symfony/mime
 - Fix #8188: AutoContrast and Sign In Footer Links
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ HumHub Changelog
 - Enh #8170: Handle controllers with using external modules
 - Enh #8176: Upgrade Twig package to v3.26.0
 - Enh #8187: Update composer package symfony/mime
+- Fix #8188: AutoContrast and Sign In Footer Links
 
 1.18.3 (May 18, 2026)
 ---------------------
@@ -67,7 +68,7 @@ HumHub Changelog
 ----------------------
 
 > This release also fixes a [security](https://github.com/humhub/humhub/security/advisories/GHSA-qxjh-478x-23gm) issue.
- 
+
 - Fix #8003: `Migration::foreignIndexExists()` doesn't find tables in braces
 - Fix #8002: Comment dropdowns truncated (e.g. to select a title)
 - Fix #8007: Fix unsaved changes warning on Profile Edit page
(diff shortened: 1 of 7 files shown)


XananasX7:
I have read the CLA Document and I hereby sign the CLA

luke-:
@XananasX7 Thanks for the contribution and pointing out this. 

Unfortunately, especially when it comes to custom modules, hard-coding this list is currently not an option. Some extra Event handling here, is probably too much overhead.

Am I correct in thinking that the main risk is that an attacker could gain control of the database?

Quick ideas to get allowed class names:

1.  Search the deserialized data for the class, check if it is of type “LiveEvent,” and if so, auto whitelist it
2. Also store the class in the table and whitelist it.

XananasX7:
Yes, that's the right mental model. The main risk is that if an attacker can write arbitrary data to the database row that stores the serialized `LiveEvent` (e.g. through a separate SQL injection or a compromised admin account), they could plant a crafted PHP object. When `unserialize()` processes it without `allowed_classes`, PHP will instantiate whatever class is named in the payload — and if a class with a useful `__wakeup()` or `__destruct()` method exists anywhere in the autoloader, it becomes a gadget chain. So yes, database write access would be the prerequisite.

Your two suggested approaches both make sense. Option 1 — inspecting the serialized data for the class name first — is essentially what `allowed_classes` does at the C level, just done in PHP userland. Option 2 (storing the class name in the table) is more robust and future-proof for custom modules.

A simpler middle ground that might work for your architecture: emit a filter/hook before the `unserialize()` call so custom modules can register their expected classes, then pass the merged list as `allowed_classes`. That keeps the hardening in core without needing a DB schema change and gives module authors a clear extension point. Happy to sketch out a patch along those lines if it's useful.

luke-:
@XananasX7 I think

luke-:
@XananasX7 I would probably prefer this approach:
https://github.com/humhub/humhub/pull/8201

Requiring modules to whitelist their LiveEvent class would be too much overhead.

Do you see any problems here?

XananasX7:
Thanks @luke-! Had a look at #8201 — the `EVENT_BEFORE_UNSERIALIZE` hook approach is much better than a hard-coded allowlist. Modules can register their own classes, core stays clean, no DB schema changes needed. That's the right call. Closing this PR in favour of #8201.

PR #8193, opened 2026-06-01
Fix rendering of the widget `AuthChoice`

yurabakhtin:

- [x] Bugfix

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
https://github.com/humhub/humhub-internal/issues/1197

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ HumHub Changelog
 - Fix #8189: Scroll to highlighted comment when navigating from Last Activities
 - Enh #8187: Update composer package symfony/mime
 - Fix #8188: AutoContrast and Sign In Footer Links
+- Fix #8193: Fix rendering of the widget `AuthChoice`
 
 1.18.3 (May 18, 2026)
 ---------------------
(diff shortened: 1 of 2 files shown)


luke- (changes requested):
@yurabakhtin Can you please add comments to explain we're doing this?

yurabakhtin:
@luke- Maybe my fix looks like a hack but really it is very strange why `yii\authclient\widgets\AuthChoice` was developed with `echo` inside the method `init()` [here](https://github.com/yiisoft/yii2-authclient/blob/master/src/widgets/AuthChoice.php#L271) as @marc-farre detected, so I find only one approach is to store all output content from the `Yii\AuthChoice\init()` to the variable `_initOutput` and then print it on the method `run()`, so it works now this way:

- `AuthChoice::widget()`:
  - `init()` - store the output content into the var `$this->_initOutput`(`<div id="...">`), don't render here
  - `beforeRun()` - if `false` then `run` is not called, else:
    - `run()` - print here `$this->_initOutput` + all other expected content like the end tag `</div>`.

luke-:
@yurabakhtin Could you please add comments in the code to explain why it's solved this way?

yurabakhtin:
> Could you please add comments in the code to explain why it's solved this way?

@luke- Commit https://github.com/humhub/humhub/pull/8193/changes/32beae34b05445ee6adfe2cb985d042ea330da26

luke- (approved):

PR #8200, opened 2026-06-02
Restrict content container URL fallback to container controllers

gevorgmansuryan:
https://github.com/humhub/humhub-internal/issues/1179


luke-:
@gevorgmansuryan Can you please base this PR against `develop`?

PR #8241, opened 2026-06-23
Fix AssetImage stale-cache crash and mail header path/render bugs

luke-:

- [x] Bugfix

- [x] No

## Summary

Three regressions introduced by the asset-system refactor (#8011):

### 1. `AssetImageRegistry` never invalidated the cached `fileExists` flag

The `EVENT_AFTER_REQUEST` cache-update used `??` where `||` was intended:

```php
if (!isset($this->_cache[$name]['fileExists']) ?? $this->_cache[$name]['fileExists'] !== $assetImage->fileExists) {
```

`!isset(...)` is always a bool (never `null`), so the null-coalescing right-hand side was never evaluated — the condition collapsed to "only when not yet cached". As a result `fileExists` was cached once and **never updated again**.

After **deleting or replacing** an asset image, the registry kept reporting it as existing, so `AssetImage::getUrl()` ran `convert()` → `Filesystem::read()` on the now-missing file and threw `League\Flysystem\UnableToReadFile`. This crashed the **admin Design settings page** and the **login page** (both call `loginBackground->getUrl()`), which surfaced as "the form is broken / cannot upload an image".

### 2. `mailHeader` shared the icon's storage path

```php
'mailHeader' => ['file' => '/icon/icon.png'],
```

Uploading a mail header overwrote the site icon (and vice versa). It now has its own path `/mail-header/header.png` (matching the pre-refactor `mail-header/header.png` location).

### 3. `MailHeaderImage` rendered the logo instead of the mail header

The widget called `Yii::$app->img->logo->getUrl(...)` in the mail-header branch, so an uploaded mail header was never actually shown.

## Testing

Verified on a local `develop` instance:
- Upload → delete → reload of the design page no longer throws `UnableToReadFile` (crashed before the fix).
- Valid login background (PNG and JPEG, 1920×1080) saves and renders.
- Mail header now stores under `uploads/mail-header/` without touching the site icon, and its preview/mail render uses the correct image.


protected/humhub/components/assets/AssetImage.php
```diff
@@ -95,7 +95,7 @@ public function getUrl(?array $options = null, $scheme = false): string
             //Yii::debug("AssetImage: Check file exists: " . $scaledFileName);
 
             if (!$this->exists() && empty($this->defaultFile)) {
-                throw new InvalidValueException('File and DefaultFile cannot be empty.');
+                return '';
```
luke-, line 98: @yurabakhtin I'd prefer to use an exception so it's clear when a file doesn't exist. 

Just an empty string would look a bit wrong. Do you see a way to keep the design using an exception?

luke- (commented):
@yurabakhtin Can you please take a look?

protected/humhub/components/assets/AssetImage.php
yurabakhtin: @luke- [Here](https://github.com/humhub/humhub-internal/issues/1238#issuecomment-4786778407) I wrote why I decided to modify the method `getUrl()` to return empty string instead of the exception.

I prefer to return empty string or we can return `null` to know the Asset Image has no custom uploaded image and has no Default File.
As I wrote there several cases exist, for example [here](https://github.com/humhub/humhub/blob/develop/protected/humhub/widgets/views/logo.php#L14-L20) that may work wrong when `logo` image will have a defined `defaultFile` in some custom `common.php`:

```php
if (Yii::$app->img->logo->exists()) {
   print Yii::$app->img->logo->getUrl(...)
}
```

so for correct work we should update the `Yii::$app->img->logo->exists()` to `Yii::$app->img->logo->exists() || Yii::$app->img->logo->defaultFile` like this:

```php
if (Yii::$app->img->logo->exists() || Yii::$app->img->logo->defaultFile) {
   print Yii::$app->img->logo->getUrl(...)
}
```

instead of such long code in "template" files I prefer:

```php
$logoUrl = Yii::$app->img->logo->getUrl(...);
if ($logoUrl) {
   print $logoUrl;
}
```

---

or implement new method `has()` => `return $this->exists() || $this->defaultFile` and use it as:
```php
if (Yii::$app->img->logo->has()) {
   print Yii::$app->img->logo->getUrl(...)
}
```

PR #8254, opened 2026-06-28
Enh: Allow reading content in State mode for owner

marc-farre:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**

https://github.com/humhub/humhub-internal/issues/1232

### DIFF
--- a/protected/humhub/modules/content/components/ActiveQueryContent.php
+++ b/protected/humhub/modules/content/components/ActiveQueryContent.php
@@ -80,7 +80,9 @@ class ActiveQueryContent extends ActiveQuery
      */
     public function readable()
     {
-        $this->andWhere($this->stateFilterCondition);
+        if ($this->stateFilterCondition !== null) {
+            $this->andWhere($this->stateFilterCondition);
+        }
 
         $this->joinWith(['content', 'content.contentContainer', 'content.createdBy']);
         $this->leftJoin('space', 'contentcontainer.pk=space.id AND contentcontainer.class=:spaceClass', [':spaceClass' => Space::class]);
--- a/protected/humhub/modules/content/components/ActiveQueryContent.php
+++ b/protected/humhub/modules/content/components/ActiveQueryContent.php
@@ -56,29 +56,30 @@ class ActiveQueryContent extends ActiveQuery
      */
     public ?array $stateFilterCondition = null;
 
-    /**
-     * Only returns user readable records
-     *
-     * @param User $user
-     * @return ActiveQueryContent
-     * @throws Throwable
-     */
-    public function readable($user = null)
+    public function init()
     {
-        if ($user === null && !Yii::$app->user->isGuest) {
-            $user = Yii::$app->user->getIdentity();
-        }
-
         if ($this->stateFilterCondition === null) {
             $this->stateFilterCondition = ['OR', ['content.state' => Content::STATE_PUBLISHED]];
-            if ($user !== null) {
+            if (!Yii::$app->user->isGuest) {
                 $this->stateFilterCondition[] = [
                     'AND',
                     ['content.state' => Content::STATE_DRAFT],
-                    ['content.created_by' => $user->id],
+                    ['content.created_by' => Yii::$app->user->id],
                 ];
             }
         }
+
+        parent::init();
+    }
+
+    /**
+     * Only returns user readable records
+     *
+     * @return ActiveQueryContent
+     * @throws Throwable
+     */
+    public function readable()
+    {
         $this->andWhere($this->stateFilterCondition);
 
         $this->joinWith(['content', 'content.contentContainer', 'content.createdBy']);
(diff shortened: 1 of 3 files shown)


protected/humhub/modules/content/components/ActiveQueryContent.php
```diff
@@ -42,35 +42,45 @@ class ActiveQueryContent extends ActiveQuery
...
-     * @param User $user
      * @return ActiveQueryContent
      * @throws Throwable
      */
-    public function readable($user = null)
+    public function readable()
     {
-        if ($user === null && !Yii::$app->user->isGuest) {
-            $user = Yii::$app->user->getIdentity();
+        if ($this->stateFilterCondition) {
```
yurabakhtin: @marc-farre I think here we can remove the condition, because the property `$stateFilterCondition` can be only array, and even if we set it as empty array like:

```php
 $query = Post::find();
 $query->stateFilterCondition = [];
 $posts = $query->readable()->all();
```

then it selects all states.

yurabakhtin (changes requested):

protected/humhub/modules/content/components/ActiveQueryContent.php
marc-farre: Done in commit https://github.com/humhub/humhub/pull/8254/commits/2a9485b1125b64a212dd71b3615d557757c11a1f

yurabakhtin (approved):

protected/humhub/modules/content/components/ActiveQueryContent.php
```diff
@@ -42,35 +42,42 @@ class ActiveQueryContent extends ActiveQuery
...
 
     /**
      * Only returns user readable records
      *
-     * @param User $user
      * @return ActiveQueryContent
      * @throws Throwable
      */
-    public function readable($user = null)
+    public function readable()
```
luke-, line 82: @yurabakhtin I would prefer if we can keep the param. Do you see a chance to keep it? Even it's mostly omitted.

luke- (changes requested):

protected/humhub/modules/content/components/ActiveQueryContent.php
yurabakhtin: @luke- yes, I don't see a problem to keep it, it was @marc-farre 's idea to remove it and I confirmed because as we discussed it is not used at all.

marc-farre: @luke- the issue is that we also add user related filters in `__construct()`. So if we leave the `$user` param, we could have a query with a mix of 2 different users.

luke-: @yurabakhtin Ok, can you please re-add it. Would be good, if such low level methods, can be used without Session context as well

luke- (changes requested):

protected/humhub/modules/content/components/ActiveQueryContent.php
marc-farre: @luke- see my comment (sorry, I missed the "PENDING" tag on it because I needed to submit my review)

yurabakhtin: > the issue is that we also add user related filters in `__construct()`. So if we leave the `$user` param, we could have a query with a mix of 2 different users.

@marc-farre Yes, you are right it can be mixed, but I also find the method `userRelated(..., $user = null)`, so we should either remove the param from all methods or init `user` param on call `__construct()` from `$config`.

luke-: Yes, we definitly should have one "user" source.

What you think about my change to move it to contructor?

yurabakhtin: > What you think about my change to move it to contructor?

@luke- I agree, but the method `userRelated()` should be updated as well.

luke-: Good catch, done — `userRelated()` now also drops the `$user` parameter and uses the constructor-resolved user, same as `readable()`.

luke-: @yurabakhtin Anything else?

protected/humhub/modules/content/components/ActiveQueryContent.php
```diff
@@ -56,13 +56,17 @@ class ActiveQueryContent extends ActiveQuery
      */
     public array $stateFilterCondition = ['OR', ['content.state' => Content::STATE_PUBLISHED]];
 
-    public function __construct($modelClass, $config = [])
+    private ?User $user = null;
+
+    public function __construct($modelClass, ?User $user = null, $config = [])
     {
-        if (!Yii::$app->user->isGuest) {
+        $this->user = $user ?? (!Yii::$app->user->isGuest ? Yii::$app->user->getIdentity() : null);
```
marc-farre: @luke- I think we don't need to test isGuest() because:

Image

```php
$this->user = $user ?? Yii::$app->user->getIdentity();
```

should be enough

luke-: Good catch, you're right — `getIsGuest()` is literally just `getIdentity() === null`, so the extra check was redundant. Simplified.

luke- (approved):

PR #8261, opened 2026-07-01
Enh: In the mobile app settings, change Whitelist domains to URLs, an…

marc-farre:
…d add SSO URLs to the list for the mobile app to open all of them in the in-app browser instead of the external one

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**

https://github.com/humhub/app-internal/issues/146


protected/humhub/modules/admin/models/forms/MobileSettingsForm.php
```diff
@@ -8,9 +8,13 @@
 
 namespace humhub\modules\admin\models\forms;
 
+use humhub\modules\user\widgets\AuthChoice;
 use humhub\services\WellKnownService;
 use humhub\widgets\bootstrap\Link;
 use Yii;
+use yii\authclient\OAuth1;
+use yii\authclient\OAuth2;
```
luke-: Do we really need this imports?

protected/humhub/modules/admin/models/forms/MobileSettingsForm.php
```diff
@@ -115,7 +105,7 @@ public function attributeHints(): array
                     WellKnownService::getFileRoute('fileAppleAssociation'),
                 )->target('_blank'),
             ]),
-            'whiteListedDomains' => Yii::t('AdminModule.settings', 'E.g. URLs of the SSO service providers'),
+            'whiteListedUrls' => Yii::t('AdminModule.settings', 'List of URLs that should be opened in the mobile in-app browser instead of the external one, and to be recognized as part of the HumHub ecosystem by the mobile app. Add * to the end of the URL to match all sub-paths. Example: https://example.com/*'),
```
luke-, line 104: I would actually always automatically assume a wildcard here. Or do you see any reasons why it might make sense to whitelist only a single URL and not any sub-URLs?

luke- (changes requested):

protected/humhub/modules/admin/models/forms/MobileSettingsForm.php
marc-farre: Thanks, I forgot to remove them...
Done in commit https://github.com/humhub/humhub/pull/8261/changes/1c3b04a71e2e77cb721541294f2bd6ff7ee08823

protected/humhub/modules/admin/models/forms/MobileSettingsForm.php
marc-farre: I asked me the same question, and I'm not sure here.
I agree that in 99% of the cases, auto-wildcard is fine, and maybe the 1% will never happen.
A very rare situation could be the need of whitlisting root pages of a website, but not some sub-URLs, e.g. a page for an LMS, webinar, video call, etc.

But we ask for an URL to whitelist, not a domain anymore, so to be sementically correct, it should apply to that URL, not all sub-URLs.
And here, the wildcard `*` makes it clear that it will apply to sub-URLs.

So if we decide to auto-wildcard, I think we should either:
- Explain it in the form field hint (e.g.: "All sub-URLs will also be whitelisted")
- Or auto-add the `*` at the end of the URL entered (I don't really like this solution)

Should I proceed to auto-wildcard and explain in the attribute hint?

luke-: @marc-farre Ok, thats right. Let's keep it with the attribute hint. Thanks, I'll merge the PR.

luke- (approved):

PR #8287, opened 2026-07-12
Allow changing container Profile and Banner image via Event

marc-farre:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] All tests are passing
- [ ] Changelog was modified

**Other information:**

- Enh: Add`ContentContainerActiveRecord::EVENT_CREATE_PROFILE_IMAGE` and `EVENT_CREATE_BANNER_IMAGE`
- Fix: `AssetImage::getUrl()` for a source file without an extension (e.g. a HumHub `File`, stored as `file`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,7 +64,7 @@ HumHub Changelog
 - Enh #8261: In the mobile app settings, change Whitelist domains to URLs, and also send auth client (SSO) URLs to the mobile app to open all of them in the in-app browser instead of the external one
 - Enh #8254: Allow reading content in State mode for owner
 - Fix #8267: Delete `\humhub\widgets\PageAddonStack` unused class
-- Enh #8287: Add `ContentContainerActiveRecord::EVENT_CREATE_PROFILE_IMAGE` and `EVENT_CREATE_BANNER_IMAGE`
+- Enh #8287: Add `ContentContainerActiveRecord::EVENT_INIT_PROFILE_IMAGE` and `EVENT_INIT_BANNER_IMAGE`
 - Fix #8287: `AssetImage::getUrl()` for a source file without an extension (e.g. a HumHub `File`, stored as `file`)
  
 1.18.4 (Unreleased)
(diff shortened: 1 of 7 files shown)


marc-farre:
@luke- Up to 1.18, I could replace the Space Profile image by overwritting [ContentContainerActiveRecord::$profileImageClass](https://github.com/humhub/humhub/blob/0f84ee6a9d42e8e79557c666db1117f1dd07d449/protected/humhub/modules/content/components/ContentContainerActiveRecord.php#L98-L98) via a `Space::EVENT_INIT`.

But in 1.19, `profileImageClass` is removed, so I have no ways anymore to replace the Space Profile image with another one via a module.
I need it for the Classified Spaces module, which allows to define a default image for all Spaces classified in a specific category.

This PR adds 2 events (for the Profile and Banner container image) and fixes a bug if the file extension should be retrieved from the file instead of using the 'extension' array key.

ArchBlood:
I agree as I see in v1.19 also deprecates ProfileImage lib class as well, so this would be interesting to look into using for the Gravatar module as well.

luke-:
Thanks @marc-farre — the use case makes sense, and events are the right mechanism for this since the `$profileImageClass` override no longer affects the new `AssetImage` path. I kept your approach and pushed two refinement commits directly to the branch:

- **Typed event class:** `ContentContainerImageEvent` replaces the generic `Event::$result`. It carries the created `AssetImage` (`$event->image`) plus the configuration it was built with (`$event->config`), so a handler can replace the image without duplicating the core defaults:
  ```php
  $event->image = new MyAssetImage($event->config);                  // swap in a subclass
  $event->image->defaultFile = '@mymodule/resources/img/space.jpg';  // or just change the default image
  ```
  Note that replacing is only required when the source `file` or the class should change — for a different default image, mutating `$event->image` is enough.
- **Renamed** the constants to `EVENT_INIT_PROFILE_IMAGE` / `EVENT_INIT_BANNER_IMAGE` (`initProfileImage`/`initBannerImage`) — closer to Yii's `EVENT_INIT` naming and to the old `Space::EVENT_INIT` mechanism, and "create" could be misread as "a new image was uploaded".
- The events fire **once per record instance** (the result is memoized); this and the mutate-vs-replace semantics are documented on the constants. The trigger helper is `private` — the constants are the public API.
- The migration guide now points former `$profileImageClass` overrides to these events.
- **Tests added:** event replacement/customization/once-per-record, plus a regression test for your `AssetImage` extension fix (it fails with `Undefined array key "extension"` without the fix).

luke-:
@marc-farre Can you please check my adjustments?

marc-farre:
@luke- Nice, thanks for these commits! I agree, it's better (I thought about creating a dedicated Event class, but I was unsure if it would be too much).
Tested successfuly.

PR #8328, opened 2026-07-20
Restore database-level like uniqueness for content likes

luke-:
## What

`idx_unique_user` (`content_id`, `content_addon_record_id`, `created_by`) contains the nullable `content_addon_record_id` — and NULL values are distinct in MySQL/MariaDB unique indexes. Likes on the content itself always store `NULL` there, so the database never enforced one-like-per-user for them: concurrent double-clicks / parallel requests could create duplicate likes (1.18 enforced this via the old `unique-object-user` index).

## How

New `like` migration that

- deduplicates existing content likes (keeps the oldest row),
- adds a generated column `content_addon_record_key` = `IFNULL(content_addon_record_id, 0)` — `PERSISTENT` on MariaDB (min version 10.1 does not know `STORED`), `STORED` on MySQL,
- creates `idx_unique_like` (`content_id`, `content_addon_record_key`, `created_by`) and drops `idx_unique_user`.

Because MariaDB/MySQL refuse a generated column based on a column carrying an `ON UPDATE CASCADE` foreign key (error 1901), `fk_like_content_addon` is re-created with `ON UPDATE RESTRICT` — no behavioural change, `record_map` ids are never updated. `ON DELETE CASCADE` stays as is (verified: deleting a `record_map` row still cascades).

An addon like and a content like of the same user on the same content still coexist (`0` vs. the record-map id in the key column).

## Test

- New `LikeUniquenessTest::testDuplicateContentLikeIsRejectedByDatabase` — inserted a duplicate content like without error before the migration, now fails with an `IntegrityException` on `idx_unique_like`.
- Migration sequence additionally replayed on a scratch database (dedupe, coexistence, cascade) on MariaDB.

Part of the 1.19 before-stable items from the release assessment (P1).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ HumHub Changelog
 
 1.19 (TBD)
 ----------
+- Fix #8328: Duplicate likes on the same content were possible via concurrent requests — the unique index contained the nullable `content_addon_record_id` and therefore never applied to likes on the content itself; uniqueness is now enforced via a generated key column
 - Enh #8321: Add the `UnreadCountChangedEvent`, triggered whenever a user's unseen notification count changes or by modules such as Messenger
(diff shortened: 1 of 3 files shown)


luke-:
Closing without merge: we decided the added complexity (generated key column, driver-specific PERSISTENT/STORED syntax, re-created foreign key) outweighs the benefit. The application-level guard in LikeService::like() (getCurrentLikeRecord() before insert) remains the uniqueness guarantee for content likes; the residual race window only allows a rare duplicate like row, which is cosmetic. If duplicates ever show up in practice, a dedupe can be added to the like module's integrity check instead.

PR #8339, opened 2026-07-22
Make tour navigation buttons translatable

luke-:
## What

The introduction tour was migrated to [driver.js](https://driverjs.com) in 1.18. While the step titles and descriptions are translatable, the navigation buttons (Next / Previous / Done) and the progress text were left at driver.js' hardcoded English defaults and never passed through `Yii::t()`.

This exposes those labels via `Yii::t('TourModule.base', ...)` so they can be translated like the rest of the tour.

## How

The labels are merged into `driverJsOptions` in `Module::init()` using `+=`, so any option configured by an admin (or the existing `showProgress` default) still wins. The default English strings are kept identical to driver.js' defaults, so nothing changes for English installations. The `{{current}}` / `{{total}}` driver.js placeholders are preserved in the `progressText` string.


luke-:
@marc-farre Can you please check?

marc-farre:
@luke- I added a commit to prevent from an infinit loop due to Yii:t():

> The infinite loop came from calling Yii::t() during         
>  Module::init(): HumHub's I18N::getMessageSource() iterates every   
>  module to auto-assign message sources, which re-instantiates the   
>  tour module → init() → Yii::t() → recursion. 

Now it works correctly.

luke-:
@marc-farre Great, thanks!

PR #8341, opened 2026-07-22
Re-initialize eager widgets after async module registration

luke-:
## Problem

When a module's asset bundle is served from a **different origin** than the HumHub site — an S3 bucket or CDN configured via the `assets` mount — module JavaScript loaded inside an ajax/pjax response (typically a modal) is executed **asynchronously**. As a result `[data-ui-init]` widgets in that response are initialized **before** their module registered, and silently fail on the first open (working only from the second open on):

```
Required a non initialized module: collabora.Editor
```

Affects any module that registers its asset bundle inside a modal view — e.g. **collabora**, **text-editor**, **drawio**. Not reproducible with same-origin (local) assets.

### Root cause

`Modal.setDialog()` does `this.$.empty().append(content); this.applyAdditions();`, assuming all `<script src>` of the response have executed by the time `.append()` returns. For **cross-origin** scripts jQuery's script transport appends a real `<script>` element and completes on its `load` event — `async: false` is silently ignored (documented jQuery behavior). So `applyAdditions()` runs before the module JS executed, `require(ns)` resolves an empty namespace stub, and the widget never initializes.

## Fix

In `humhub.core.js`, after a module is registered **asynchronously** (post `humhub:ready`), re-apply the `ui.widget` addition to the `[data-ui-init]` nodes that map to the just-registered module and whose namespace now resolves.

- **Path-agnostic** — triggers on module registration, covering modals, Pjax and stream inserts alike; no modal-specific timing hack.
- **Idempotent** — already-initialized widgets are skipped by the existing `Component._getInstance` guard (the same guard that keeps today's repeated `applyTo` calls safe).
- **No side effects / no noise** — lookups use `resolveNameSpace(ns, false)`, which never creates a stub or logs a warning; only runs after initial page load, so zero startup cost.
- **Handles multi-script responses** — each cross-origin script that lands re-scans for its own widgets, so ordering between several async bundles resolves naturally.

## Docs

- `docs/develop/ui-js-overview.md`: note on the async re-init behavior and the cross-origin assets caveat.

## Testing

- Reproduced with the collabora module and the S3 `assets` mount (cross-origin `baseUrl`): modal opened with title but empty body, iframe `src` never set, `Required a non initialized module: collabora.Editor` on first open. With this change the widget initializes on the first open.
- Same-origin assets: unchanged behavior; already-initialized widgets are not re-initialized.
- `node --check` passes.

Reported for the S3 assets mount in humhub/s3#2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ HumHub Changelog
 
 1.19.0-beta.2 (TBD)
 -------------------
+- Fix #8341: When the `assets` mount was served from a different origin than the HumHub site (an S3 bucket or CDN), module JavaScript loaded inside an ajax/pjax response (e.g. a modal) executed asynchronously — the browser cannot load cross-origin `<script src>` synchronously — so `[data-ui-init]` widgets in the response were processed before their module registered and silently failed on the first open (`Required a non initialized module: …`), only working from the second open on; eager widgets of an asynchronously registered module are now re-initialized right after registration (idempotent, covers modals, pjax and streams)
 - Enh #8337: Added a mail (SMTP) delivery configuration step to the web installer (after the basic step) reusing the admin mailing settings form, with an inline "Test" button that verifies the entered settings via AJAX; configuring mail never blocks completion, and the step is skipped on automated/managed-hosting installs (fixed mail config)
 - Enh #8337: The `php` mail transport (native mail()/sendmail) is no longer offered under Docker (`HUMHUB_DOCKER`) — hidden in the mailing settings form and rejected in validation, since containers ship no local MTA
 - Enh #8336: Unified the two near-identical `UserModule.auth` translation keys for the maintenance mode message that differed only by a trailing period — the maintenance page heading now reuses the punctuated key, so the sentence no longer has to be translated twice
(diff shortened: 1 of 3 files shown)


luke-:
**On hold — likely insufficient on its own.**

Reproduced against a develop instance with cross-origin S3 assets: this fix deploys and runs, but a Space Calendar still fails because the real problem is broader than "widget's own module not registered." Cross-origin `<script src>` tags execute in download-completion order, not document order, so a module can run before a sibling **library dependency** it needs (e.g. `FullCalendar`) has executed — `ReferenceError: FullCalendar is not defined`. Late widget re-init can't fix a missing sibling dependency.

The more complete fix is ordered + awaited script execution in the content-insertion layer (pjax `context.html`, modal `setContent`/`setDialog`), which would make this PR redundant. Full analysis: humhub/s3#2 (comment). Holding this until we decide scope, given the network-wide blast radius.

luke-:
Superseded by #8344. The late widget re-init only covered the "widget initialized before its module registered" failure mode — it cannot help when a module executes before a sibling library dependency (e.g. FullCalendar), which the synchronous script transport in #8344 fixes at the root by restoring document-order execution. Verified against the same reproduction (space calendar with a cross-origin S3 assets mount): #8344 alone makes the first open work. Branch kept for reference.

Issue #5334, opened 2021-10-08
How to Exclude Notifications only for Files and Images


glibertdelyon:
Thanks to `streamExcludes` in `common.php` new files and new images are excluded from the stream,
So far so good...
... but  notifications  are still there.

Recently one of our member added **40 new images to an existing gallery**.
The result was that each member of the space recieved **40 emails**. Not good for the planet and very annoying.

Please, **is there a way to avoid notification only for Files and Images and to keep them for posts ??**

luke-:
You can try to add `public $silentContentCreation = true;` here:
https://github.com/humhub-contrib/gallery/blob/master/models/Media.php#L41

marc-farre:
@luke- I've tested this in `protected/common.php`:

```php
    'container' => [
        'definitions' => [
            \humhub\modules\cfiles\models\File::class => [
                'silentContentCreation' => true,
            ],
            \humhub\modules\gallery\models\Media::class => [
                'silentContentCreation' => true,
            ],
        ]
    ],
```

But it doesn't work. Do you know why?

ArchBlood:
> @luke- I've tested this in `protected/config/common.php`:
> 
> ```
>     'container' => [
>         'definitions' => [
>             \humhub\modules\cfiles\models\File::class => [
>                 'silentContentCreation' => true,
>             ],
>             \humhub\modules\gallery\models\Media::class => [
>                 'silentContentCreation' => true,
>             ],
>         ]
>     ],
> ```
> 
> But it doesn't work. Do you know why?

Maybe it's like `class_exists()` where `::class` can't be used?
```php
<?php

return [
    // Other configurations...

    'container' => [
        'definitions' => [
            'humhub\modules\cfiles\models\File' => [
                'silentContentCreation' => true,
            ],
            'humhub\modules\gallery\models\Media' => [
                'silentContentCreation' => true,
            ],
        ],
    ],

    // More configurations...
];
```

ArchBlood:
> > Maybe it's like `class_exists()` where `::class` can't be used?
> 
> @ArchBlood Thanks. But it should work with `::class` too, see this example: https://community.humhub.com/s/installation-and-setup/wiki/80/Configuration+Examples#hide-activity-box-on-given-space

Did you make sure that there was a `silentContentCreation` property in those models?

ArchBlood:
> > > Maybe it's like `class_exists()` where `::class` can't be used?
> > 
> > 
[...]

PR #8369, opened 2026-08-01
Enh: Added the content module option `silentContentCreationTypes`, li…

marc-farre:
…sting content record classes whose creation is never announced — no content created notification, no activity and no new content marker in the space chooser.

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**

https://github.com/humhub/humhub/issues/5334


luke-:
Thanks @marc-farre — the PR itself is clean and minimal, and I appreciate that it stays consistent with `streamExcludes`. My hesitation is not about the implementation, it's about the approach itself.

**Why I'd prefer not to merge this:**

1. **It suppresses information instead of condensing it.** The option mutes *every* creation of the configured type — including the single, genuinely relevant file upload — not just the mass-upload case that actually causes the pain. Once admins rely on this option, it becomes permanent API surface that we'd need to deprecate again once the real solution lands.

2. **It moves control from the user to the admin.** Notification preferences in HumHub are deliberately user-facing (per category, per space). A global kill switch per content class works against that model.

3. **It overlaps with what's already on the roadmap:**
   - v1.19 already ships activity grouping, which condenses mass uploads on the activity side — this option would switch that off as well, together with the new-content marker.
   - The stream already has content grouping ("Show more") and a hide option.
   - For v1.20 we're planning to delay and group notifications, which solves the actual "40 uploads → 40 e-mails" problem generically while keeping the information.

**What I'd suggest instead as a short-term fix:** The real-world case behind #5334 is a *batch* upload. That can be solved properly in the module that knows the semantics: cfiles/gallery could set `$silentContentCreation` on all records of a multi-upload except one — or better, create a single batch notification ("X added 40 images to gallery Y"). That's exactly the grouping v1.20 will provide generically, just implemented locally where the batch context is known. No core change, no new config API, and the issue is actually solved rather than muted.

So I'd rather close this one in favor of the module-level fix and the v1.20 notification grouping. What do you think — would you be interested in tackling the batch handling in cfiles/gallery instead?

marc-farre:
@luke- thanks for your review.
Sure, grouping notifications is definitively better.
Yes, I can create a PR for these 2 modules, but I'm not quite sure what 1.20 will add, and what consequence on these PRs.

luke-:
@marc-farre At the moment, there's only a very rough plan. So it would probably be more of a temporary solution.

PR #8373, opened 2026-08-03
Ensure oEmbed consent prompt encodes requested URL

kilgor-trout-c22:
### What

Hardens the oEmbed consent-prompt rendering so a requested URL can no longer inject active markup into the prompt shown to viewers.

When embedded content requires consent, the prompt is built from the requested URL. A malformed value that still matches a provider pattern (e.g. containing `youtube.com`) fell back to the full URL, which was inserted into the prompt without output encoding. The rendering path then reconstructed embedded `<script>` elements with a valid CSP nonce, so attacker-supplied markup could execute in another authenticated user's session.

Refs security report / CVE-2026-18526.

### Changes

- `UrlOembed::confirmationContent()` now encodes the URL text (`Html::encode`) so any embedded markup is rendered inert.
- `UrlOembed::getOEmbed()` returns the consent prompt directly instead of running it through the `<script>` nonce-reconstruction step. That step is meant for trusted provider previews only and must not apply to the untrusted prompt (defense in depth).
- Added a regression test asserting the prompt contains no live `<script>` and shows the encoded form instead.

### Testing

- `UrlOembedTest`: 9/9 pass (incl. new test).
- Reproduced end-to-end before the change (prompt executed the injected script) and verified it no longer does after the change (URL rendered as inert text).

See: https://github.com/humhub/humhub-internal/issues/1313

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ HumHub Changelog
 - Enh #8346: Force Select2 dropdowns to always open below the field instead of flipping above when there's not enough space
 - Fix #8359: Display user sub name on invite after space creating
 - Fix #8365: Encode deletion reason in comment and content deletion notifications
+- Fix #8373: Ensure the oEmbed consent prompt encodes the requested URL so embedded markup stays inert
 
 1.18.4 (July 21, 2026)
 ----------------------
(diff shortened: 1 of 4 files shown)


luke-:

yurabakhtin (approved):

PR #8374, opened 2026-08-03
Ensure adding group members is restricted to groups the current user manages

kilgor-trout-c22:
Internal report: https://github.com/humhub/humhub-internal/issues/1312

## Background

In `admin/GroupController`, every group-write action binds its permission check to the
group being written to via `checkGroupAccess()`, which resolves to `Group::canManage()`.
`actionAddMembers` was the only one that did not: it checked `canModifyMembers()`, which
resolves to `Yii::$app->user->isGroupManager()` and is therefore global (manager of *any*
group) rather than group-scoped.

`AddGroupMemberForm::save()` did not close the gap either, since it only guarded the
administrator group. As a result a manager of one group could add members, including
themselves, to any other non-admin group and inherit that group's permission set. The
administrator group was already protected, so administrator itself was not reachable.

## Changes

- `GroupController::actionAddMembers()` now calls `checkGroupAccess()` on the group taken
  from the submitted form, matching the six sibling write actions.
- `AddGroupMemberForm::save()` now requires `$group->canManage()`. This is strictly
  stronger than the previous administrator-group-only guard and covers it, so the
  direct-POST path is bound to the target group independently of the controller.

## Tests

New `AddGroupMemberTest` (admin unit suite), 5 tests / 21 assertions. Verified as a
differential: with the two source changes reverted, exactly the two escalation tests fail,
while the three tests covering legitimate behaviour pass either way.

| Test | without fix | with fix |
| --- | --- | --- |
| manager cannot add to a group they do not manage (form) | fail | pass |
| manager cannot add to a group they do not manage (controller) | fail | pass |
| manager can add to a group they manage | pass | pass |
| administrator can add to any group | pass | pass |
| manager cannot add to the administrator group | pass | pass |

Full `admin` unit + functional suite is green (27 tests, 277 assertions).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ HumHub Changelog
 - Fix #8359: Display user sub name on invite after space creating
 - Fix #8361: Prevent non-creators from attaching or deleting another user's unassigned file
 - Fix #8365: Encode deletion reason in comment and content deletion notifications
+- Fix #8374: Ensure adding group members is restricted to groups the current user manages
 
 1.18.4 (July 21, 2026)
 ----------------------
--- a/protected/humhub/modules/admin/controllers/GroupController.php
+++ b/protected/humhub/modules/admin/controllers/GroupController.php
@@ -286,8 +286,12 @@ class GroupController extends Controller
         }
 
         $form = new AddGroupMemberForm();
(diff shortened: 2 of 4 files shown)


luke-:

yurabakhtin (approved):

PR #8376, opened 2026-08-04
Ensure space member role changes are limited to assignable roles

kilgor-trout-c22:
Validates the space user group (role) written by the member administration role dropdown, and makes the container access check reject roles it does not know.

Details: humhub/humhub-internal#1315 (private).

### Changes

- `space\models\Membership`: adds `SCENARIO_EDIT_ROLE`, whose only active attribute is `group_id`, and restricts `group_id` to `getAssignableUserGroups()` (`admin`, `moderator`, `member`) in that scenario. The owner role is not assignable: it is derived from `space.created_by` and changed through `member/change-owner`.
- `space\modules\manage\controllers\MemberController::actionIndex()`: applies that scenario to the role dropdown request, so it can only set the role of the membership it looked up, and rejects a request targeting the space owner's membership with 403. The dropdown already excluded that row via `readonly`.
- `space\models\Space::getUserGroup()`: a stored role outside the assignable set is reported as `member`, so records holding an unexpected value cannot pass elevated role checks.
- `content\components\UserGroupAccessValidator`: a user group that is not part of the container's known groups no longer satisfies a non-strict `userGroup` rule. `getUserGroupLevel()` keeps its current return contract, so an unknown group listed in a rule's allowed groups still denies access.

### Compatibility

No migration, no schema change. Assigning member, moderator and admin through the dropdown is unchanged, as is `Space::addMember()` with an explicit group. New public API: `Membership::SCENARIO_EDIT_ROLE`, `Membership::getAssignableUserGroups()`, `UserGroupAccessValidator::getUserGroupLevels()`.

Custom content containers that define user groups beyond those in `UserGroupAccessValidator::$spaceGroupLevel` / `$profileGroupLevel` now get no access from non-strict `userGroup` rules instead of full access. Both behaviour changes and the new API are noted in `MIGRATE-DEV.md`.

### Tests

New `space/tests/codeception/functional/MemberRoleChangeCest.php` (7 tests) covers assigning a regular role, the non-assignable and unknown role cases, the space owner's membership, and that a stored unexpected role grants no elevated access.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ HumHub Changelog
 - Fix #8359: Display user sub name on invite after space creating
 - Fix #8361: Prevent non-creators from attaching or deleting another user's unassigned file
 - Fix #8365: Encode deletion reason in comment and content deletion notifications
+- Fix #8375: Ensure space member role changes are limited to assignable roles
 
 1.18.4 (July 21, 2026)
 ----------------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ HumHub Changelog
 - Fix #8359: Display user sub name on invite after space creating
 - Fix #8361: Prevent non-creators from attaching or deleting another user's unassigned file
 - Fix #8365: Encode deletion reason in comment and content deletion notifications
-- Fix #8375: Ensure space member role changes are limited to assignable roles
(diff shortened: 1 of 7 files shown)


luke-:

yurabakhtin (approved):

luke- (approved):

PR #8378, opened 2026-08-04
Ensure marketplace module metadata is encoded in administration views

kilgor-trout-c22:
Module metadata returned by the marketplace API (name, description, versions,
image URL, license key) was rendered into the administration views without HTML
encoding. This change treats that metadata as untrusted input and encodes it at
every output sink.

Internal issue: [humhub-internal#1317](https://github.com/humhub/humhub-internal/issues/1317)

### Changes

- `models/Module.php`: `marketplaceLink()` now encodes its label by default and
  takes a new `bool $encode = true` parameter. The label is the anchor's inner
  content, so the flag is only disabled where that content is locally generated
  markup.
- `widgets/views/module-installed-card.php`, `module-uninstalled-card.php`:
  pass `false` for the `Html::img()` logo link (safe markup, URL already encoded
  by `Html::img()`); `latestVersion` is now encoded. `name` and `description`
  are covered by the new default.
- `widgets/views/module-update-card.php`: encode `name` and the displayed
  versions.
- `views/purchase/list.php`: encode `moduleImageUrl` (attribute context),
  `name`, `description`, `latestVersion` and `licence_key`.

Locally sourced values (installed module version, module id, installation id)
are unchanged.

### Compatibility

`Module::marketplaceLink()` is public and its default output changes from raw to
encoded. No core caller relies on the previous behaviour; external callers that
pass ready-made markup need to opt out with `$encode = false`.

### Testing

Verified against a mocked marketplace API response: metadata renders as text in
both the browse cards and the purchase list, while module logos and marketplace
links render unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ HumHub Changelog
 - Fix #8359: Display user sub name on invite after space creating
 - Fix #8361: Prevent non-creators from attaching or deleting another user's unassigned file
 - Fix #8365: Encode deletion reason in comment and content deletion notifications
-- Fix #XXXX: Ensure marketplace module metadata is rendered as encoded text in the administration views
+- Fix #8378: Ensure marketplace module metadata is rendered as encoded text in the administration views
 
 1.18.4 (July 21, 2026)
 ----------------------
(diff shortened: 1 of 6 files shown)


luke-:

yurabakhtin (approved):

luke- (approved):

PR #8379, opened 2026-08-05
Ensure request supplied membership button options stay inert

kilgor-trout-c22:
See: https://github.com/humhub/humhub-internal/issues/1318

### Changes

- `MembershipButton::sanitizeRequestOptions()` (new): reduces request supplied options to known buttons and to the presentation options that actually need to survive the round trip (`mode`, `mode_method`, `visible`, `groupClass`, `togglerClass` and `attrs.class`, with class values reduced to `[\w \-]`). Titles, URLs and action attributes stay server generated. Accepts a JSON string or an already decoded array and returns an empty array for malformed, scalar or empty input, so a malformed value no longer surfaces as an uncaught exception.
- `MembershipController::actionRequestMembershipForm()`: passes the submitted options through `sanitizeRequestOptions()` instead of `Json::decode()`.
- `MembershipController::getActionResult()`: passes the posted options through `sanitizeRequestOptions()`. This is the second entry point into the same widget and is reachable from `revoke-membership`, `request-membership` and `invite-accept`.
- `views/membership/requestMembershipSave.php`: emits the button via `Json::htmlEncode()` rather than interpolating it into a quoted literal, so the value is safe both as a JavaScript string and inside the `<script>` element. `$spaceId` is cast to int.

### Behavior change

Option keys outside the presentation set are now dropped when they arrive from the request. Widget callers that pass options directly are unaffected. Third party code that relies on round tripping other option keys through the membership request would lose them, so a `MIGRATE-DEV.md` entry may be warranted.

### Tests

New `RequestMembershipSaveViewTest` (space unit suite), 5 tests: rejection of non presentation options, tolerance of malformed input, preservation of the presentation options used by the space header and the space directory, and the encoding in the save response.

- Space unit: 11/11 pass.
- Space functional: 14/14 pass.

Also verified against a running instance that the join flow is unchanged: the Pending button still renders with its icon, class set and confirm text, and the `mode => link` options set by `profileHeaderControls.php` still survive the round trip.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ HumHub Changelog
 - Fix #8365: Encode deletion reason in comment and content deletion notifications
 - Fix #8374: Ensure adding group members is restricted to groups the current user manages
 - Fix #8373: Ensure the oEmbed consent prompt encodes the requested URL so embedded markup stays inert
-- Fix #XXXX: Ensure membership button options supplied by the request are limited to presentation values and safely encoded in the response
+- Fix #8379: Ensure membership button options supplied by the request are limited to presentation values and safely encoded in the response
 
 1.18.4 (July 21, 2026)
 ----------------------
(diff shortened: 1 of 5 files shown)


kilgor-trout-c22:

yurabakhtin (approved):

yurabakhtin:
> Can you please review?

@kilgor-trout-c22 https://github.com/humhub/humhub-internal/issues/1318#issuecomment-5191590063

luke-:
Should be duplicate of: https://github.com/humhub/humhub/pull/8381

Issue #8375, opened 2026-08-04
My Spaces Loading Issues


ArchBlood:
### What steps will reproduce the problem?
On either mobile app or desktop when the loading is happening on the dashboard quickly click on the My Spaces dropdown and it will not load spaces directly till it is closed then reopened, I have tested on multiple different HumHub instances as well as the community site with all instances showing the same results.

Original Post: https://community.humhub.com/content/perma?id=297176

### What is the expected result?
The loader should either display a loading progress wheel or some kind of message stating that loading has either been interrupted or is in the process of loading.
 
### What do you get instead?

https://github.com/user-attachments/assets/73378f75-80a0-43c6-ae35-449d1f3d9eaa

### Additional info

| Q                | A
| ---------------- | ---
| HumHub version   | 1.19
| PHP version      |  8.2 & 8.3
| Operating system | LinuxCloud OS

> Note: Also provide the full error stacks from `Administration --> Information --> Logging` and errors from your browser console if possible.
> Note: Please use the [community](https://community.humhub.com) site for any questions on customizing HumHub.

yurabakhtin:
@ArchBlood @luke- Fixed https://github.com/humhub/humhub/pull/8384

Before fix:

https://github.com/user-attachments/assets/a5dea11b-f276-492e-9eda-a3aae2bcf38a

After fix:

https://github.com/user-attachments/assets/a01b9dd9-ef7d-41f8-87e8-a526127208c2

ArchBlood:
> @ArchBlood @luke- Fixed https://github.com/humhub/humhub/pull/8384
> 
> Before fix:
> 
> https://github.com/user-attachments/assets/a5dea11b-f276-492e-9eda-a3aae2bcf38a
> 
> After fix:
> 
> https://github.com/user-attachments/assets/a01b9dd9-ef7d-41f8-87e8-a526127208c2

Just a thought, would it be easier and for better maintainability to implement a helper in this case?

PR #8390, opened 2026-08-17
Remove redundant space chooser lazy-load click handler

luke-:
Follow-up to #8384 (issue #8375), which landed on `master` and reached `develop` with the last `master → develop` merge.

`SpaceChooser.initEvents()` carried two independent workarounds for the same problem — the My Spaces dropdown can be opened before its widget has finished initializing (the module init is async since #7913), so `shown.bs.dropdown` fires with no listener attached and the lazy space list is never requested:

- the `#space-menu` click handler added in #7913, which re-checked the state on a *later* toggle click, and
- the init-time check from #8384, which detects the already-open dropdown and starts the load immediately.

The click handler is redundant now:

- On the normal open, it does nothing. `triggerRemoteSearch()` runs synchronously from `shown.bs.dropdown` and `ui.loader.set()` replaces the content of `#space-menu-remote-search` with the loader, so by the time the handler's `setTimeout(…, 0)` fires, the `$remoteSearch.children().length` guard already short-circuits.
- On the race it was written for, it only recovered on the *next* click, leaving the dropdown visibly empty until the user reopened it. The init-time check covers the whole window: if the widget initializes after `shown.bs.dropdown` fired, the `show` class is present and the load starts right away; if it initializes before, the listener is attached in time.

No behaviour change on the normal path, and the race it guarded is handled earlier and without the extra reopen.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ HumHub Changelog
 
 1.19.0-beta.2 (TBD)
 -------------------
+- Enh #8390: Removed the redundant My Spaces dropdown click handler — the lazy space list is already loaded on open, and the case where the dropdown was opened before its widget initialized is covered by the init-time check from #8384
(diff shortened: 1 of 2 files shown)


luke-:
@yurabakhtin Can you please check?

yurabakhtin:
@luke- I agree the code removing because the `master` PR https://github.com/humhub/humhub/pull/8384 fixes it.

Issue #3941, opened 2020-03-23
Enhanced asset management


buddh4:
### Description

Improve core and module asset loading. Currently, all core assets are loaded in the head dom section without usage of `defer` or `async`.  Module assets are mostly loaded either at the beginning or bottom of the dom body.

### Concept

This issue describes possible meassures in order to improve the current script/style loading:

#### Split  [AppAsset](https://github.com/humhub/humhub/blob/v1.4.3/protected/humhub/assets/AppAsset.php)

Core assets within [AppAsset](https://github.com/humhub/humhub/blob/v1.4.3/protected/humhub/assets/AppAsset.php) should be split into two seperate asset bundles.  The old AppAsset should only include main core scripts as jquery and a small humhub core. The other bundle will include non essential scripts and will be loaded with `defer` which still assures script loading order. This needs to be reflected in `config/assets.php`.

#### `humhub\components\asset\AssetBundle`

A base AssetBundle class should be implemented which sets some default values (e.g. defer) and provides additional utility properties e.g. `forceCopy`, `position` directly in property instead of `jsOption` and `publishOption`. Also set the default dependency to the deferred bundle.

#### Enable [lazy loading](https://docs.humhub.org/docs/develop/javascript-index#module-dependencies) in humhub modules by default

If lazy loading is enabled by default, the script loading order is not that important as long as dependencies are only used within `init` or after initialization.

#### Preload scripts and stylesheets

Preload some core scripts and styles with `<link rel="preload">`.

#### Remove deprecated scripts

Some scripts currently part of `AppAssets` can be removed.

#### Configuration

In order to stay compatible (see Migration), this behavior should be configurable. This probably would only require disabling the deferred bundle in production.

#### Migration

Modules should extend the new `AssetBundle`, which will automatically add the dependency to the deferred bundle and set  `defer` script loading by default.

PR #8411, opened 2026-08-25
Remove the dead defaultDepends asset bundle mechanism

luke-:
`humhub\components\assets\AssetBundle::$defaultDepends` promises that every bundle implicitly depends on `CoreBundleAsset` without listing it in `$depends`. It has never done that: the code applying it reads a misspelled property.

```php
if ((!(Yii::$app instanceof Application) || !Yii::$app->request->isAjax) && !empty($this->dependsDefault)) {
    $this->depends[] = ArrayHelper::merge($this->depends, $this->defaultDepends);
}
```

`$this->dependsDefault` does not exist, so the branch has been dead since it was added in 1.5 (#3941). Its body is wrong too — it would append the merged array as a single nested element of `$depends`.

## Why remove instead of repair

With the property name corrected, the mechanism takes the site down: every bundle inside `CoreBundleAsset`'s own dependency tree then depends on `CoreBundleAsset`, and Yii aborts the request.

```
yii\base\InvalidConfigException: A circular dependency is detected for bundle 'humhub\assets\CoreBundleAsset'.
```

(Verified locally by fixing the property name and loading a page.) Making it work would mean auditing every bundle in that tree and opting each one out — for a behaviour nobody has been missing for five years. Bundles that need the core bundle list it in `$depends` themselves, which is what they have been doing all along.

## Scope

- the property, its dead branch and the imports that only served them
- the `public $defaultDepends = false;` opt-outs in 12 core bundles and the two entries in the asset build config — left behind they would set a property that no longer exists
- a `Removed` note in `docs/develop/module-migrate.md`

No runtime behaviour changes: the removed branch never executed.

## Module impact

A search across 167 module repositories (`humhub`, `humhub-contrib`, `cuzy-app`) finds no use of `defaultDepends`. A module that declares the property keeps working — it simply becomes an unused own property. A bundle *configured* with `'defaultDepends' => ...` would now fail with `Setting unknown property`; no module does that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ HumHub Changelog
 
 1.19.0-beta.3 (Unreleased)
 -------------------------------
+- Enh #8411: Removed the `AssetBundle::$defaultDepends` mechanism — a property-name typo kept it from ever running, and activating it makes the core bundle depend on itself, see `docs/develop/module-migrate.md`
 - Fix #8401: Module version migration crash when module versions cannot be determined
 - Fix #8400: Button text horizontal centering and spacing in People heading (since 1.19.0-beta.2)
 - Fix #8404: Remove deprecated function `curl_close()`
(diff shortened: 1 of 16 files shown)


luke-:

yurabakhtin (approved):

PR #8412, opened 2026-08-25
Fix applying a UI addition by its id

luke-:
Two defects on the same path in `humhub.ui.additions.js` — applying a registered addition by its id.

## `extend(..., {applyOnInit: true})` throws

```js
if (options.applyOnInit) {
    module.apply('body', id);   // apply() does $element.find(...)
}
```

`apply()` takes a jQuery object, so this throws `TypeError: $element.find is not a function` the moment it fires.

## An addition registered without a selector can never be applied by id

`register()` supports the selector-less form documented in [UI additions](../docs/develop/ui-js-uiadditions.md#register-addition-without-selector-since-v14): the addition is dispatched per element through the generic `[data-ui-addition]` addition, and the docs state its "actual selector" is `[data-ui-addition="<id>"]`. The stored selector was `null` instead, so `apply()` fell back to it, did `$element.find(null)` and matched nothing — applying such an addition explicitly was a silent no-op. `register()` already worked around this internally by passing the derived selector by hand.

Such additions now store that derived selector themselves. They stay out of `_order` as before, so `applyTo()` still does not sweep them globally and the per-element dispatch is unchanged.

## Verified in the browser

On a running instance, before and after, using a fresh `<time data-ui-addition="timeago" datetime="…">RAW</time>`:

| | before | after |
|---|---|---|
| `additions.apply($('body'), 'timeago')` | `"RAW"` (nothing happens) | `"about 5 hours ago"` |
| `additions.extend(id, fn, {applyOnInit: true})` | `TypeError: $element.find is not a function` | no error |
| timeago elements rendered by the page | `"7 days ago"` | `"7 days ago"`, no console errors |

No module in the `humhub`, `humhub-contrib` or `cuzy-app` orgs uses `applyOnInit`; core's own Vue island runtime documents the first bug and works around it, and that workaround can go once this lands.


luke-:

yurabakhtin:
@luke- I approve the fixes.

Tests:

1.

```js
humhub.modules.ui.additions.register('demoTest', function($m){ console.log('applied', $m.length); });
humhub.modules.ui.additions.extend('demoTest', function(){}, {applyOnInit: true});
```

test1

2.

```js
humhub.modules.ui.additions.register('demoTest', function ($match) {
    console.log('[demoTest] fired, $match.length =', $match.length,
        'is body?', $match.is('body'), 'is #target?', $match.is('#target'));
}, { overwrite: true });

var $el = $('<div id="target" data-ui-addition="demoTest">test</div>').appendTo('body');
humhub.modules.ui.additions.apply($('body'), 'demoTest');
```

test2

yurabakhtin (approved):

PR #8416, opened 2026-08-26
Auto-focus the newly loaded content after a pjax swap, so Tab no longer restarts at the top of the page

### LINKED ISSUE #438: Wrong tab-order after opening or editing a wiki page
Accessibility: If I use the keyboard to open or edit the page and then hit "tab" again, the focus is not as expected at the next element for the loaded content. It is set to the profile picture on the top left. So I need to tab through the whole previous content again.
Link to BITV: https://bitvtest.de/pruefschritt/bitv-20-web/bitv-20-web-9-2-4-3-schluessige-reihenfolge-bei-der-tastaturbedienung

### Steps to reproduce the issue
1. Press tab to select a page and hit enter
2. The page opens
3. Press tab again

yurabakhtin:

- [x] Feature

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
https://github.com/humhub/wiki/issues/438


luke-:
> [!NOTE]
> This review was generated with AI assistance (Claude Code) and reviewed by me before posting. Please push back on anything that looks wrong — some of it is reasoning about code paths rather than something I ran in a browser.

Thanks for tackling this — the underlying issue is real and worth fixing (focus resets to `<body>` after a pjax swap, so Tab restarts at the top). A few things concern me about the approach, though.

### The runtime diff re-derives something the markup already states

`findContentTarget()` spends ~130 lines guessing which subtree is new. But the layouts already mark this explicitly — `modules/space/views/space/_layout.php`:

```php
<aside class="col-lg-2 layout-nav-container" aria-label="Sidebar">   <!-- exactly what we want to skip -->
<div   class="col-lg-7 layout-content-container">                     <!-- exactly the target we want -->
<aside class="col-lg-3 layout-sidebar-container" aria-label="Sidebar">
```

Same structure in `user/views/profile/_layout.php`, `dashboard/views/dashboard/index.php` and `notification/views/overview/index.php`. So `container.querySelector('.layout-content-container') || container` gets us the same landing point deterministically, with no clone, no HTML serialization and no regex. For pages that don't carry the class (admin/module pages) the fallback is `#layout-content` — which is what this PR already falls back to when the diff yields nothing — and adding the class to those layouts is a one-line change that helps CSS and acceptance tests too.

### The heuristic misses the main case it targets

`isEquivalentNode()` compares normalized `outerHTML`, but the active menu entry carries an `active` class (`modules/ui/menu/MenuLink.php:69`). So the space menu is *always* "changed" when navigating within a space:

* **Without a sidebar** (`hasSidebar() === false`): the children of `.row.space-content` are `[aside.layout-nav-container, div.layout-content-container]` — both count as changed, so `hasUnchangedSibling` stays `false`, the loop breaks, and focus lands on `.row.space-content`, i.e. *above* the nav. Tab then walks the whole left menu again — the exact behaviour this PR is meant to fix.
* **With a sidebar**: the pick comes down to `textContent.length` between nav and content. On sparse pages (empty member list, small settings form) the nav wins and focus lands in the menu.

### Smaller points

* **Cost per navigation:** `cloneNode(true)` on `#layout-content` on every pjax request, plus `outerHTML` serialization of both trees at each level. On a stream page that's a lot of string building per navigation; `DIFF_MAX_ELEMENTS = 8000` only caps the pathological case, not the normal one.
* **Focus stealing:** the guard checks `container.contains(activeElement)`, so focus parked *outside* the container gets stolen. `humhub.modules.client.reload()` (`humhub.client.js:100`, used after modal actions among others) would then pull focus out of an open `#globalModal` (body-level) into the background content and break the focus trap — an a11y regression rather than a fix.
* **Back/forward isn't covered:** the pjax cache path only fires `pjax:beforeReplace` + `pjax:end` (`jquery.pjax.modified.js:545-552`), no `beforeSend`/`success`, so browser back still drops focus to `<body>`.
* **No announcement:** focusing a `div[tabindex="-1"]` isn't reliably announced by screen readers. For a navigation change the target usually wants to be a heading (`h1` / `.panel-title`), and/or an `aria-live` region announcing the new page — we don't have such a region in core yet.
* **Only `.d-none` is filtered:** elements hidden via `hidden`, `style="display:none"` or `.collapse` can still win the `textContent.length` contest, and `focus()` on them is a no-op — the fix then fails silently.
* **Nits:** `NOISE_TAGS` and `AUTO_ID_TOKEN_PATTERN` are re-created on every call instead of living at module scope; `tabindex="-1"` is never cleaned up on blur; there's no opt-out (a `module.config.autoFocus` or `data-pjax-focus-prevent` would match the existing `module.config.active` pattern); `isEquivalentNode` collides in name with the DOM's own `isEqualNode`.

### Suggestion

Could we go markup-driven instead? Something like a target chain of `#layout-content [data-pjax-focus]` → `.layout-content-container` → a heading inside it → `#layout-content`, plus:

* widening the guard to "focus is somewhere meaningful / a modal is open",
* handling `pjax:end` so back/forward is covered,
* a short screen-reader announcement.

That would be considerably less code, deterministic, testable — and it would work in the sidebar-less space case where the current heuristic doesn't.

What do you think? Happy to be wrong about any of this — especially the sidebar-less trace, which I reasoned through from the markup rather than by stepping through it in a browser. If you've verified that case manually, that would settle it. And is there a reason the layout classes couldn't be used that I'm missing?

yurabakhtin:
@luke- New changes https://github.com/humhub/humhub/pull/8416/changes/c3f4002f8482b8850f0b5291c6726fde2aaee932:

**Fixed:**
- **Active nav class defeating the "skip unchanged menu" step** — confirmed this live: cross-section navigation (e.g. Home → Wiki within a space) landed focus one level too high because the space menu's `active` link makes its markup genuinely differ between renders. `isEquivalentNode()` now strips the `active` class before comparing.
- **Markup-driven target, as you suggested** — `manageFocus()` now checks for `.layout-content-container` first (present on space/profile/dashboard/notification layouts) and only falls back to the diff heuristic when it's absent. Cheaper and more reliable for the common case, and it also sidesteps the active-class issue on its own.
- **Back/forward (bfcache) navigation wasn't covered at all** — verified in `jquery.pjax.modified.js`: a cache hit only fires `pjax:beforeReplace`/`pjax:end`, never `beforeSend`/`success`. Moved the capture/manage hooks to the former pair, which fires on both the normal and cached paths.
- **Focus could be stolen from an open modal** — `#globalModal` lives outside `PJAX_CONTAINER_SELECTOR` (rendered once via `LayoutAddons`, skipped on pjax requests), so a background `reload()` while a modal is open and focused could steal focus out of it. Added an early bail-out via `activeElement.closest('.modal.show')`.
- **Hidden-element filtering was `.d-none`-only** — extended to also cover `.collapse:not(.show)`, the `hidden` attribute, and inline `display:none`. Deliberately not using `getComputedStyle()`: both the pre-swap snapshot and everything `normalize()` compares are detached clones, and computed-style resolution isn't reliable on a node that isn't attached to the document — the direct property checks are both cheaper and correct there.
- **The scope nit** — `NOISE_TAGS`, `AUTO_ID_TOKEN_PATTERN`, `DIFF_MAX_ELEMENTS`/`DIFF_MAX_DEPTH` and `LAYOUT_CONTENT_SELECTOR` are now module-scope constants instead of being recreated inside `manageFocus()` on every navigation.

**Left as-is:**
- **Screen-reader announcements** — agreed this is a real gap (a `tabindex="-1"` focus target isn't announced), but it's a separate, self-contained addition (an `aria-live` region) rather than a fix to what's here, so leaving it for a follow-up rather than folding it into this PR.
- **`cloneNode`/`outerHTML` cost** — kept the diff algorithm as-is. It's bounded by `DIFF_MAX_ELEMENTS`, runs once per navigation rather than per frame, and with the markup-driven fast path above it no longer even runs for the standard layouts — so the actual cost in practice is smaller than it looked before this round of fixes.

luke-:
> [!NOTE]
> Same as before: reviewed with AI assistance (Claude Code) and checked by me before posting. Push back where I'm wrong — this is reasoning over the code paths, not something I stepped through in a browser.

Thanks for the quick turnaround — this is a lot better. The `.layout-content-container` fast path is exactly what I meant, and it makes the sidebar-less space case deterministic.

One thing I want to check before anything else: I traced whether the `.active` stripping is still needed now that the fast path exists, and it is — `modules/admin/views/layouts/main.php` uses bare `col-lg-3` / `col-lg-9` with no marker class, so the diff still runs there, and it's only the stripping that makes `AdminMenu` compare as unchanged so `col-lg-9` wins. Good call keeping both.

### A correction on my side

My back/forward point was right about the code path but mostly theoretical in practice: `PjaxLayoutContent.php:34` sets `cache => false`, so the cached popstate branch never gets `contents` and falls through to `pjax(options)` — i.e. the normal request path, which already fired `beforeSend`/`success`. Sorry for sending you after that; the move to `beforeReplace`/`end` is still defensible (it's correct if `cache` is ever enabled), but it's worth knowing the gap it closes is narrow, because it brought the next point with it.

### `pjax:end` fires when no swap happened

`options.complete` (`jquery.pjax.modified.js:280-287`) fires `pjax:complete`/`pjax:end` on abort, error and timeout as well as on success — and pjax aborts in-flight requests itself (`:418` when a second navigation starts, `:488` on popstate), with `timeout: 5000` as our default (`PjaxLayoutContent.php:35`). So `manageFocus()` can now run against a container that was never replaced. The `container.contains(activeElement)` guard catches most of it, so this isn't severe, but `pjax:beforeReplace` only fires on paths that actually swap — a `swapPending` flag set there and consumed in `manageFocus()` would make it exact for ~3 lines.

### Nested pjax containers (my miss in the first round)

The handlers sit on `document`, and `pjax:beforeReplace` / `pjax:end` / `pjax:success` all bubble up from whichever container fired them. `modules/admin/views/setting/topics.php` wraps a `GridView` in `Pjax::begin(['id' => 'global-topics'])`, so paging/sorting that grid triggers these handlers too. In that specific case it degrades gracefully (admin has no marker class, so the diff still lands inside the grid area), but generically it's wrong: a module with a nested pjax container on a layout that *does* have `.layout-content-container` would get focus thrown to the top of the whole content area instead of staying at the grid.

Both handlers receive `options`, on the cached path too, so:

```js
if (!options || options.container !== PJAX_CONTAINER_SELECTOR) {
    return;
}
```

This was already the case in the first revision via `pjax:success`, so it's not a regression from this round — I just didn't spot it.

### Two smaller things

* **The snapshot clone still runs on every navigation.** `captureFocusSnapshot()` does `cloneNode(true)` on `#layout-content` unconditionally, so the cost argument ("no longer even runs for the standard layouts") holds for `findContentTarget()` but not for the deep clone, which is the more expensive half and is thrown away unused whenever the fast path wins. An early `if (container.querySelector(LAYOUT_CONTENT_SELECTOR)) return;` in the snapshot would skip it.
* **Modal guard**: `document.querySelector('.modal.show')` would be more robust than `activeElement.closest('.modal.show')` — if focus has ended up on `<body>` while a modal is open, the current check doesn't fire.

Nothing here is a blocker from my side beyond the container scoping, which I'd want in before merge. Does the `options.container` guard look right to you, or is there a reason the inner containers should be handled as well?

yurabakhtin:
@luke- Commit https://github.com/humhub/humhub/pull/8416/changes/28ce4914f0a3b4a730adf3b2e30dae1a7de6d508:

**On `.active` stripping**: right, confirmed `modules/admin/views/layouts/main.php` has no marker class (bare `col-lg-3`/`col-lg-9`), so the diff still runs there and the stripping is what makes `AdminMenu` compare as unchanged. Keeping it.

**On back/forward**: no worries — good to know the gap it closes is narrow given `cache: false` in `PjaxLayoutContent`. Leaving the `beforeReplace`/`end` hooks as-is since they're still correct if caching is ever turned on, just not claiming it fixes an active bug today.

**`pjax:end` firing without a swap**: fixed with the `swapPending` flag you suggested — set in `captureFocusSnapshot()`, checked and cleared at the top of `manageFocus()`. If `pjax:end` fires from an aborted/errored request that never got a `beforeReplace` for our container, `manageFocus()` now bails immediately.

**Nested pjax containers — to answer your question directly**: yes, going with the `options.container !== PJAX_CONTAINER_SELECTOR` guard on both handlers, no special-casing for inner containers. Verified against your `#global-topics` example specifically — sorting that grid fires `beforeReplace`/`end` with `options.container === '#global-topics'`, so the guard correctly skips it now. Didn't see a reason to handle nested containers instead of just ignoring them — they manage their own focus concerns (or don't need to), and `#layout-content` is the only container this module owns.

**Two smaller ones**: also fixed both — `captureFocusSnapshot()` now skips the clone when the current container already has `LAYOUT_CONTENT_SELECTOR` (fast path will win, so the clone would've been wasted), and the modal check is now `document.querySelector('.modal.show')` instead of `activeElement.closest(...)`, so it still catches the case where focus already fell to `<body>` while a modal was open.

For keyboard-only users this is already fully solved — Tab order works correctly, verified live across several scenarios. The `aria-live` announcement gap only matters for screen-reader users; leaving it for a follow-up, not because it's technically hard (it's a small addition — a hidden live region plus reusing `document.title`, which pjax already keeps in sync), but because it's a separate, self-contained feature rather than a fix to what's in this diff, and it touches the shared layout markup in addition to this JS file.

PR #8442, opened 2026-09-08
DatePicker month-name mismatch between jQuery UI translations and PHP intl/ICU for en-GB and 14 other locales

yurabakhtin:

- [x] Bugfix

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
https://github.com/humhub/humhub-internal/issues/1332


luke- (approved):

luke-:
Reviewed this together with the analysis in humhub-internal#1332 and pushed a follow-up commit ea36d89773 on top.

**Confirmed**
- Root cause and failure chain reproduce on ICU 78.3 as well (`en-GB` → "Sept", strict parse of "15 Sep 2005" fails; `yii\jui\DatePicker::renderWidget()` then re-formats the invalid value via `asDate()`, which is where the extra "t" appears).
- Precondition is a textual input format: the Birthday field hard-codes `medium`, while `dateInputFormat` defaults to `short` (numeric), so the Date/DateTime profile fields were never affected.
- Same check over all 48 shipped locales (strict parse as in `yii\validators\DateValidator`): 17 affected, the 15 listed here plus `am`/`ht`/`uz`, which f86b44b now covers.

**Follow-up commit ea36d89773: only sync the month names the format actually uses**

jQuery UI uses `monthNames` also for the calendar header and `monthNamesShort` for the month dropdown, and ICU's `MMMM` is the grammatical *format* form. Overriding both arrays unconditionally changed the header month in 23 of 48 locales (10 grammatically wrong, e.g. ru "сентября" instead of "Сентябрь", also uk, pl, hr, lt, sk, fi, el, ca "de setembre", vi; 12 capitalization-only) and the dropdown labels even for numeric formats, i.e. for the default `short` configuration (de "Sept." instead of "Sep").

`getIntlMonthNames()` now takes the jQuery UI date format and generates only the arrays it uses for formatting/parsing the input value: `M` → `monthNamesShort` (ICU `MMM`), `MM` → `monthNames` (ICU `MMMM`). Numeric formats stay untouched and the header keeps jQuery UI's standalone names. Name generation is pinned to UTC/Gregorian (the arrays are indexed by Gregorian month). Added `DatePickerTest` (5 cases, incl. `uz` and a round trip through `DbDateValidator`) and a MIGRATE-DEV entry for the removed `DatePickerRussianLanguageAsset`.

Verified locally: `DatePickerTest` + `DBDateValidatorTest` green, php-cs-fixer clean, and in headless Chromium with the real jQuery UI + `datepicker-en-GB.js`: before → input "15 Sep 2005" and `parseDate("15 Sept 2005")` throws "Unexpected literal at position 6"; after → "15 Sept 2005" round-trips, header "September", dropdown "Sept"; ru header stays "Сентябрь".

**Notes on the later commits**
- On ICU 78.3 the `ru`/`uk` medium patterns contain plain spaces, so the U+202F handling doesn't trigger here; it is defensive for the builds where it does, and the validator tests pass with it.
- Removing the public class `humhub\assets\DatePickerRussianLanguageAsset` in a patch release is functionally fine now (ICU names cover `ru`), but consider keeping a `@deprecated` stub for one release in case a theme or module registers it. I recorded it under 1.18.6 in MIGRATE-DEV.md as removed for now.
- Still separate: `cy`/`sw` have no jQuery UI i18n file at all (404 on asset load), and `fa-IR`.

luke-:
@yurabakhtin Can you please check my commit and let me know when I can merge.

yurabakhtin:
> Can you please check my commit and let me know when I can merge.

@luke- Thanks for the improvments, I've reviewed by Claude and tested several languages manually, it works well, this PR is ready for merging.

---

- **The gap**: my original fix unconditionally overrode *both* `monthNames` and `monthNamesShort` with ICU values, even though jQuery UI only needs one of them (whichever the field's own `dateFormat` actually uses) to fix the Save/re-open bug.
- **Side effect**: overriding both also changed the calendar header and month dropdown in roughly half of HumHub's locales, replacing jQuery UI's grammatically-correct translation with ICU's wording (e.g. Russian nominative vs. genitive) — an unintended UI change.
- **The fix**: `getIntlMonthNames()` now takes the widget's `dateFormat` and, via regex on its tokens (`M` vs `MM`, ignoring quoted literals), generates only the array that format actually consumes — numeric formats get no override at all.
- **Extras**: `IntlDateFormatter` is now built with explicit `UTC`/`GREGORIAN` instead of server defaults, removing an environment-dependent edge case.
- **Tests**: a new 5-case `DatePickerTest.php` locks in this behavior, so a future regression on any of these locale/format combinations would be caught automatically.

PR #8448, opened 2026-09-10
Destructive admin actions (remove all space members, delete profile category, reset invite link, remove licence) ran on GET via CSRF

yurabakhtin:

- [x] Bugfix

- [ ] All tests are passing
- [x] Changelog was modified

**Other information:**
https://github.com/humhub/humhub-internal/issues/1335

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ HumHub Changelog
 - Fix #8423: Fix tab order of the reset filters button in the search area
 - Fix #8438: Endless scrolling stalled whenever the loaded stream entries did not push the stream end indicator out of the observed area (compact streams, short entries, viewport not filled), because an `IntersectionObserver` only reports state *changes* — the stream now keeps loading until the observed area is filled
 - Fix #8438: The mobile "Load more" button of a stream was hardwired to the `#wallStream` id, so it failed with "Handler not found" in any stream rendered with a custom `id`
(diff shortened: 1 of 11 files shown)


luke-:
The same bug class is still open in `MembershipController`, the controller this PR hardens: `actionRevokeNotifications()` and `actionReceiveNotifications()` (lines 149 and 157) still mutate state on GET without `forcePostRequest()`.

Both flip the requesting user's `send_notifications` setting for the space, so a logged-in victim can be silenced by a plain ``. Impact is lower than the actions fixed here, since it only affects the victim's own setting and destroys nothing, but it is the identical GET-CSRF pattern.

Worth noting that the view side is already correct: both `MenuLink` entries in `HeaderControlsMenu` (lines 95 and 103) already carry `'data-method' => 'POST'`. Only the server-side enforcement is missing, so this is a two-line change with no UX impact and no risk to any confirm dialog:

```php
public function actionRevokeNotifications()
{
    $this->forcePostRequest();
    // …
}

public function actionReceiveNotifications()
{
    $this->forcePostRequest();
    // …
}
```

The rest of the PR looks correct to me. I checked that all five call sites were converted consistently, and that the `data-method` → `data-action-method` rewrite in `Button::run()` keeps the confirm dialogs working for the converted links.

yurabakhtin:
@luke- Done https://github.com/humhub/humhub/pull/8448/commits/44953bf4077bb7ecfc6175c21b9f5f5f68042328

PR #8493, opened 2026-09-16
Fix TypeError in getCurrentAuthClient() for guest users

marc-farre:
## Summary
- `AuthController::actionLogout()` calls `Yii::$app->user->getCurrentAuthClient()` without checking whether the user is still authenticated (no access rule blocks guests from that action)
- When the session's identity is already gone (stale tab resubmitting the logout form, session expired, already logged out in another tab), `getAuthClientUserService()` is instantiated with a `null` identity, and `AuthClientUserService::__construct()` throws a `TypeError`
- `getCurrentAuthClient()` now returns `null` immediately for guests, since a guest has no current auth client by definition

## Test plan
- [x] Log out normally with an auth-client-based session, confirm SLO handling still works
- [x] POST to `/user/auth/logout` as a guest (e.g. resubmit a stale logout form after already being logged out) and confirm no error is thrown


marc-farre:
@luke- This is a AI generated PR to fix this issue I receive in the log with HumHub 1.19-beta-2, but after analysing the code, the same error could happen on 1.18, hence this PR on master:

```
TypeError:
humhub\modules\user\services\AuthClientUserService::__construct():
Argument #1 ($user) must be of type humhub\modules\user\models\User,
null given, called in
/var/www/protected/humhub/modules/user/components/User.php
on line 279 and defined in
/var/www/protected/humhub/modules/user/services/AuthClientUserService.php:28
Stack trace:
#0
/var/www/protected/humhub/modules/user/components/User.php(279):
humhub\modules\user\services\AuthClientUserService->__construct(NULL)
#1
/var/www/protected/humhub/modules/user/components/User.php(134):
humhub\modules\user\components\User->getAuthClientUserService()
#2
/var/www/protected/humhub/modules/user/controllers/AuthController.php(585):
humhub\modules\user\components\User->getCurrentAuthClient()
```

luke- (approved):

PR #8495, opened 2026-09-17
Honour isFixed() on the mailer transport-security checkboxes

marc-farre:
# Fixed mailer settings: two checkboxes ignore `isFixed()`, and the env test documents a dead form

## The problem

HumHub lets an instance pin a setting from the environment, `HUMHUB_FIXED_SETTINGS__BASE__MAILER_HOSTNAME=…`.
`SettingsManager::getFixed()` then always wins over the database, and `SettingsManager::set()` silently discards any
write to that setting. The admin UI reflects that: every field in Administration → Settings → Advanced → Mailing is
rendered read-only when its setting is fixed.

**Two fields were left out.** `useSmtps` and `allowSelfSignedCerts` are rendered as plain checkboxes:

```php
<?= $form->field($model, 'useSmtps')->checkbox() ?>
<?= $form->field($model, 'allowSelfSignedCerts')->checkbox() ?>
```

On an instance that fixes `mailerUseSmtps`, an administrator therefore sees a form where the host, port, username,
password and sender address are visibly locked, while the transport-security checkbox next to them looks editable.
Ticking it and saving appears to succeed, the setting does not change, and the form comes back showing the old
value with no message explaining why. The same two lines appear in the installer's mail step, which has the same
`isFixed()` treatment on every other field.

This matters most to hosted and containerised setups, where the mailer is exactly what the operator fixes and the
tenant administrator is exactly the person who will try to change it.

## The fix

Pass `disabled` from `isFixed()`, the way `admin/views/setting/file.php` already does it for its checkboxes:

```php
<?= $form->field($model, 'useSmtps')->checkbox(['disabled' => $settings->isFixed('mailerUseSmtps')]) ?>
<?= $form->field($model, 'allowSelfSignedCerts')->checkbox(['disabled' => $settings->isFixed('mailerAllowSelfSignedCerts')]) ?>
```

in both `modules/admin/views/setting/mailing_server.php` and `modules/installer/views/config/mail.php`.

`readonly` is not an option: it has no effect on a checkbox in HTML, which is likely why these two were skipped
when the read-only treatment was applied to the text fields around them.

## Second change: `ConfigTest::testFixedSettings()`

The test asserted the nested form:

```php
'HUMHUB_FIXED_SETTINGS__BASE__MAILER__DSN' => 'smtp://...',   // → fixed-settings['base']['mailer']['dsn']
```

The assertion is true — `EnvHelper` does produce that — but `mailer.dsn` stopped being a setting name in 1.18, when
`m250226_125226_rename_mailer_vars` flattened the mailer settings to `mailerDsn`, `mailerHostname` and friends.
`SettingsManager::getFixed()` reads `fixed-settings['base'][<setting>]`, so the shape the test documents configures
nothing at all, and nothing anywhere reads the nested key as a fallback. Anyone using the test as the reference for
writing an env file gets a variable that parses, starts the container, and silently does not apply.

`testFixedSettings()` now asserts the flat form, and the nested behaviour keeps its own test,
`testNestedFixedSettingsDoNotReachTheSettingName()`, which pins the `__` nesting *and* asserts that the result is
not the setting name.

## Compatibility

**Instances that fix nothing — the overwhelming majority — render byte-identical HTML.** `isFixed()` returns a real
`bool`, and `BaseHtml::renderTagAttributes()` omits a boolean attribute whose value is `false`, so `disabled => false`
emits no attribute. Verified by rendering both forms and comparing:

```
no options (today):        <input type="hidden" … value="0"><input type="checkbox" … value="1" checked>
disabled => false:         <input type="hidden" … value="0"><input type="checkbox" … value="1" checked>   identical
disabled => true:          <input type="hidden" … value="0" disabled><input type="checkbox" … value="1" checked disabled>
```

**Instances that do fix one of the two settings**: `Html::booleanInput()` disables the hidden "uncheck" input
alongside the checkbox ("make sure disabled input is not sending any value"), so the attribute is absent from the
POST entirely. `MailingSettingsForm::init()` has already loaded the attribute from the fixed setting, `load()`
leaves it untouched, the `boolean` validator passes on that value, and `save()`'s `set('mailerUseSmtps', …)` was
already being discarded. No value changes, nothing is lost, and the UI now tells the truth.

**The view's JavaScript** shows and hides `#encryptionOptions` from `$('#mailingsettingsform-usesmtps').prop('checked')`
on load, and on the checkbox's `change` event. A disabled checkbox still reports `checked` correctly, so the initial
state is unchanged; the `change` handler cannot fire, which is right, because the value can no longer change.

**Fixing only one of the two settings** behaves as expected: only that checkbox is disabled.

**The test change** cannot affect a running instance.

## Testing

The two rewritten test bodies were executed directly against `EnvHelper::toConfig()` (flat mapping, nested shape,
and the absence of the flat key in the nested case): all pass. The full Codeception suite was **not** run — it needs
a `humhub_test` database and a `tests/config/env/env.php`, which the machine this was written on does not have; an
untouched test file fails there identically, so this is an environment limitation, not a result. Please let CI run
it.

## Note for the changelog

Both entries are written with a `#NNNN` placeholder — replace it with this pull request's number.


marc-farre:
One thing this PR deliberately leaves alone, for the team to decide on.

Disabling the field is the right behaviour, but the form still never explains *why* it is disabled. On an instance that fixes mailer settings through the environment, an administrator sees greyed-out inputs with no indication that the value comes from `HUMHUB_FIXED_SETTINGS__*` (or `config.php`) and cannot be changed from the UI at all. A greyed-out field usually reads as "not applicable right now" or "you lack a permission", rather than "this one is owned by whoever runs the server".

This is not specific to the two checkboxes here: it applies to every fixed setting — `baseUrl`, `defaultTimeZone`, the other mailer fields, `useXSendfile`, and any module setting that follows the same pattern. A hint would presumably belong at the form/`ActiveField` level, rendered whenever `isFixed()` is true, rather than being repeated in each view.

I left it out on purpose. It is a UX change spread across many forms while this PR is a bugfix, and the choice between adding a hint everywhere, adding it only where operators commonly fix settings, or leaving it as it is belongs to you. Happy to open a separate issue for it, or a pull request, if you would like it done.

marc-farre:
@luke- This PR and 1st comment are AI generated.
It's about:
image
That should be disabled if they are already configured in a file config (fixed config).

luke- (approved):

